PR #776, opened 2014-12-23
WIP: get all records on startup, make fs synchronous

mykmelez:
This is an experiment to see if it improves perf to get all fs db records on startup and then call fs functions (mostly) synchronously.


mykmelez:
I'm not currently planning to continue work on this, so I'm going to close the PR. I'll reopen (or create a new one) if I revisit it at some point.

marco-c:
This could be useful to be able to compile more functions with the baseline compiler, but first of all we should figure out whether functions that use these natives are worth optimizing or not.

Issue #728, opened 2014-12-13
Support encoding audio with the platform AMR codec


marco-c:
This is supported for certified apps since Firefox OS 1.4

yxl:
Gecko [bug 1111135] was fixed, and the platform encoding is supported for the privileged app since fireofox 37 and for certified app since Firefox OS 1.4.

PR #794, opened 2014-12-30
Extract amr from 3gpp container

yxl:
It made necessary changes to fix bug #728 from the j2me.js side.  Once gecko bug https://bugzilla.mozilla.org/show_bug.cgi?id=1111135 is fixed, we are able to use this patch to record audio to 3gp file and convert the 3gp to amr file required by the j2me media API.


midp/media.js
```diff
@@ -83,6 +83,40 @@ Media.supportedImageFormats = ["JPEG", "PNG"];
 Media.EVENT_MEDIA_END_OF_MEDIA = 1;
 Media.EVENT_MEDIA_SNAPSHOT_FINISHED = 11;
 
+Media.convert3gpToAmr = function(uint8Array) {
```
yxl: Fixed.

midp/media.js
```diff
@@ -83,6 +83,40 @@ Media.supportedImageFormats = ["JPEG", "PNG"];
...
+    while (inOffset + 8 < uint8Array.length) {
+        // Get the box size
+        var size = 0;
+        for (var i = 0; i < 4; i++) {
+            size = uint8Array[inOffset + i] + (size << 8);
+        }
+        // Search the box of type mdat.
+        var type = textDecoder.decode(uint8Array.subarray(inOffset + 4, inOffset + 8));
+        if (type === "mdat" && inOffset + size <= uint8Array.length) {
+            // Extract raw ARM data from the box and append to the out buffer.
```
yxl: Fixed.

midp/media.js
```diff
@@ -753,7 +788,7 @@ AudioRecorder.prototype.stop = function() {
             // The audio data we received are encoded with a proper format, it doesn't
             // make sense to concatenate them like the socket, so let just override
             // the buffered data here.
-            this.data = new Uint8Array(message.data);
```
yxl: I neglected that there are other types except 3gp can be recorded.  I fixed this with my last commit ` e1be4e6`.  It will the check the required file format before doing 3gp to amr conversion.

PR #801, opened 2014-12-31
Just in time compiler.

brendandahl:
A lot of the changes are just moving things over to typescript. 
We can probably ignore reviews of things that come directly from shumway:
jit/c4/*
shumway/*

Currently, the compiled code is not part of the default build. I'll file a follow up bug for that.


marco-c:
> @marco-c What js engine/version does your "js" command point to(needs to be a recent spidermonkey version) ? 

It's an old version, I'll download a more recent one

classfile/reader.ts
```diff
@@ -0,0 +1,121 @@
...
+      return data;
+    }
+
+    readDouble() {
+      var data = this.view.getFloat64(this.offset, false);
+      this.offset += 8;
+      return data;
+    }
+
+    readStringFast(length: number): string {
```
marco-c: This is similar to the util.javaUTF8Decode function. We should determine which one is faster and avoid code duplication

classfile/reader.ts
```diff
@@ -0,0 +1,121 @@
...
+      if (length === 1) {
+        var c = this.u8[this.offset];
+        if (c <= 0x7f) {
+          this.offset ++;
+          return String.fromCharCode(c);
+        }
+      } else if (length < 1024) {
+        return this.readStringFast(length);
+      }
+      return this.readStringSlow(length);
```
marco-c: For small strings the JS implementation is faster than TextDecoder?

frame.js
```diff
@@ -6,6 +6,7 @@
 Array.prototype.push2 = function(value) {
     this.push(value);
     this.push(null);
+    return value;
```
marco-c: Why does push2 return *value*?

midp/gfx.js
```diff
@@ -187,15 +189,16 @@ var currentlyFocusedTextEditor;
...
                 resolve();
             }
-            img.onerror = function(e) {
-                URL.revokeObjectURL(img.src);
-                img.src = '';
-                reject(new JavaException("java/lang/IllegalArgumentException", "error decoding image"));
-            }
-        });
-    }, true);
+            //img.onerror = function(e) {
```
marco-c: Probably forgot to uncomment this one. We should add a test for a image decoding error

frame.js
brendandahl: This was part of an old revision. I'll wipe it out.

midp/gfx.js
brendandahl: I'll uncomment.

marco-c:
I've downloaded a recent version of SpiderMonkey (JavaScript-C37.0a1), but I'm still seeing the same error on *make aot*.

Since there are so many changes, it'd be great if you could provide an overview (for example I've noticed some files have just been ported from JavaScript to TypeScript, it's probably less important to review those compared to the other changes).

classfile/reader.ts
mbebenita: It's similar but it doesn't behave the same. I optimized it because it was showing up in profiles.

1. Avoids temporary string allocation by using a string array and then calling join.
2. It attempts to avoid recompilation in the JS engine caused by setting `array.length`. 
3. It's optimized for use in the reader rather than from an array view.

We could change the one in utils to use this somehow.

classfile/reader.ts
mbebenita: I'm pretty sure it is, the threshold was chosen somewhat arbitrarily though. Let me measure it better.

frame.js
mbebenita: This is already fixed in my latest branch.

classfile/reader.ts
mbebenita: The difference between readStringFast and readStringSlow is considerable. 

https://docs.google.com/spreadsheets/d/1u6BWy6a-OcTbFl0arW7PMo5vVKbIwKp_qNmShEugk1A/edit?usp=sharing

We should do away with the text decoder, since the perf is so bad and also it requires the allocation of a new buffer.

andreasgal: Please file a gecko bug and cc bz.

marco-c: Seems like the difference between the two is decreasing when the string length increases, could you also test with longer strings? (if we never hit long strings when loading classes, then just remove *readStringSlow*)

@EricRahm introduced the TextDecoder version after determining it was faster, so maybe it is faster in some cases.

mbebenita: I was testing in the shell where we apparently use a polyfill because `TextDecoder` is not available. Measuring in the browser is a different story. The sweet spot there is around 40 bytes (for strings that don't contain 2 or 3 byte code points.)

http://jsbin.com/sefiyubose/1/edit?html,js,output

```
Read: 0 bytes, x faster in JS vs TextDecoder: 2.12 (3.58 / 1.69)
Read: 1 bytes, x faster in JS vs TextDecoder: 4.79 (5.42 / 1.13)
Read: 2 bytes, x faster in JS vs TextDecoder: 6.04 (3.23 / 0.53)
Read: 3 bytes, x faster in JS vs TextDecoder: 4.59 (2.66 / 0.58)
Read: 4 bytes, x faster in JS vs TextDecoder: 4.44 (2.89 / 0.65)
Read: 5 bytes, x faster in JS vs TextDecoder: 3.55 (2.50 / 0.70)
Read: 6 bytes, x faster in JS vs TextDecoder: 3.73 (2.86 / 0.77)
Read: 7 bytes, x faster in JS vs TextDecoder: 3.29 (2.64 / 0.80)
Read: 8 bytes, x faster in JS vs TextDecoder: 5.09 (4.95 / 0.97)
Read: 9 bytes, x faster in JS vs TextDecoder: 2.91 (2.66 / 0.91)
Read: 10 bytes, x faster in JS vs TextDecoder: 2.53 (2.53 / 1.00)
Read: 11 bytes, x faster in JS vs TextDecoder: 3.75 (3.92 / 1.05)
Read: 12 bytes, x faster in JS vs TextDecoder: 2.22 (2.65 / 1.19)
Read: 13 bytes, x faster in JS vs TextDecoder: 2.17 (2.69 / 1.24)
Read: 14 bytes, x faster in JS vs TextDecoder: 3.08 (4.12 / 1.34)
Read: 15 bytes, x faster in JS vs TextDecoder: 1.89 (2.63 / 1.39)
Read: 16 bytes, x faster in JS vs TextDecoder: 1.82 (2.54 / 1.40)
Read: 17 bytes, x faster in JS vs TextDecoder: 1.77 (2.57 / 1.45)
Read: 18 bytes, x faster in JS vs TextDecoder: 2.31 (3.72 / 1.61)
Read: 19 bytes, x faster in JS vs TextDecoder: 1.72 (2.59 / 1.50)
Read: 20 bytes, x faster in JS vs TextDecoder: 1.60 (2.64 / 1.65)
Read: 21 bytes, x faster in JS vs TextDecoder: 3.11 (5.43 / 1.74)
Read: 22 bytes, x faster in JS vs TextDecoder: 1.49 (3.52 / 2.36)
Read: 23 bytes, x faster in JS vs TextDecoder: 1.45 (2.89 / 2.00)
Read: 24 bytes, x faster in JS vs TextDecoder: 1.93 (4.29 / 2.23)
Read: 25 bytes, x faster in JS vs TextDecoder: 1.40 (3.12 / 2.24)
Read: 26 bytes, x faster in JS vs TextDecoder: 1.35 (3.10 / 2.29)
Read: 27 bytes, x faster in JS vs TextDecoder: 1.87 (4.27 / 2.29)
Read: 28 bytes, x faster in JS vs TextDecoder: 1.29 (3.17 / 2.47)
Read: 29 bytes, x faster in JS vs TextDecoder: 1.32 (3.22 / 2.43)
Read: 30 bytes, x faster in JS vs TextDecoder: 1.26 (3.17 / 2.52)
Read: 31 bytes, x faster in JS vs TextDecoder: 1.80 (4.54 / 2.53)
Read: 32 bytes, x faster in JS vs TextDecoder: 1.11 (3.01 / 2.70)
Read: 33 bytes, x faster in JS vs TextDecoder: 1.21 (3.02 / 2.50)
Read: 34 bytes, x faster in JS vs TextDecoder: 1.67 (4.48 / 2.67)
Read: 35 bytes, x faster in JS vs TextDecoder: 1.20 (3.26 / 2.71)
Read: 36 bytes, x faster in JS vs TextDecoder: 1.13 (3.13 / 2.77)
Read: 37 bytes, x faster in JS vs TextDecoder: 1.59 (4.45 / 2.80)
Read: 38 bytes, x faster in JS vs TextDecoder: 1.12 (3.20 / 2.85)
Read: 39 bytes, x faster in JS vs TextDecoder: 1.10 (3.36 / 3.06)
Read: 40 bytes, x faster in JS vs TextDecoder: 1.49 (4.41 / 2.96)
Read: 41 bytes, x faster in JS vs TextDecoder: 1.14 (3.40 / 2.99)
Read: 42 bytes, x faster in JS vs TextDecoder: 1.08 (3.40 / 3.15)
Read: 43 bytes, x faster in JS vs TextDecoder: 1.04 (3.32 / 3.18)
Read: 44 bytes, x faster in JS vs TextDecoder: 1.39 (4.54 / 3.27)
Read: 45 bytes, x faster in JS vs TextDecoder: 1.00 (3.21 / 3.21)
Read: 46 bytes, x faster in JS vs TextDecoder: 0.96 (3.40 / 3.53)
Read: 47 bytes, x faster in JS vs TextDecoder: 1.79 (6.20 / 3.46)
Read: 48 bytes, x faster in JS vs TextDecoder: 0.77 (3.67 / 4.74) 
```

mykmelez:
This passes all tests on my local machine but hangs on Travis. It also hangs on a mobile device when testing manually with a midlet. I don't know of any changes on master that would address this issue, but in any case I've merged from master over in my branch of this branch (https://github.com/mykmelez/j2me.js/tree/brendandahl-jit-merge), and we should test that on Travis and with devices before we dig deeper to find the problem.

brendandahl:
Today I can't seem to reproduce the hangs.  I've run it 5 times, got 3 fails on the textbox intermittent failure and two passes.

yxl:
I expect this to be merge soon. So that we could do further improvement base on this, and fix hidden bugs introduced by this huge pull request :-)

mykmelez:
> Since there are so many changes, it'd be great if you could provide an overview (for example I've noticed some files have just been ported from JavaScript to TypeScript, it's probably less important to review those compared to the other changes).

@brendandahl Perhaps you could provide this overview?

brendandahl:
> @brendandahl Perhaps you could provide this overview?

Sure, after lunch/meeting I'll try to do it.

brendandahl:
Here are the main changes per file. Mbx is currently updating https://github.com/andreasgal/j2me.js/wiki/JIT---VM-API---DRAFT for a more general overview.

**Makefile**
New targets:
j2me: compiles all the type script files
aot: (Ahead Of Time) compiles classes and program.jar

**actors.ts**
Metadata readers for the class files. This is more or less the old classinfo.js.

**bench/***
Scimark2 benchmarks. Run `make -C bench` to build.

**bindings.ts**
This will be the new way to do native functions. Gives us the ability to use typescript to do the type checking for native functions. TODO: We should probably put this in it's own folder and break everything out by class when we move natives to this format.

**blackBox.js**
Functions that make debugging easier by blackboxing. TODO: remove throwPause

**bytecodes.ts**
Code for reading bytecode streams and meta data about the opcodes. Similar to the old opcodes.js file. TODO: Remove bytecode reading in Frame and use this in the interpreter instead.

**classRegistry.ts**
Loads jar files. Basically the old classes.js.

**classfile/***
Ports to typescript.

**context.js**
- Since we now have to go in between interpreter, compiled, and native frames the new idea of "frame sets" is introduced to help group interpreter frames. A new frame set is created on every method call, but it is possible for a frame set to have multiple interpreter frames. This allows us to separate the frames to keep the frame order when we bailout and start creating interpreter frames to replace the compiled frames. TODO: This may not be necessary and I'll file a follow up issues to look into removing this. 
- There's a new method `executeNewFrameSet` to help with executing a frame synchronously and getting the result.
- Reduced the execute methods so that there is only start() and resume() that may be publicly called.

**interpreter.ts**
- Mainly a port to typescript and adapting things.

**java/***
Removing synchronization code (may or may not be bad according to mbx).

**jit/blockMap.ts**
Splits bytecodes into basic blocks. A block is a continuous stretch of code where the branch is always at the end.

**jit/builder.ts**
Builds an intermediate representation(IR) for the C4 backend to compile.

**jit/c4/***
Shumway files for compiling the IR.

**jit/compiler.ts**
Reads in class data and compiles each method with builder.ts.

**jit/jit.ts**
TODO: Needs to be moved into another file.

**jit/jvm.ir.ts**
IR nodes that are specific to the JVM. Other IR nodes can be found in c4/backend.ts.

**jsc.ts**
Shell helper script for compiling methods.

**jvm.ts**
Mainly a port to TS.

**metrics.ts**
Helpers for profiling and other metrics.

**options.ts**
Option parser for jsc.ts.

**references.ts**
Links all the TS files together required for j2me.js.

**runtime.ts**
Port of the old runtime.js with additions.  Runtime.ts is now responsible for creating the JS representation of java objects and also performing all of the linking of a classes' methods to the JS object prototype (those methods may be native, interpreted, or compiled). There are also a number of helper methods for compiled code.

**shumway/***
Code for the new profiler.

**native.js, override.js, midp/***
- Removed wrappers so we get direct calls in new compiled code.
- Updated how objects/arrays/strings are created to use the new functions in runtime.ts.
- Updated how exceptions are created/thrown.
- Use the new asyncImpl method for async functions. TODO: we should maybe make some helper functions for executing async code, because it is very easy to trigger code that needs to have the current runtime/context set.
- TODO: remove all getField stuff and just directly read/write object properties.

mbebenita:
jit/builder.ts
TODO: Move out the `canYield` code and figure out a better way to manage the list of async methods.

mykmelez:
Overall, I'm pretty happy with this.  There's a ton more to do, not the least of which is to get compilation working. But the interpreter works correctly, from what I can tell, in both automated and manual testing, so we should land it and start iterating on the perf improvements it enables.

Here are two more TODOs:

* The `make app` target needs updating, as it no longer creates a working output/ subdirectory.
* *README.md* needs additional updating, as it still talks about *Native.create* and *Override.create*.

And a few nits:

A couple dozen files are missing newlines at their ends.

There are a bunch of unnecessary changes to comments in the new customs *String.java*, *StringBuffer.java*, *Hashtable.java*, *Vector.java*, and *Connector.java*. Mostly they're changes to class references (making them absolute, or removing absolute paths from them), but there's also this global search/replace snafu in *Vector.java*:

> -     * This implements the unsynchronized semantics of ensureCapacity.
> +     * This implements the unsemantics of ensureCapacity.

These customs should contain the minimum necessary changes from their originals, so comparisons are easier to read.

Otherwise, the only non-synchronized change in these customs is in *Connector.java*, where some code in *openPrim* has been refactored into a *checkProtocolName* method that has strange formatting (too much indentation in the conditionals) and no obvious change. Its formatting should be cleaned up, and it should get a comment explaining why it's been refactored.

Finally, *RunTestsShell.java* looks unused (and likely to become out-of-date, given its hardcoded list of classes). Is it useful?

brendandahl:
Nits should be taken care of.

marco-c:
I'm still seeing an error while doing *make aot*:

    js build/jsc.js -cp java/classes.jar -d -jf java/classes.jar -cff classes.txt > build/classes.jar.js
    .../2me.js/instrument.js:14:2 ReferenceError: config is not defined

marco-c:
The *make closure* target doesn't work because there's no tools/closure.jar. I think we should either include it in the repo or make the makefile download it if needed or add an explanation to the readme.

marco-c:
+aot: java j2me
+	js build/jsc.js -cp java/classes.jar -d -jf java/classes.jar -cff classes.txt > build/classes.jar.js
+	js build/jsc.js -cp java/classes.jar program.jar -d -jf program.jar -cff classes.txt > build/program.jar.js

What is "program.jar"? I guess it's the MIDlet that we want to run? If it is, we can't always compile it ahead of time, so we should remove this.

The js executable in many Linux distributions is a system executable (for example, on Fedora there's the package "js-1.8.5-19" that contains SpiderMonkey 1.8.5 that is installed by default). We should explain how to run *make aot* in the readme, pointing out that you need to download a recent version of SpiderMonkey and use it instead of the version provided by the OS.

marco-c:
+  export class PrimitiveClassInfo extends ClassInfo {
    +    constructor(className: string, mangledName: string) {
    +      false && super(null);

Nit: could we use easier to read alternatives here and in other cases?

marco-c:
+  export class BytecodeSwitch {
+//    /**
+//     * The {@link BytecodeStream} containing bytecode array or {@code null} if {@link #code} is not {@code null}.
+//     */
+//    private final BytecodeStream stream;
+//

Nit: I guess we can remove this.

+    /**
+     * Gets the key at {@code i}'th switch target index.
+     * @param i the switch target index
+     * @return the key at {@code i}'th switch target index
+     */
+    public keyAt(i: number): number {
+      throw Debug.abstractMethod("defaultOffset");
+    }

Wrong message here and in the following methods.

marco-c:
<button id="clearstorage">Clear storage</button>
    +      <button id="deleteDatabase">Delete Database</button>

We probably don't need both buttons. Is there a reason why you needed to completely remove the DB? If it's needed, then we can just remove the "Clear storage" button and only leave the "Delete database".

marco-c:
+  <script type="text/javascript" src="build/j2me.js" defer></script>
    +  

Instead of this, we could make the closure compiler overwrite the files under build/. This way we'd use the normal file when you don't run *make closure*, the optimized file if you run *make closure*

marco-c:
+Promise.all(loadingPromises).then(function() {
    +  setTimeout(function () {
    +    start();
    +  }, 1000);
    +});

Why is this 1 second timeout needed?

marco-c:
-Override.create("com/ibm/oti/connection/file/Connection.decode.(Ljava/lang/String;)Ljava/lang/String;", function(string) {
    -    return decodeURIComponent(util.fromJavaString(string));
    -});
    -
    -Override.create("com/ibm/oti/connection/file/Connection.encode.(Ljava/lang/String;)Ljava/lang/String;", function(string) {
    -    return util.fromJavaString(string).replace(/[^a-zA-Z0-9-_\.!~\*\\'()/:]/g, encodeURIComponent);
    -});

Why have you removed these overrides?

marco-c:
+Native["com/ibm/oti/connection/file/Connection.deleteFileImpl.([B)Z"] = function(path) {
+    asyncImpl("Z", new Promise(function(resolve, reject) {
+        fs.remove(getAbsolutePath(path), function(x) { resolve(x ? 1 : 0); });
+    }));
+};

Nit: replace x with something more meaningful, like "removed"

marco-c:
In style/main.css:
    +<<<<<<< HEAD
    +=======
    +.debug-mode .splash-screen {

marco-c:
+//# sourceMappingURL=file:////Users/mbebenita/Workspaces/MSTS/built/local/tsc.js.map

This probably needs to be changed

marco-c:
+  // Remove
    +  function newArray(classInfo, size) {
    +    return J2ME.newArray(classInfo.klass, size);
       }

Can we remove this and the other functions that are just calling J2ME functions?

marco-c:
function newPrimitiveArray(type, size) {
    -    var constructor = ARRAYS[type];
    -    if (!constructor.prototype.class)
    -      CLASSES.initPrimitiveArrayType(type, constructor);
    +    // !!! ask mbx should this be using newArray instead?
    +    var constructor = J2ME.getArrayConstructor(type);
         return new constructor(size);
       }

Do you need to fix this before the merge or do you plan to do it after?

marco-c:
Looks like *utilities.ts* is coming from shumway, can we move it under the shumway directory?

marco-c:
We shouldn't have two implementations of javaUTF8Decode (in util.js and reader.ts)

marco-c:
If the files under jit/build/ are build products, we shouldn't include them in the repo.

marco-c:
diff --git a/compatibility.js b/compatibility.js
    new file mode 100644

We already have a *legacy.js* file, could you add these functions in legacy.js instead 
of creating a new file?

marco-c:
+    "java/lang/Object": {
    +      native: {
    +        "hashCode.()I": function (): number {
    +          var self: J2ME.java.lang.Object = this;
    +          if (self._hashCode) {
    +            return self._hashCode;
    +          }
    +          return self._hashCode = $.nextHashCode();
    +        }
    +      }
    +    }

Out of curiosity, is there a "native" key because we want to distinguish between natives and overrides?

marco-c:
+Native["gnu/testlet/vm/NativeTest.getLongReturnInt.(J)I"] = function(val) {
    +  return ~~(40 + val.toNumber());
    +};

Do we need to do the cohercion in real natives as well?
IIRC there was an issue related to this that myk filed a while ago (that has been fixed since then)

marco-c:
+    // if (Instrument.profiling) {
    +    //   Instrument.exitAsyncNative(key, promise);
    +    // }

Can we re-enable profiling of async natives?

marco-c:
+Native["com/sun/j2me/content/AppProxy.midletIsAdded.(ILjava/lang/String;)V"] = function(suiteId, className) {
    +  // ???
    +};

Could you add the usual "not implemented" warning here?

marco-c:
+Native["java/io/DataInputStream.bytesToUTF.([B)Ljava/lang/String;"] = function(bytearr) {

We should do in this native the same thing we've decided to do in reader.ts

marco-c:
-        ctx.runtime.setStatic(CLASSES.getField(classInfo, "S._API_access_ok.I"), 1);
    +        // ctx.runtime.setStatic(CLASSES.getField(classInfo, "S._API_access_ok.I"), 1);
    +        var isolate = classInfo.getStaticObject($.ctx);
    +        CLASSES.getField(classInfo, "S._API_access_ok.I").set(isolate, 1);

Nit: remove the commented out line

marco-c:
MIDP.copyEvent = function(obj, isolateId) {
         var e = MIDP.nativeEventQueues[isolateId].shift();
    -    obj.class.getField("I.type.I").set(obj, e.type);
    -    obj.class.fields.forEach(function(field) {
    +    obj.klass.classInfo.getField("I.type.I").set(obj, e.type);
    +    obj.klass.classInfo.fields.forEach(function(field) {
    +        if (e[field.name] === undefined) {
    +          return;
    +        }
             field.set(obj, e[field.name]);
         });
     }

Why is this needed?

mykmelez:
> I'm still seeing an error while doing make aot:

Fixed in d0b13b3f99d7605e28b82447fcff7a9baac936d3 (part of https://github.com/brendandahl/j2me.js/pull/7).

marco-c:
P.S.: These PRs and issues are related to int coercion:
https://github.com/andreasgal/j2me.js/pull/605
https://github.com/andreasgal/j2me.js/pull/604
https://github.com/andreasgal/j2me.js/pull/266

IIRC, one of the Media classes was failing because we weren't coercing values

mykmelez:
> The make closure target doesn't work because there's no tools/closure.jar. I think we should either include it in the repo or make the makefile download it if needed or add an explanation to the readme.

I added an explanation to the README in bacae3d0f466e1f5b9da1ff2c062b081499a65f4 (part of brendandahl#7).

TODO: make the Make target download it if needed.

mykmelez:
> What is "program.jar"? I guess it's the MIDlet that we want to run? If it is, we can't always compile it ahead of time, so we should remove this.
> 
> The js executable in many Linux distributions is a system executable (for example, on Fedora there's the package "js-1.8.5-19" that contains SpiderMonkey 1.8.5 that is installed by default). We should explain how to run make aot in the readme, pointing out that you need to download a recent version of SpiderMonkey and use it instead of the version provided by the OS.

I think that's what program.jar is, but unsure, so I made its compilation conditional on its existence and explained that you need a recent version of the JavaScript shell to run `make aot` in 153bf6a and 5dd28c5.

mykmelez:
> Nit: I guess we can remove this.
…
> Wrong message here and in the following methods.

Both fixed in ff4b044.

brendandahl:
> MIDP.copyEvent = function(obj, isolateId) {
>      var e = MIDP.nativeEventQueues[isolateId].shift();
> -    obj.class.getField("I.type.I").set(obj, e.type);
> -    obj.class.fields.forEach(function(field) {
> +    obj.klass.classInfo.getField("I.type.I").set(obj, e.type);
> +    obj.klass.classInfo.fields.forEach(function(field) {
> +        if (e[field.name] === undefined) {
> +          return;
> +        }
>          field.set(obj, e[field.name]);
>      });
>  }
> 
> Why is this needed?
This avoids introducing `undefined` values into the js object which is not allowed in the new code.

mykmelez:
> We probably don't need both buttons. Is there a reason why you needed to completely remove the DB? If it's needed, then we can just remove the "Clear storage" button and only leave the "Delete database".

Completely removing the database is handy when it gets into an inconsistent state. And it covers the "clear" case as well. So I've removed the "clear storage" button in 5856f1a.

marco-c:
-Native.create("java/lang/Class.forName.(Ljava/lang/String;)Ljava/lang/Class;", function(name, ctx) {
    +Native["java/lang/Class.forName.(Ljava/lang/String;)Ljava/lang/Class;"] = function(name) {
         try {
             if (!name)
    -            throw new Classes.ClassNotFoundException();
    +            throw new J2ME.ClassNotFoundException();

Shouldn't this be *throw $.newClassNotFoundException() ?

brendandahl:
> +Native["java/io/DataInputStream.bytesToUTF.([B)Ljava/lang/String;"] = function(bytearr) {
> 
> We should do in this native the same thing we've decided to do in reader.ts

Let's mark this as a TODO as it's an optimization.

mykmelez:
> Instead of this, we could make the closure compiler overwrite the files under build/. This way we'd use the normal file when you don't run make closure, the optimized file if you run make closure

Done in 4c7be46, where I also added its dependencies to the `make closure` target and added a conditional check for the existence of build/program.jar.js (like the check I added earlier to `make aot` for the existence of program.jar).

marco-c:
> Let's mark this as a TODO as it's an optimization.

Or even better, let's file an issue about this.

marco-c:
Done in 4c7be46, where I also added its dependencies to the make closure target and added a conditional check for the existence of build/program.jar.js (like the check I added earlier to make aot for the existence of program.jar).

There are a few other instances in the same file:

mykmelez:
>> Let's mark this as a TODO as it's an optimization.
> Or even better, let's file an issue about this.

I expect all of the TODOs we've identified in this PR to become either issues or PRs!

brendandahl:
> Can we re-enable profiling of async natives?
I'm going to punt on this as well. There's not currently a way to get the async function's name. I'd like to change this when we rework async's as mentioned in a previous TODO.

mykmelez:
> There are a few other instances in the same file:
> 
> 
> 
> 

Hunh, I thought I checked for these and didn't find them. Oh well, removed in cc8e203. :-)

mbebenita:
> Looks like utilities.ts is coming from shumway, can we move it under the shumway directory?

`utilities.ts` comes from Shumway but we shouldn't move it to the shumway directory. Shumway is only used for the profiler, and should not be part of the build. In fact, the shumway directory has a duplicate version of `utilities.ts`.

mykmelez:
> Why is this 1 second timeout needed?

It isn't, @mbebenita was just using it to profile startup. I removed it in b8822c2.

mbebenita:
> Out of curiosity, is there a "native" key because we want to distinguish between natives and overrides?

This is the new bindings mechanism. We didn't want to port everything ever, to reduce merge conflicts. The full format is:
```javascript
"java/lang/Object": {
  native: {
    "hashCode.()I": function (): number {
      var self: J2ME.java.lang.Object = this;
      if (self._hashCode) {
        return self._hashCode;
      }
      return self._hashCode = $.nextHashCode();
    }
  },
  fields: {
    staticSymbols: {
      "javaFieldName": "javaScriptFieldName" 
    },
    instanceSymbols: {
      "javaFieldName": "javaScriptFieldName"
    }
  }
}
```
The `native` section has the native / override methods. The fields section creates getters / setters for java fields so they can be accesses easily from JS code.

mykmelez:
> Nit: replace x with something more meaningful, like "removed"

Fixed (plus added a missing semicolon nearby) in 39cc9f4.

mbebenita:
> Do we need to do the cohercion in real natives as well?
> IIRC there was an issue related to this that myk filed a while ago (that has been fixed since then)

All natives and overrides need to return proper Java values. There are no methods in-between that do the coercion for you.

mykmelez:
> In style/main.css:
> 
> +<<<<<<< HEAD
> +=======
> +.debug-mode .splash-screen {

Fixed in a73d191.

mbebenita:
> Can we re-enable profiling of async natives?

Instrumentation has changed and we haven't really kept the old system running. We can re-enable the functionality in the future.

brendandahl:
> We shouldn't have two implementations of javaUTF8Decode (in util.js and reader.ts)

Let's combine this work with the change for fixing "java/io/DataInputStream.bytesToUTF.([B)Ljava/lang/String;"

brendandahl:
> Shouldn't this be *throw $.newClassNotFoundException() ?

This is somewhat confusing and we should maybe rename. The ClassNotFoundException is different than the actual java ClassNotFoundException. The classRegistry throws this exception when a class isn't found.  We don't want to use the actual exception since the compiler needs to use this code without a runtime($) defined.

brendandahl:
> function newPrimitiveArray(type, size) {
> -    var constructor = ARRAYS[type];
> -    if (!constructor.prototype.class)
> -      CLASSES.initPrimitiveArrayType(type, constructor);
> +    // !!! ask mbx should this be using newArray instead?
> +    var constructor = J2ME.getArrayConstructor(type);
>      return new constructor(size);
>    }
> 
> Do you need to fix this before the merge or do you plan to do it after?

Let's TODO this. I started changing it but we really need to cleanup/remove all the utilities newArray functions. It's a lot of code churn, so I'd rather do it in a different PR.

mykmelez:
> +//# sourceMappingURL=file:////Users/mbebenita/Workspaces/MSTS/built/local/tsc.js.map
> 
> This probably needs to be changed

This copy of tsc.js appears to be compiled from a fork of TypeScript, probably https://github.com/mbebenita/shumway.ts. We should remove it in favor of downloading/installing a copy from a canonical source.

In theory, that is just `npm install -g typescript`, but I get "jit/c4/optimizer.ts(295,50): error TS2339: Property 'project' does not exist on type 'Node'." when I try to compile with the one in Node. Perhaps ours is from https://github.com/mbebenita/shumway.ts (@mbebenita should know).

In any case, let's add a TODO to replace this file with one that we download/install from a suitable location.

mbebenita:
> In any case, let's add a TODO to replace this file with one that we download/install from a suitable location.

TS changes over time, we'll need a TODO to make sure everything builds with their latest release.

brendandahl:
Follow up issues: #853, #854, #855, #856, #857, #858, #859, #860, #861, #862, #863, #864

PR #811, opened 2015-01-02
Add Nokia-specific Launcher gfx test

marco-c:


tests/gfx/TestLauncher.java
```diff
@@ -0,0 +1,55 @@
...
+            length += l;
+            if (length == buffer.length) {
+                byte[] b = new byte[buffer.length + 4096];
+                System.arraycopy(buffer, 0, b, 0, length);
+                buffer = b;
+            }
+        }
+
+        return buffer;
+    }
```
mykmelez: (Note to self: the next time I see one of these *read(InputStream)* methods, or similar methods like the *read(String)* method in #794, I should request that they get refactored into a utilities class. ;-) )

PR #813, opened 2015-01-03
Move some test-specific functions in gnu.testlet.TestUtils to avoid code...

marco-c:
... duplication


tests/gnu/testlet/TestUtils.java
```diff
@@ -0,0 +1,91 @@
...
+
+        while ((l = is.read(buffer, length, buffer.length - length)) != -1) {
+            length += l;
+            if (length == buffer.length) {
+                byte[] b = new byte[buffer.length + 4096];
+                System.arraycopy(buffer, 0, b, 0, length);
+                buffer = b;
+            }
+        }
+
```
yxl: There is a bug with `read()` that we doesn't remove the unused bytes if the buffer is large that needed. It could be fixed by add something like this:
```Java
if (length < buffer.length) {
  byte[] b = new byte[length];
  System.arraycopy(buffer, 0, b, 0, length);
  buffer = b;
}
```

tests/gnu/testlet/TestUtils.java
```diff
@@ -0,0 +1,91 @@
...
+            // skip the space
+            start = (end + 1);
+        } while (end != inputString.length());
+
+        return sb.toString();
+    }
+
+    public static String getEmojiString(String code) {
+        return getSurrogatePairs(checkCodeFormat(code));
```
yxl: `getSurrogatePairs()` and `checkCodeFormat()` are called by `getEmojiString` only, so you don't need to expose them as public methods.

tests/gnu/testlet/TestUtils.java
marco-c: Done!

tests/gnu/testlet/TestUtils.java
marco-c: Done!

marco-c:
I've also made the new TestAudioRecorder test use TestUtils

PR #818, opened 2015-01-05
Return 0 in nSetLocator, to let DirectRecord handle writing the recordin...

marco-c:
...g to a file


marco-c:
The behavior of the DirectRecord class depends on the value returned by nSetLocator:
1) -1 - Throws an exception
2) 0 - Handles writing to a file / uploading in the class itself
3) Everything else - Lets the native handle writing/uploading

PR #819, opened 2015-01-05
Replace the Java LF implementation with a native one

marco-c:


java/Makefile
```diff
@@ -1,5 +1,5 @@
 SRCS=$(shell find ./cldc1.1.1 -name *.java) $(shell find ./vm -name *.java) $(shell find ./midp -name *.java) $(shell find ./jsr-075 -name *.java) $(shell find ./custom -name *.java)
-JPP_DEFS=-DENABLE_JSR_205 -DENABLE_CHAMELEON -DENABLE_SSL -DENABLE_PUBLICKEYSTORE -DENABLE_JSR_211 -DENABLE_MULTIPLE_ISOLATES -DRECORD -DUSE_FILE_CONNECTION -DENABLE_JSR_234
```
TimAbraldes: Sweet

java/midp/javax/microedition/lcdui/CanvasLFImpl.java
```diff
@@ -159,112 +151,79 @@ public void uCallHide() {
...
-                Display.handleThrowable(t);
             }
+        } catch (Throwable t) {
+            Display.handleThrowable(t);
         }
 
-        g.restoreMIDPRuntimeGC();
+        // If there are any video players in this canvas,
+        // let the helper class invoke video rendering
+	// Update frames of any video players displayed on this Canvas
```
andreasgal: Identation.

java/midp/javax/microedition/lcdui/CanvasLFImpl.java
```diff
@@ -159,112 +151,79 @@ public void uCallHide() {
...
+	if (mmHelper != null) {
+            for (Enumeration e = embeddedVideos.elements();
+                                            e.hasMoreElements();) {
+		mmHelper.paintVideo(e.nextElement(), g);
+            }
+	}
+//        g.restoreMIDPRuntimeGC();
```
andreasgal: Dead comment.

java/midp/javax/microedition/lcdui/CanvasLFImpl.java
```diff
@@ -159,112 +151,79 @@ public void uCallHide() {
...
-        }
+        super.uCallPaint(g, target);
 
         // We prevent the Canvas from drawing outside of the
-        // allowable viewport - such as over the command labels
-        // or over the theme area.
+        // allowable viewport.
         // We also need to preserve the original translation.
-        g.preserveMIDPRuntimeGC(0, 0, viewport[WIDTH], viewport[HEIGHT]);
+//        g.preserveMIDPRuntimeGC(x, y, WIDTH, HEIGHT);
```
andreasgal: dead code

TimAbraldes:
I'm not totally sure how to complete this review because github refuses to show the whole diff. I can email you comments but if there's some way to complete this in github it would probably be better. Anyone have ideas on this?

java/midp/javax/microedition/lcdui/CanvasLFImpl.java
marco-c: I haven't modified the code under java/midp/ because it's from the phoneme RI.

marco-c:
> I'm not totally sure how to complete this review because github refuses to show the whole diff. I can email you comments but if there's some way to complete this in github it would probably be better. Anyone have ideas on this?

I have the same problem with the JIT PR... If nobody knows how to avoid this problem, then an email is fine (or maybe you could leave comments here, copying the relevant part of the diff)

mykmelez:
You can add `.diff` to the URL to get a raw diff:

https://github.com/andreasgal/j2me.js/pull/819.diff

Of course you don't get colorizing or built-in commenting, but you can save locally and open in a colorizing text editor to get the former, then copy/paste snippets and comment on them here. I think that's the best you can do, as GitHub intentionally limits how much of a diff it'll show.

TimAbraldes:
When I run this branch and get to the "do you want to send a verification code" alert I see this:
    Attempted to invoke missing native: javax/microedition/lcdui/ItemLFImpl.hide0.(I)V

It seems like we should implement that

TimAbraldes:
diff --git a/main.html b/main.html
index d57ed0e..d9fd7f7 100644
--- a/main.html
+++ b/main.html
@@ -111,6 +111,19 @@
   </div>
 
   <form role="dialog" data-type="confirm"
+        class="lcdui-alert"
+        id="lcdui-alert" style="display: none">
+    <section>
+      <h1 class="title"></h1>
+      <p class="text"></p>
+      <progress style="display: none"></progress>
+    </section>
+    <menu>
+      <button class="recommend">Done</button>
+    </menu>
+  </form>
+
+  <form role="dialog"

This looks good but we'll have to allow for more buttons to exist: See the next comment.

Also, when I run this I get vertical scroll bars on the "do you want to send a verification SMS" Alert.

+    Native.create("javax/microedition/lcdui/NativeMenu.updateCommands.([Ljavax/microedition/lcdui/Command;I[Ljavax/microedition/lcdui/Command;I)V",
+    function(itemCommands, numItemCommands, commands, numCommands) {
+        console.warn("javax/microedition/lcdui/NativeMenu.updateCommands.([Ljavax/microedition/lcdui/Command;I[Ljavax/microedition/lcdui/Command;I)V");
+    });

We must implement this in order for `Alert`s to work properly. For each `Command` in `itemCommands` and in `commands`, we have to make a button available with the correct label and command id. When a button is pressed, we need to send a native `MIDP.COMMAND_EVENT` - something like this:
    MIDP.sendNativeEvent({ type: MIDP.COMMAND_EVENT, intParam1: pressedButton.cmdId }, MIDP.foregroundIsolateId);

TimAbraldes:
Overall this looks great!

It doesn't incorporate the "one canvas per Displayable" idea but we can always investigate adding that later. That approach might run into correctness issues anyway since the only way to get performance benefits is to not call a Displayable's repaint methods when it becomes visible, breaking assumptions that apps might have made.

Also, we've discussed this outside of the issue tracker, but it's important to note that we're likely breaking all Displayables other than Canvas and Alert. We can implement the other Displayable types natively over time, though, so I think this is a step in the right direction.

I'm really excited to see us rip out chameleon

marco-c:
Oh, right! I totally forgot Alerts may have buttons!

andreasgal:
What's the perf impact of this? And startup?

marco-c:
> What's the perf impact of this? And startup?

Unsure about perf, since many MIDlets draw their own UI elements using a Canvas. This patch does remove some unneeded computations related to Chameleon, but I'm not sure what their impact was.
The only j2me elements some MIDlets use are Alerts, their performance is much better now (we had to override one of the Alerts function before, because it was too slow on mobile!).

There's also a lot of Java code removed, indeed the classes.jar file is now 2,6 MB (it was 3,2 MB before).

marco-c:
@TimAbraldes I've added support for Commands (max 2 commands, since IIRC this is the maximum I've seen used) with an automated test. I haven't tested it yet on a real world MIDlet, will do so tomorrow.

TimAbraldes:
@marco-c a 2 command max should get us most of the way there - I think most apps use `Alert`s only in OK/Cancel situations. Some day it would be nice to make the secondary button open a menu with subcommands; that's how MIDlets expect to be displayed.

marco-c:
I've literally spent hours investigating a bug with the SMS listener dialog, but the issue was the stupidest one: https://github.com/marco-c/j2me.js/commit/f12c8fbeb5f6a1731493098b2ed7e325d9fbd96b

marco-c:
The latest commit changes the behavior of the SMS dialog: we aren't focusing the input element anymore (it made the dialog not completely visible due to the virtual keyboard, and forced users to unfocus it and refocus it after reading the instructions).

TimAbraldes:
+    Native.create("javax/microedition/lcdui/DisplayableLFImpl.initialize0.()V", function() {
    +        console.warn("javax/microedition/lcdui/DisplayableLFImpl.initialize0.()V not implemented");
    +    });

Even if we eventually implement all the Displayable types natively, we probably won't do anything in this function, right? Can we just remove the warning?

    +    Native.create("javax/microedition/lcdui/DisplayableLFImpl.deleteNativeResource0.(I)V", function(nativeId) {
    +        console.warn("javax/microedition/lcdui/DisplayableLFImpl.deleteNativeResource0.(I)V not implemented");
    +
    +        var el = document.getElementById("displayable-" + nativeId);
    +        if (el) {
    +            el.parentElement.removeChild(el);
    +            if (currentlyFocusedTextEditor) {
    +                currentlyFocusedTextEditor.focus();
    +            }
    +        }
    +    });

Maybe we should warn only in the case that we didn't actually do anything?

    +    Native.create("javax/microedition/lcdui/CanvasLFImpl.createNativeResource0.(Ljava/lang/String;Ljava/lang/String;)I", function(title, ticker) {
    +        console.warn("javax/microedition/lcdui/CanvasLFImpl.createNativeResource0.(Ljava/lang/String;Ljava/lang/String;)I not implemented");
    +        curDisplayableId = nextMidpDisplayableId++;
    +        return curDisplayableId;
    +    });

Although we don't technically create a native resource for each CanvasLF, we probably don't want to warn every time we call this function. We've just chosen an alternative abstraction. Let's consider as a follow-up potentially creating individual HTML canvas elements for each CanvasLF, but let's remove the warning in the meantime.

    +    Native.create("javax/microedition/lcdui/NativeMenu.updateCommands.([Ljavax/microedition/lcdui/Command;I[Ljavax/microedition/lcdui/Command;I)V",
    +    function(itemCommands, numItemCommands, commands, numCommands) {
    +        if (numItemCommands != 0) {
    +            console.error("NativeMenu.updateCommands: item commands not yet supported");
    +        }

Let's file an issue to support item commands

    +
    +        if (numCommands > 2) {
    +            console.error("NativeMenu.updateCommands: max two commands supported");
    +        }

Let's file an issue to support more commands

    --- a/midp/sms.js
    +++ b/midp/sms.js
    @@ -92,7 +92,7 @@ function promptForMessageText() {
                                                      " " + MIDlet.SMSDialogTimeoutText;
     
         document.body.appendChild(el);
    -    input.focus();
    +    currentlyFocusedTextEditor = null;

What happens if the SMS prompt comes up over a CanvasLF that had a TextBox active? Does the keyboard correctly come back up after dismissing the SMS prompt?

    +    .thenOpen("http://localhost:8000/index.html?midletClassName=midlets.TestAlertWithGauge&jars=tests/tests.jar&logConsole=web,page")
    +    .withFrame(0, function() {
    +        this.waitUntilVisible(".lcdui-alert.visible .button1", function() {
    +            this.click(".lcdui-alert.visible .button0");
    +            this.waitForText("You pressed 'Yes'", function() {
    +                test.assertTextDoesntExist("FAIL");
    +
    +                this.click(".lcdui-alert.visible .button1");
    +                this.waitForText("You pressed 'No'", function() {
    +                    test.assertTextDoesntExist("FAIL");
    +                });
    +            });
    +        });
    +    });

Nice, thanks for adding this test.


This looks great! I think we can pretty much land this and then iterate as needed. Could you clean up the nits? I'd also like to test on device in the meantime

marco-c:
> What happens if the SMS prompt comes up over a CanvasLF that had a TextBox active? Does the keyboard correctly come back up after dismissing the SMS prompt?

It doesn't come up. I did this on purpose because I think it would be confusing to automatically focus an editor (in the canvas) after the user has interacted with another editor (in the SMS dialog). After all, in the eyes of the user the SMS dialog is part of the application.

TimAbraldes:
>> What happens if the SMS prompt comes up over a CanvasLF that had a TextBox active? Does the keyboard correctly come back up after dismissing the SMS prompt?
>
> It doesn't come up. I did this on purpose because I think it would be confusing to automatically focus an editor (in the canvas) after the user has interacted with another editor (in the SMS dialog). After all, in the eyes of the user the SMS dialog is part of the application.

Do MIDlets correctly notice that the keyboard has gone away? Perhaps we need to fire a keyboard visibility change event? I haven't seen issues in the MIDlets that I test with, but none of them send an SMS from a CanvasLF. I think we can go ahead and merge and deal with this as a followup if it comes up as an issue

TimAbraldes:
Tested on device, switching between this branch and master a few times. Everything seems to look OK!

marco-c:
@yxl @PinZhang could you test this PR with your MIDlet?

yxl:
Unfortunately, the PR breaks our MIDlet. It doesn't show the title bar and context buttons appearing at the bottom of the original implementation. Our MIDlet depends on these context buttons.

The MediaSampler MIDlet under the `tests/midlets` directory, provides similar UI as our MIDlet's:

MediaSampler

We want to have a title bar showing `Media Sampler` at the top and an `exit` button, which is at the bottom with the original implementation and might be in the title bar with other implementations.

You can run the MediaSampler if you host the j2me.js at http://localhost:8000/
```
http://localhost:8000/index.html?jars=tests/tests.jar&midletClassName=com.nokia.example.mmapi.mediasampler.MediaSamplerMIDlet&jad=tests/midlets/MediaSampler/MediaSampler.jad
```

mykmelez:
This needs a rebase, and AlertTest is failing on my Mac with 1942 different pixels:

alerttest alerttest-mac

TimAbraldes:
The code looks good to me and I've tested locally with at least one MIDlet. Please rebase this on top of the recent jit branch merge

marco-c:
The LauncherTest result is this:
index



Instead of this:
launchertest

marco-c:
The log is too long to be displayed, but tests passed.

PR #825, opened 2015-01-06
Jit merge debug2

brendandahl:


Makefile
```diff
@@ -8,6 +11,28 @@ test: all
...
+	node tools/tsc.js --sourcemap --target ES5 jsc.ts --out build/jsc.js
+
+j2me: build/j2me.js build/jsc.js
+
+aot: java j2me
+	js build/jsc.js -cp java/classes.jar -d -jf java/classes.jar -cff classes.txt > build/classes.jar.js
+	js build/jsc.js -cp java/classes.jar program.jar -d -jf program.jar -cff classes.txt > build/program.jar.js
+
+closure:
+	java -jar tools/closure.jar --language_in ECMASCRIPT5 -O SHUMWAY_OPTIMIZATIONS build/j2me.js > build/j2me.cc.js
```
yxl: `tools/closure.jar` is missing.

brendandahl: Yeah, we haven't decided yet if we're going to keep it in the repo. It's not currently a required build step, so it should be okay. I'm just trying to investigate the intermittent travis failures on the jit branch.

Issue #823, opened 2015-01-06
Correctly handle foreground/background events for multiple MIDlets within a JAR


TimAbraldes:
A JAR can contain multiple MIDlets that will run simultaneously. Each one has a `Display` object and an associated current `Displayable`. Only the foreground MIDlet's current `Displayable` should be visible at any given time.

In our implementation, all running MIDlets draw to the "display" simultaneously, regardless of whether they are in the foreground or background. This causes problems if, for example, the background MIDlet sets its current Displayable (it will draw over the foreground MIDlet).

Our implementation should work correctly with multiple MIDlets in a JAR - drawing only the foreground MIDlet's current Displayable but still keeping track of all required state for each MIDlet (even the background ones).

marco-c:
I'd guess background MIDlets aren't going to use the display, but we should keep this issue in mind in case some MIDlet will decide to do it.

TimAbraldes:
This is the same as #1384. Closing this bug in favor of that one since that has a more accurate description

PR #826, opened 2015-01-06
Implement splash screen 

### LINKED ISSUE #822: Show a splash screen until the MIDlet draws to the screen
Currently we have a long startup time. It would be nice to show a screen saying something like "Please wait while your app loads" and "Powered by J2ME.js".

TimAbraldes:
Fixes #822.

Maybe we can get someone to choose better verbiage. This looks a little wonky on device (the title should probably wrap but it doesn't) but it's basically what I was thinking.


mykmelez:
> NOP change to make travis run again

Instead of adding a space, use `git commit --allow-empty` to add a commit without making any changes.

But you don't even need to do that, since you can restart any build on Travis by pressing the *Restart Build* button in the top-right area of the build page, next to the *Download Log* button. If you don't see that button, then try signing into Travis (creating an account on the site if necessary).

main.html
```diff
@@ -110,6 +110,18 @@
     <button id="fire">fire</button>
   </div>
 
+  <div class="splash-screen" id="splash-screen">
+    <h1 class="title">Please wait while your app loads</h1>
```
mykmelez: I would make this say *Loading [app name]…*.

main.html
```diff
@@ -110,6 +110,18 @@
...
+    <div class="vertical-flex-container">
+      <div class="flex-buffer"></div>
+      <div>
+        <progress></progress>
+      </div>
+      <div class="flex-buffer"></div>
+    </div>
+    <p class="text">Powered by j2me.js</p>
```
mykmelez: I wouldn't display this to users, as it isn't understandable nor useful to them.

midp/gfx.js
```diff
@@ -59,6 +59,7 @@ var currentlyFocusedTextEditor;
 
     Native.create("com/sun/midp/lcdui/DisplayDevice.gainedForeground0.(II)V", function(hardwareId, displayId) {
         console.warn("DisplayDevice.gainedForeground0.(II)V not implemented (" + hardwareId + ", " + displayId + ")");
+        document.getElementById("splash-screen").style.display = "none";
```
mykmelez: Is there anything else to do here? If not, we should remove the "not implemented" warning.

style/main.css
```diff
@@ -148,6 +148,71 @@ button {
...
+.autosize .splash-screen {
+  display: flex;
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+}
+
+.debug-mode .splash-screen {
```
mykmelez: This ruleset should go into the *STYLE OVERRIDES FOR DEBUG MODE* section.

midp/gfx.js
TimAbraldes: I believe we need to modify this function to fix #823, but that doesn't necessarily mean that this warning is useful to us

PR #827, opened 2015-01-07
Fix isLowerCase and isLowerCase bug.

yxl:
j2me divides all characters into 3 groups: 
1. lowercase, such as `a`, `b`, ...
2. uppercase, such as `A`, `B`, ...
3. others, such as `;`, `你`, ...

The characters from the 3rd group are not case sensitive, such as punctuation symbols. They doesn't fall into either the lowercase or uppercase group. So calling `isLowerCase()` and `isUpperCase()` over these characters should always return false.

But commit 76801f1c669a9a7 modified the original methods and made `isLowerCase()` and `isUpperCase()` return true over these case insensitive characters.

I reverted the changes to `isLowerCase()` and `isUpperCase()` to recover the original behavior.

### DIFF
--- a/java/cldc1.1.1/com/sun/cldc/i18n/uclc/DefaultCaseConverter.java
+++ b/java/cldc1.1.1/com/sun/cldc/i18n/uclc/DefaultCaseConverter.java
@@ -1,7 +1,7 @@
 /*
  *   
  *
- * Copyright  1990-2007 Sun Microsystems, Inc. All Rights Reserved.
+ * Copyright  1990-2009 Sun Microsystems, Inc. All Rights Reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER
  * 
  * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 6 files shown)


yxl:
A known fail in new_Long test is caused by this bug and it is fixed :-)

marco-c:
commit 76801f1c669a9a7 was probably done to improve performance. We should profile to see how many times these methods are called, if they're called too much we should implement them in JS as overrides.

java/cldc1.1.1/com/sun/cldc/i18n/uclc/DefaultCaseConverter.java
```diff
@@ -30,12 +30,57 @@
...
+     * @since   1.0
+     */
     public static boolean isUpperCase(char ch) {
-	return DefaultCaseConverter.toUpperCase(ch) == ch;
+        return (ch >= 'A'  && ch <= 'Z')
+            || (ch >= 0xC0 && ch <= 0xD6)
+            || (ch >= 0xD8 && ch <= 0xDE );
     }
 
     public static native char toLowerCase(char ch);
```
marco-c: Could you revert the entire commit? (Removing the two native definitions from the class and removing their implementation from native.js)

marco-c: Forget it, I thought you also reverted the changes to toLowerCase and toUpperCase.

So, discard my last comment! Could you copy the original phoneME class to java/midp/ and move the modified one to java/custom/?

yxl:
I recovered the original java file and make the native implementation of `toLowerCase` and `toUpperCase` as overrides.

PR #828, opened 2015-01-07
Class.forName should reports the correct name of ClassNotFoundException

yxl:
If class `A` imports a missing class `B`, `Class.forName("A")` will report `A` is not found exception other than `B` is not found.

This patch fixes this and let `Class.forName` report correct name of `ClassNotFoundException`.


marco-c:
This probably happens when there's a super class missing

andreasgal:
Nice find. Thanks!

PR #831, opened 2015-01-08
Load JAD and JAR files in parallel; create JAD and JAR files in parallel

marco-c:


main.js
```diff
@@ -145,25 +155,24 @@ if (urlParams.downloadJAD) {
...
+            Promise.all([
+              new Promise(function(resolve, reject) {
+                fs.create("/midlet.jad", new Blob([ data.jadData ]), function() {
+                  resolve();
+                });
+              }),
+              new Promise(function(resolve, reject) {
+                fs.create("/midlet.jar", new Blob([ data.jarData ]), function() {
+                  resolve();
+                });
```
mykmelez: These two anonymous functions that just call *resolve* could simply be references to *resolve* itself.

PR #834, opened 2015-01-09
Fix HttpUrl.java

### LINKED ISSUE #185: Fix TestHttpUrl
There are 5 known failures here, we should fix them.

yxl:
This fixed #185.

The `HttpUrl.java` file we copied from the phoneme_feature reference implementation is from the trunk branch of the sun's phoneme project and it is not fully tested and has bug.

The buggy version we used:
https://java.net/projects/phoneme/sources/svn/content/components/midp/trunk/src/protocol/gcf/reference/classes/com/sun/midp/io/HttpUrl.java?rev=20547

I replaced the `HttpUrl.java`  with the latest released version from the sun's phoneme project of the version mr2-rel-b23 and solved the test failure issue.

The good version I replaced:
https://java.net/projects/phoneme/sources/svn/content/releases/phoneme_feature-mr2-rel-b23/midp/src/protocol/gcf/reference/classes/com/sun/midp/io/HttpUrl.java?rev=20547

### DIFF
--- a/java/midp/com/sun/midp/io/HttpUrl.java
+++ b/java/midp/com/sun/midp/io/HttpUrl.java
@@ -1,7 +1,7 @@
 /*
  *   
  *
- * Copyright  1990-2009 Sun Microsystems, Inc. All Rights Reserved.
+ * Copyright  1990-2007 Sun Microsystems, Inc. All Rights Reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER
  * 
  * This program is free software; you can redistribute it and/or
@@ -209,13 +209,7 @@ public class HttpUrl {
             start = afterScheme;
         } else {
             // this is not relative URL so the path must begin with "/"
-            // can be [IPv6/60].../
-            int posSqBr = url.indexOf(']', startOfAuthority);
-            if (posSqBr > -1) {
-                start = url.indexOf('/', posSqBr);
-            } else {
-                start = url.indexOf('/', startOfAuthority);
-            }
+            start = url.indexOf('/', startOfAuthority);
         }
 
         // do not parse an empty path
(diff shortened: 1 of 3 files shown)


yxl:
The stable release version breaks the test of `TestUrl` and introduces other bugs... Let's keep the trunk version...

mykmelez:
> The HttpUrl.java file we copied from the phoneme_feature reference implementation is from the trunk branch of the sun's phoneme project and it is not fully tested and has bug.

Technically, our version is from this "mr4 b01" release:

http://download.java.net/mobileembedded/phoneme/feature/phoneme_feature-mr4-rel-src-b01-20_apr_2009.zip

Which is the most recent release in this directory of Phoneme Feature downloads:

http://download.java.net/mobileembedded/phoneme/feature/

But the version of the file in that release is identical to the one on trunk, except that the copyright notice at the top has a different date.

@yxl: Can you describe the bug you're seeing? Even if we don't replace the file, we should still address that issue (or at the very least track it).

yxl:
The bug with current implementation is that it doesn't allow the host to start with digits, such as `123.domain`. But from [wikipedia] (http://en.wikipedia.org/wiki/Hostname) and RFC 1123, it should be allowed.

The old version (mr2-rel-b23) implementation doesn't have such issue, however it isn't capable of checking and recognizing ipv6 URLs and couldn't pass tests of [TestUrl.java](https://github.com/andreasgal/j2me.js/blob/08ac88adbdd0bc0499db33a2a2a42b27ae0a077f/tests/com/sun/midp/io/TestUrl.java#L61)

mykmelez:
> The bug with current implementation is that it doesn't allow the host to start with digits, such as 123.domain. But from wikipedia and RFC 1123, it should be allowed.

Ok, I filed #990 on it so we track resolution of the issue.

Issue #758, opened 2014-12-20
AudioPlayer::getMediaTime returns 0 when the audio has ended


marco-c:
Is this the expected behavior?

PR #839, opened 2015-01-10
WIP: use slimerjs 0.10pre to kill and report unresponsive scripts

mykmelez:
This branch takes advantage of the support in slimerjs 0.10pre for trapping and killing unresponsive scripts, which I suspect are the cause of casperjs hangs like the ones on the compiler branch and [this one](https://travis-ci.org/andreasgal/j2me.js/builds/46243537) on a minor change (from #826, although the change has been removed, perhaps in a rebase).

We might want to be more specific about which pre-release build we download. This branch currently downloads the latest, which could break automation non-obviously. But there isn't a link to a specific build on the slimerjs download server (that I can tell), so we'd have to host one ourselves.

This passes tests locally, but *javax.microedition.media.TestAudioPlayer-2* seems to fail on Travis, so this is still a work-in-progress.


marco-c:
Nice, with slimerjs 0.10pre we should also be able to enable socket tests

mykmelez:
@yxl, @marco-c Do either of you know why *TestAudioPlayer-2* is failing on this branch? It looks like the audio file starts and then finishes playing (thus reseting *mediaTime* to `0`) while the thread is still asleep:

> playerUpdate event: started 0
> playerUpdate event: endOfMedia 4735000
> javax.microedition.media.TestAudioPlayer-2: fail [Math.abs(0 - 501) < 100]

marco-c:
Mmmmh, the audio is five seconds long and the amount of time the thread actually slept is 500 milliseconds, so it shouldn't have finished playing! In any case, when it finishes playing, getMediaTime should return the duration of the audio, not 0 (after the fix to #758).

What's the value of mediaTimeBeforeSleep?

marco-c:
Maybe a rounding error? What happens if you do:

    long mediaTimeBeforeSleep = player.getMediaTime();
    Thread.sleep(500);
    long actualTimeSlept = 1000 * (System.currentTimeMillis() - currentTimeBeforeSleep);
    long mediaTime = player.getMediaTime() - mediaTimeBeforeSleep;

marco-c:
I've tested this, the problem is indeed that the media has finished playing, even before the first call to player.getMediaTime(). So mediaTimeBeforeSleep and mediaTime are the same because they're both 4,735 seconds.

marco-c:
This is the output if I remove the *while(player.getMediaTime() <= 0)*:

    playerUpdate event: started 0
    playerUpdate event: endOfMedia 4735000
    mediaTimeBeforeSleep: 0
    javax.microedition.media.TestAudioPlayer-2: fail [Math.abs(4735 - 501) < 100]

So, looks like the audio finishes playing almost instantaneously.

marco-c:
I can't reproduce the failure locally either.

mykmelez:
Hmm, it looks like the test failure is a regression in XULRunner.

Travis installs SlimerJS 0.9.2, which bundles XULRunner 31.0. But this branch installs SlimerJS 0.10pre, which bundles XULRunner 33.1.1 (but can be configured to use an older version of XULRunner). When I configure it to use the same version of XULRunner as SlimerJS 0.9.2, then Travis doesn't experience the test failure (on a test build in my fork). (Still unsure why we can't reproduce locally.)

So commit f67919b switches to the "lightweight" edition of SlimerJS (which comes without XULRunner, so you have to provide it yourself) and XULRunner 31.0. We should be able to use this combination of SlimerJS and XULRunner, as SlimerJS advertises itself as compatible with Firefox 17 - 33.1.

marco-c:
> Still unsure why we can't reproduce locally.

Indeed! Let's hope the regression gets fixed in newer releases!

Could you test the detection of unresponsive scripts actually works on Travis (using a fake PR)?

mykmelez:
> Could you test the detection of unresponsive scripts actually works on Travis (using a fake PR)?

Yes, although I don't need a fake PR for it, as I've enabled Travis for commits on my fork. Commit https://github.com/mykmelez/j2me.js/commit/e17472a2c239cb3571bffa0211feb833a3ed3412 adds an infinite loop to the unresponsive-script branch, and https://travis-ci.org/mykmelez/j2me.js/builds/46744642 demonstrates the behavior of the test harness on Travis when it encounters that loop.

mykmelez:
One more thing: according to the [SlimerJS FAQ](http://slimerjs.org/faq.html), the [directory from which we're downloading the 0.10pre build](http://download.slimerjs.org/nightlies/latest-slimerjs-master/) should get updated every day with a new nightly build (even though it hasn't been updated since December 17), which could cause regressions that would be difficult to spot.

So I copied the current build to a new [directory on ftp.mozilla.org](https://ftp.mozilla.org/pub/mozilla.org/labs/j2me.js/) and updated Travis to use that build in 2ac0aaa and 8e84a43.

cc: @laurentj so he's aware that we're using this nightly build.
cc: @EricRahm so he's aware of the (platform?) regression.

PR #841, opened 2015-01-11
Support updates

marco-c:


midp/midp.js
```diff
@@ -227,7 +227,36 @@ Native.create("com/sun/midp/security/Permissions.loadGroupPermissions.(Ljava/lan
...
+            dialog.classList.add('visible');
+            document.body.appendChild(dialog);
+
+            performDownload(request, dialog, function(data) {
+              dialog.parentElement.removeChild(dialog);
+
+              Promise.all([
+                new Promise(function(resolve, reject) {
+                  fs.remove("/midlet.jad", function() {
+                    console.log(data.jadData);
```
mykmelez: Nit: remove this logging statement.

tests/Manifest1.jad
```diff
@@ -3,6 +3,6 @@ MIDlet-Jar-Size: 2436
 MIDlet-Jar-URL: tests.jar
 MIDlet-Name: AMIDlet
 MIDlet-Vendor: MIDlet Suite Vendor
-MIDlet-Version: 1.0.0
+MIDlet-Version: 1
```
mykmelez: I think MIDlet-Version needs to be in the form *N.N.N* (or at least *N.N*).

marco-c: I've had to change this because there's a bug somewhere in our code that replaces "." characters with "/" characters in the args parameter.

marco-c: Well, it was probably done on purpose to make RunTests work with class names both with points and with slashes.

mykmelez: That would be https://github.com/andreasgal/j2me.js/blob/f8e9a09acea4617bd22315eb3240cbb046828e43/libs/urlparams.js#L43.

PR #843, opened 2015-01-12
Allow importing FS from a URL

marco-c:
This PR makes it possible to import FS in the simulator or on the phone (where the file input doesn't allow to select any file)


marco-c:
I'm seeing a "QuotaExceededError" in the simulator, despite the addition of the "storage" permission, that should allow unlimited storage.

mykmelez:
> I'm seeing a "QuotaExceededError" in the simulator, despite the addition of the "storage" permission, that should allow unlimited storage.

Hmm, perhaps @jryans or @ochameau knows if this is an issue in the Simulator.

jryans:
> >I'm seeing a "QuotaExceededError" in the simulator, despite the addition of the "storage" permission, that should allow unlimited storage.
> 
> Hmm, perhaps @jryans or @ochameau knows if this is an issue in the Simulator.

I wasn't aware of that as known issue, perhaps Alex will know more.

I tried to test your branch locally, but I wasn't sure how to get to these file inputs. Do I need to edit the `launch_path`?

mykmelez:
> I tried to test your branch locally, but I wasn't sure how to get to these file inputs. Do I need to edit the launch_path?

Yes, remove the `autosize=1` parameter from *launch_path*. Then the canvas display element (which renders the midlet) won't take up the entire simulator screen, and you'll see a developer panel below it (you might have dismiss the virtual keyboard first):

screen shot 2015-01-12 at 15 36 59

Click and drag in the developer panel to scroll it until the file inputs come into view:

screen shot 2015-01-12 at 15 38 06

mykmelez:
When I try to import a file by URL on Simulator 2.2, I get "Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://localhost/~myk/fs.json. This can be fixed by moving the resource to the same domain or enabling CORS."

main.html
```diff
@@ -95,7 +95,9 @@
       <button id="clearstorage">Clear storage</button>
       <button id="exportstorage">Export storage</button>
       <div>Import storage:</div>
-      <input type="file" id="importstorage">
+      <input type="file" id="importstoragefile">
+      <input type="text" id="importstorageurl">
```
mykmelez: Nit: the field could use a placeholder attribute to suggest that it takes a URL.

marco-c:
> When I try to import a file by URL on Simulator 2.2, I get "Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://localhost/~myk/fs.json. This can be fixed by moving the resource to the same domain or enabling CORS."

This issue could be fixed by using systemXHR (by implementing another pipe), but probably we don't really need it (since we're always going to load a URL to a file in the app package or from a server we control).

mykmelez:
> This issue could be fixed by using systemXHR (by implementing another pipe), but probably we don't really need it (since we're always going to load a URL to a file in the app package or from a server we control).

Ah, I see. Ok, then I added a database dump to the package (in the root directory) and imported it by entering a relative "URL" that was just the filename: `fs.json`. Whereupon the file was imported successfully without hitting *QuoteExceededError*. The file is about 14MB. @marco-c, is that not large enough to trigger the problem?

(I also ran into the problem that focusing the URL field would show the virtual keyboard and then immediately hide it. I couldn't get the keyboard to remain visible until I clicked outside the URL field and then inside it again.)

marco-c:
> The file is about 14MB. @marco-c, is that not large enough to trigger the problem?

Mine is 22,5 MB, so maybe the limit is higher

yxl:
I guess the reason is that the `storage` permission doesn't take effect in main.html, which is not privileged. You have to move the code from main.html to index.html get the right permission for unlimited indexedDB size.

marco-c:
If this is the case, the problem isn't restricted to this PR but to the entire app. We should figure this out

mykmelez:
> If this is the case, the problem isn't restricted to this PR but to the entire app. We should figure this out

The [MDN IndexedDB page](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) says:

> Firefox has no limit on the IndexedDB database's size. The user interface will just ask permission for storing blobs bigger than 50 MB.

That limit is set by [modules/libpref/init/all.js](http://mxr.mozilla.org/mozilla-central/source/modules/libpref/init/all.js?rev=83c4b180a815#118):

> // Space to allow indexedDB databases before prompting (in MB).
> pref("dom.indexedDB.warningQuota", 50);

But [b2g/app/b2g.js](http://mxr.mozilla.org/mozilla-central/source/b2g/app/b2g.js?rev=59452425e446#304) reduces it to 5MB:

> pref("dom.indexedDB.warningQuota", 5);

Nevertheless, the simulator doesn't hit the error when it creates files larger than 5MB. I hacked together this quota stress test and ran it in the 2.2 simulator:

https://github.com/mykmelez/j2me.js/compare/quota-exceeded-error

I do hit QuotaExceededError, but not until the blob size grows to about 25MB and my database grows to about 900MB.

The behavior was similar on my Flame device with FxOS 2.1, except that FxOS killed the app (presumably for using too much memory as it tried to create the humongous files it writes to the database) before it hit an error. (Nevertheless, the last file it successfully wrote was 13MB, which is much higher than 5MB; and it should have written at least 91MB to the database.)

Reading through some platform code suggests there are other reasons why the error might appear, though. For example, [test_lowDiskSpace.js](http://mxr.mozilla.org/mozilla-central/source/dom/indexedDB/test/unit/test_lowDiskSpace.js) expects it when merely creating a database or an object store in "low disk mode." So @marco-c's error might be unrelated to the size of the import file. (In my case, my laptop is pretty low on disk space—I have ~10GB left of a 768GB disk—so that might explain the behavior I'm seeing.)

We should confirm with platform folks like @fabricedesre or @erahm, but it seems like the error is unrelated to @marco-c's changes and isn't a major issue (although we should figure out how to make the virtual fs fail gracefully when it hits the error).

mykmelez:
@marco-c It seems like we can merge this, despite hitting the QuotaExceededError in some cases, as that error isn't actually caused by your changes but rather exposed by them. (We should still figure out exactly why that error is occurring and make sure we aren't going to run into it on user devices; but we can do that separately.) What do you think?

marco-c:
Yes, I agree we can merge this. I've also tested the patch on a Flame, and it works correctly (so it's probably a simulator-only issue).

PR #851, opened 2015-01-13
move configuration from launch path params to midlet.js

mykmelez:
Unfortunately, the Marketplace validator rejects query parameters in launch paths because of [bug 943710](https://bugzilla.mozilla.org/show_bug.cgi?id=943710), so this makes them configurable via a *config* Object defined in midlet.js.

It isn't entirely functionally equivalent, since the query parameters allow us to specify parameters that only apply when the app is installed as an app (as opposed to being loaded in the browser). But it's close enough.

I've tested and confirmed that I can configure the project for a midlet, package it, run the midlet on a device as before, and pass validation on Marketplace.

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -1,7 +1,7 @@
 {
   "name": "j2me.js",
   "description": "j2me interpreter for firefox os",
-  "launch_path": "/index.html?autosize=1&logConsole=web",
+  "launch_path": "/index.html",
   "origin": "app://j2mejs.mozilla.org",
   "icons": {
     "128": "/img/icon-128.png"
--- a/classinfo.js
+++ b/classinfo.js
@@ -95,7 +95,7 @@ function MethodInfo(opts) {
       this.consumes++;
     }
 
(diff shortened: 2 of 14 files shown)


manifest.webapp
```diff
@@ -1,7 +1,7 @@
 {
   "name": "j2me.js",
   "description": "j2me interpreter for firefox os",
-  "launch_path": "/index.html?autosize=1&logConsole=web",
+  "launch_path": "/index.html",
```
TimAbraldes: It looks like we don't default to autosize=1 anymore; does this cause apps to appear incorrectly on device and/or in emulators?

mykmelez: > It looks like we don't default to autosize=1 anymore; does this cause apps to appear incorrectly on device and/or in emulators?

Yes, that's the functionality regression to which I was referring when I noted (obliquely), "It isn't entirely functionally equivalent, since the query parameters allow us to specify parameters that only apply when the app is installed as an app (as opposed to being loaded in the browser)."

PR #871, opened 2015-01-15
Enable aot target by default and add optimized compiler tests.

brendandahl:
To load compiled js files I've added the jsFiles option to config/urlparam. I'm open to suggestions on other ways to do this, as it doesn't seem all that great.

BTW: We're still having a problems a midlet with compiled code enabled, but all the test are passing.


marco-c:
The first part of this PR was done in #916, are you still planning to do the second part?

brendandahl:
> The first part of this PR was done in #916, are you still planning to do the second part?

Yes, we really need to do multiple stages of tests:

1) no jit
2) jit only
3) jit and aot
4) non-release(debug) build with jit and aot

I probably won't get around this until next week.

mykmelez:

PR #872, opened 2015-01-16
Revoke video object URLs.

### LINKED ISSUE #736: Camera DumbPipe in index.js leaks object URL

KevinGrandon:
I haven't tested this at all yet. I'm going to figure out how to do that next.

This is probably meant to fix #736.

### DIFF
--- a/index.js
+++ b/index.js
@@ -334,6 +334,7 @@ DumbPipe.registerOpener("audiorecorder", function(message, sender) {
 
 DumbPipe.registerOpener("camera", function(message, sender) {
   var mediaStream = null;
+  var url = null;
 
   var video = document.createElement("video");
   document.body.appendChild(video);
@@ -359,8 +360,13 @@ DumbPipe.registerOpener("camera", function(message, sender) {
     audio: false,
   }, function(localMediaStream) {
     mediaStream = localMediaStream;
+    url = URL.createObjectURL(localMediaStream);
 
-    video.src = URL.createObjectURL(localMediaStream);
+    video.onerror = video.onloadeddata = function() {
+      URL.revokeObjectURL(url);
+    };
+
+    video.src = url;
     video.play();
   }, function(err) {
     console.log("Error: " + err);


index.js
```diff
@@ -359,8 +360,14 @@ DumbPipe.registerOpener("camera", function(message, sender) {
     audio: false,
   }, function(localMediaStream) {
     mediaStream = localMediaStream;
+    url = URL.createObjectURL(localMediaStream);
 
-    video.src = URL.createObjectURL(localMediaStream);
+    video.onerror = video.onloadeddata = () => {
+      URL.revokeObjectURL(url);
+      video.removeAttribute('src');
```
yxl: It is not required to remove the `src` attribute. The object url is a plain string. Holding it won't increase the references to the target object.

index.js
```diff
@@ -359,8 +360,14 @@ DumbPipe.registerOpener("camera", function(message, sender) {
     audio: false,
   }, function(localMediaStream) {
     mediaStream = localMediaStream;
+    url = URL.createObjectURL(localMediaStream);
 
-    video.src = URL.createObjectURL(localMediaStream);
+    video.onerror = video.onloadeddata = () => {
```
mykmelez: Nit: we're trying to maintain as much compatibility with Chrome as possible, since their developer tools have some features that are missing in Firefox (like the ability to break on an unresponsive script), so we've been avoiding Firefox-specific JS features like fat-arrow functions.

mykmelez:
> I haven't tested this at all yet. I'm going to figure out how to do that next.

Testing this is tricky! Since the code that calls this code runs in the child frame (main.js), while this code runs in the parent frame. I'm not sure if there's even an existing test in the codebase (but @yxl or @marco-c should know). The AudioRecorder.java midlet (manual test) might be the closest thing to it.

In theory, you could write a Testlet (Java test class that gets run by the RunTests test runner) that accesses the relevant Java API, then calls a native defined in tests/native.js (which runs in the child frame) that uses a DumbPipe to call a method in tests/index.js (which runs in the parent frame) that locates the &lt;video> element, determines the status of its URL, and passes that back to the native, which passes it back to the test.

Alternately, you could bypass the Java code and write a JavaScript test that accesses the JavaScript API in media.js directly and then does the same DumbPipe dance to confirm that the object URL has been revoked. Or even one that calls the DumbPipe opener in the parent frame directly. You might even be able to write that as a test SlimerJS test function (in tests/automation.js), although most of our JavaScript tests are in separate HTML/JS files that automation.js loads.

But in this case the fix is straightforward enough that I don't think we need an automated test for it, and I'm comfortable landing it without one (provided existing tests continue to pass and our favorite midlets continue to work).

index.js
KevinGrandon: That sounds more like a problem with the Firefox devtools, but I'll go ahead and remove this for now.

KevinGrandon:
@mykmelez thanks for the detailed information. I hate not writing tests, but I agree this can probably land now if you're happy with it. I've gone ahead and made the requested changes, feel free to land this if it looks good.

Thanks for the help, I just wanted to play with j2me a bit :)

mykmelez:
> That sounds more like a problem with the Firefox devtools, but I'll go ahead and remove this for now.

Indeed! And unfortunately not the only one.

> @mykmelez thanks for the detailed information. I hate not writing tests, but I agree this can probably land now if you're happy with it. I've gone ahead and made the requested changes, feel free to land this if it looks good.

It looks great. Merged!

> Thanks for the help, I just wanted to play with j2me a bit :)

You bet, thanks for the fix!

marco-c:
Do we need to revoke the object URL also when we're done playing the media?

PR #873, opened 2015-01-16
Baseline JIT and Lazy Linking

mbebenita:


brendandahl:
Tests don't currently pass with compiled code enabled. However, the midlet is working. If you want try out a midlet you should:

1) make aot
2) uncomment the program.jar.js include in main.html
3) comment out the tests.jar.js include in main.html
4) ignore the startup time and enjoy the less clunky feeling interface

We should also merge in #871 to this branch, but it's getting late and I'll try that tomorrow.

Issue #874, opened 2015-01-16
Override the Graphics class


marco-c:
This is used a lot, we could override its methods instead of implementing its natives

marco-c:
Done in #882.

Issue #853, opened 2015-01-14
Move natives and overrides to new bindings


brendandahl:
Move natives and overrides to new bindings

This can probably split into multiple parts:
1) Move code out of bindings.ts and create a new bindings folder that mimics the java folder structure.
2) Move native functions over to the new binding files. (see below)
3) Fix natives to use the new field bindings. i.e. Remove all uses of getField in natives and directly use `javaScriptFieldName` on the object or static object.

Example
```js
"java/lang/Object": {
  native: {
    "hashCode.()I": function (): number {
      var self: J2ME.java.lang.Object = this;
      if (self._hashCode) {
        return self._hashCode;
      }
      return self._hashCode = $.nextHashCode();
    }
  },
  fields: {
    staticSymbols: {
      "javaFieldName": "javaScriptFieldName" 
    },
    instanceSymbols: {
      "javaFieldName": "javaScriptFieldName"
    }
  }
}
```

mbebenita:
When porting natives over we should also us the appropriate type annotations. This will only be possible if the natives themselves are ported to TS. For instance, 

```javascript
Native["java/lang/System.getProperty0.(Ljava/lang/String;)Ljava/lang/String;"] = function (key) { ...
``` 

Should be written as:

```javascript
"java/lang/System.getProperty0.(Ljava/lang/String;)Ljava/lang/String;": function (key: java.lang.String): java.lang.String { ...
```

This means we'll need to also define the String class, which we can do by defining an interface:

```javascript
export module java.lang {
    export interface String extends java.lang.Object {
      str: string;
    }
}
```

Java fields are mangled and the implementer of a native should not be concerned with the internal name, however if a symbol is created then that name needs to appear in this interface.

It is now possible to call Java methods in natives. Methods that may unwind can cause problems and we need to emit appropriate unwinding code. (Resuming however is not possible, if we really need to support this then we'll have to write natives in a way that they can resume.)

mbebenita:
This issue should have a high priority. Calls to `getField(...)` are very slow, and hot methods should never ever use it. For instance, this hot method is called ~81K during some midlet startups. 

```
Override["com/sun/cldc/io/ResourceInputStream.available.()I"] = function() {
    var handle = this.klass.classInfo.getField("I.fileDecoder.Ljava/lang/Object;").get(this);
    if (!handle) {
        throw $.newIOException();
    }
    return handle.data.length - handle.pos;
};
```

[...]

PR #882, opened 2015-01-17
Implement `Graphics` methods natively

TimAbraldes:
#874


mykmelez:
Actually appearing to draw is good! :wink: How's perf?

TimAbraldes:
> Actually appearing to draw is good! :wink: How's perf?

There's not enough drawing to really compare :)

marco-c:
Instead of copying the Graphics class in custom, couldn't we override its methods with JS?

TimAbraldes:
> Instead of copying the Graphics class in custom, couldn't we override its methods with JS?

We could; it would mean doing a lot of `this.klass.classInfo.getField("I.someField.Ljava/lang/someType;").get(this)` in the native methods. I don't know if that would affect performance at all but it's definitely a little harder to read. Is there a reason to avoid copying the Graphics class into custom?

marco-c:
> We could; it would mean doing a lot of this.klass.classInfo.getField("I.someField.Ljava/lang/someType;").get(this) in the native methods. I don't know if that would affect performance at all but it's definitely a little harder to read. Is there a reason to avoid copying the Graphics class into custom?

We tried to move as less classes as possible in custom/, this is why we introduced overrides.

You could avoid the need to use getField in two ways:
- Don't use getField(...).set and getField(...).get since you're overriding the entire class (for example, see ByteArrayInputStream in override.js)
- Use the new bindings (but it would require fixing #853 first)

TimAbraldes:
If we override the methods of the `Graphics` class but don't modify Graphics.java, we'll have to trade `synchronized` methods for non-synchronized ones. This updated patch calls natively-implemented methods from `synchronized` ones where appropriate, and re-implements methods natively when possible.

marco-c:
Have you tried making the methods non synchronized? It would probably be better performance-wise and, given the "SYNC NOTE" in the class, probably safe.

TimAbraldes:
> Have you tried making the methods non synchronized? It would probably be better performance-wise and, given the "SYNC NOTE" in the class, probably safe.

I'll try that locally and see if any issues crop up!

java/custom/javax/microedition/lcdui/Graphics.java
```diff
@@ -0,0 +1,1639 @@
...
+     * @param anchor the anchor point for positioning the region within
+     *        the destination image
+     *
+     * @throws IllegalStateException if the destination of this
+     * <code>Graphics</code> object is the display device
+     * @throws IllegalArgumentException if the region to be copied exceeds
+     * the bounds of the source image
+     *
+     */
+     private native void copyArea0(int x_src, int y_src, int width, int height, int x_dest, int y_dest, int anchor);
```
mykmelez: Nit: looks like there's an extra space of indentation here.

TimAbraldes:
>> Have you tried making the methods non synchronized? It would probably be better performance-wise and, given the "SYNC NOTE" in the class, probably safe.
> I'll try that locally and see if any issues crop up!

Removing all synchronization didn't seem to have any ill effect on WA. The Travis failure looks unrelated; can someone confirm?

mykmelez:
> Removing all synchronization didn't seem to have any ill effect on any apps I've tested. The Travis failure looks unrelated; can someone confirm?

Yes, unrelated. That's an intermittent failure I've started to see since the new VM landed.

java/custom/javax/microedition/lcdui/Graphics.java
```diff
@@ -0,0 +1,1559 @@
...
+    public static final int DOTTED = 1;
+
+    private native void init0(Font font);
+    private native void initScreen0(int displayId, int width, int height);
+    private native void initImage0(Image img, int width, int height);
+
+    /**
+     * Create a Graphics object
+     */
+    Graphics() {
```
marco-c: I think the constructor could be "nativized" as well (see for example, string.js)

TimAbraldes: > I think the constructor could be "nativized" as well (see for example, string.js)

We'd have to get the default font from somewhere. It's kind of frustrating that `getDefaultFont()` is `synchronized (Display.LCDUILock)` - it means that it's potentially much slower than it needs to be (due to locking) and also that we can't call it from JS (because it yields).

mykmelez: > We'd have to get the default font from somewhere. It's kind of frustrating that getDefaultFont() is synchronized (Display.LCDUILock) - it means that it's potentially much slower than it needs to be (due to locking) and also that we can't call it from JS (because it yields).

I don't see why it would have to be. *Font.getDefaultFont* constructs a Font object, and the Font constructor calls *Font.init*, but there's nothing in *Font.init* that looks like it needs synchronization, so I'd override *Font.getDefaultFont* with a non-synchronized implementation if synchronization is expensive here.

mykmelez: For that matter, there's a ton of what looks like unnecessary synchronization all over the Java codebase. I filed #920 on removing it.

PR #886, opened 2015-01-19
baseline compiler and lazy linking w/fixes

mykmelez:
This is @mbebenita's baseline compiler from #873 with:

1. a merge from master;
2. @brendandahl's test fixes from https://github.com/mbebenita/j2me.js/pull/15;
3. my modification of one of those test fixes in da07644 and 89de017.


runtime.ts
```diff
@@ -1149,7 +1206,18 @@ module J2ME {
   }
 
   function findCompiledMethod(klass: Klass, methodInfo: MethodInfo): Function {
-    return jsGlobal[methodInfo.mangledClassAndMethodName];
+    var name = methodInfo.mangledClassAndMethodName;
+    var method = jsGlobal[name];
+    return method;
+    if (!method) {
+      var compiledMethod = baselineCompileMethod(methodInfo, CompilationTarget.Runtime);
```
mykmelez: This looks like a JIT compilation step that has been disabled. If I enable it, tests hang. Is it a dead end? If so, we should remove it. Otherwise, we should make it work.

mbebenita: > This looks like a JIT compilation step that has been disabled. If I enable it, tests hang. Is it a dead end? If so, we should remove it. Otherwise, we should make it work.

Okay, I'll remove it form my branch.

main.js
```diff
@@ -52,12 +45,24 @@ var getMobileInfo = new Promise(function(resolve, reject) {
...
+
 jars.forEach(function(jar) {
   loadingPromises.push(load(jar, "arraybuffer").then(function(data) {
     CLASSES.addPath(jar, data);
   }));
 });
 
+var packs = [
+  // "build/classes.jar.js.pack",
+  // "build/tests.jar.js.pack"
```
mykmelez: How much does packing help? Since it isn't being used yet, and since packing is an optional extra step, let's move its code to a separate pull request to make them easier to evaluate.

actors.ts
```diff
@@ -114,6 +130,9 @@ module J2ME {
       this.name = opts.name;
       this.signature = opts.signature;
       this.classInfo = opts.classInfo;
+      if (!(this.classInfo instanceof ClassInfo)) {
+        debugger;
```
mykmelez: These and the other instances of the *debugger* statement should be assertions, *console.error* calls, or the like.

jit/baseline.ts
```diff
@@ -0,0 +1,1200 @@
...
+        //  // Check to see if we can get away without going through the switch again.
+        //  if (i < blocks.length - 1 &&
+        //    block.successors.length === 1 &&
+        //    blocks[i + 1].startBci === stream.currentBCI) {
+        //    needsFallthroughPC = false;
+        //  }
+        //  if (needsFallthroughPC) {
+        //    this.emitter.writeLn("bi = " + this.getBlockIndex(stream.currentBCI) + "; break;");
+        //  }
+        //}
```
mykmelez: This could use a comment explaining why it's commented out.

jit/baseline.ts
```diff
@@ -0,0 +1,1200 @@
...
+        case Bytecodes.IFNONNULL      : this.emitIfNull(block, stream, Condition.NE); break;
+        ///*
+        //case Bytecodes.GOTO_W         : genGoto(stream.readFarBranchDest()); break;
+        //case Bytecodes.JSR_W          : genJsr(stream.readFarBranchDest()); break;
+        //case Bytecodes.BREAKPOINT:
+        //throw new CiBailout("concurrent setting of breakpoint");
+        //default:
+        //throw new CiBailout("Unsupported opcode " + opcode + " (" + nameOf(opcode) + ") [bci=" + bci + "]");
+        //}
+        //*/
```
mykmelez: Here and above, some comments explaining why these cases are commented out would be useful.

jit/blockMap.ts
```diff
@@ -150,33 +155,34 @@ module J2ME.Bytecode {
...
+      //  case Bytecodes.DALOAD:
+      //  case Bytecodes.AALOAD:
+      //  case Bytecodes.BALOAD:
+      //  case Bytecodes.CALOAD:
+      //  case Bytecodes.SALOAD:
+      //  case Bytecodes.PUTFIELD:
+      //  case Bytecodes.GETFIELD:
+      //    return false; // ???
+      //}
+      //return false;
```
mykmelez: Nit: remove this dead code.

jit/blockMap.ts
```diff
@@ -302,7 +297,7 @@ module J2ME.Bytecode {
         if (block.isExceptionEntry) {
           // Loops that are implicitly formed by an exception handler lead to all sorts of corner cases.
           // Don't compile such methods for now, until we see a concrete case that allows checking for correctness.
-          throw new CompilerBailout("Loop formed by an exception handler");
+          // throw new CompilerBailout("Loop formed by an exception handler");
```
mykmelez: No longer needed? Then let's update the comment.

libs/console.js
```diff
@@ -20,12 +20,11 @@
...
    * of one or more of these consoles:
    *    web: the browser's Web Console (default)
-   *    page: the in-page console (an HTML element with ID "console")
    *    native: the native console (via the *dump* function)
    *    terminal: a faster canvas based console if Shumway.js is included.
    */
-  var ENABLED_CONSOLE_TYPES = (config.logConsole || "page").split(",");
-  var minLogLevel = LOG_LEVELS[config.logLevel || "log"];
+  var ENABLED_CONSOLE_TYPES = (urlParams.logConsole || "page").split(",");
+  var minLogLevel = LOG_LEVELS[urlParams.logLevel || "log"];
```
mykmelez: Hmm, this looks like a reversion of a previous change. Why is it needed?

main.html
```diff
@@ -10,9 +10,10 @@
   <link rel="stylesheet" type="text/css" href="style/drawer.css">
 
   <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1">
+  <script type="text/javascript" src="midlet.js" defer></script>
   <!-- Uncomment the line below to enable the profiler. -->
   <!--<script type="text/javascript" src="build/shumway.js" defer></script>-->
-  <script type="text/javascript" src="midlet.js" defer></script>
+  <!--<script type="text/javascript" src="libs/terminal.js" defer></script>-->
```
mykmelez: This new terminal looks handy, but since it's disabled, and since it's orthogonal to the purpose of this branch, can it be split into its own pull request?

midlet.js
```diff
@@ -13,3 +13,9 @@ var MIDlet = {
 
 var config = {
 };
+
+// The base directory of the app, relative to the current page.  Normally this
+// is the directory from which the page was loaded, but some test pages load
+// from a subdirectory, like tests/fs/, and they set this accordingly such that
+// code loads files, like libs/fs-init.js, can load them from the right place.
+var APP_BASE_DIR = "./";
```
mykmelez: This isn't midlet-specific, so I'd rather put it elsewhere. But I don't yet have a good place for it to go, so this is alright for now.

override.js
```diff
@@ -1,7 +1,7 @@
 /* -*- Mode: Java; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*- */
 /* vim: set shiftwidth=2 tabstop=2 autoindent cindent expandtab: */
 
-'use strict';
+// 'use strict';
```
mykmelez: Why is this commented out?

main.js
mbebenita: Packing was an experiment to see how much faster we could load pre-jitted code. I'll remove it for now, we can bring it back if we need to.

actors.ts
mbebenita: Will remove.

jit/baseline.ts
mbebenita: This has already been removed in my latest branch.

libs/console.js
mykmelez: > Hmm, this looks like a reversion of a previous change. Why is it needed?

Hmm, looks like it isn't, according to my tests. Fixed in 8841f32.

jit/baseline.ts
mbebenita: Fixed in latest branch.

override.js
mykmelez: > Why is this commented out?

Seems unnecessary, in my testing; commented back in, in https://github.com/mykmelez/j2me.js/commit/f367a71386b5b4fc31f7142a8a265c8be2ba53dc.

libs/console.js
mbebenita: This is likely a merge error, will fix on the latest branch.

override.js
mbebenita: Leftover type-o from trying to write a polyfill for Chrome.

PR #887, opened 2015-01-20
Header bar improvement

yxl:
One of my midlet uses the menu commands of `OK` and `BACK` heavily to navigate among different screens. The user needs two clicks to reach these commands, since they are folded in a hidden drawer. To make the midlet easy to use, I move these commands out of the drawser as the Nokia Asha 3110 does.

The `OK` command is placed on right of the header bar and the `BACK` command is placed at the bottom right corner of the midlet.

The drawer will be hidden if it is empty.

I abandon the url parameter of `nativeMenu`, which forces the header bar to show. The header bar is hidden if the midlet runs in full screen mode. Otherwise the header bar shows.

It also fixes two small UI bugs.

It may be easier to review by each commit and run `/index.html?jars=tests/tests.jar&midletClassName=gfx.CanvasWithHeaderTest` to test the header bar.


tests/automation.js
```diff
@@ -21,6 +21,7 @@ casper.options.onWaitTimeout = function() {
 var gfxTests = [
   { name: "gfx/AlertTest", maxDifferent: 1942 },
   { name: "gfx/CanvasTest", maxDifferent: 0 },
+  { name: "gfx/CanvasWithHeaderTest", maxDifferent: 2000 },
```
mykmelez: Does this really need to be as high as 2000? I see 1368 different pixels in my runs on Mac, which typically has the highest number of different pixels.

mykmelez: And the number was 969 on the Travis run.

mykmelez:
Other than that nit, this looks fine, but I'd like @marco-c to look at it too.

tests/automation.js
yxl: I reduced maxDifferent to 1400 to be compatible with that on Mac. It's value varies a lot with different platform. On my ubuntu desktop the value is 0.

marco-c:
I'll merge this once conflicts are resolved!

PR #888, opened 2015-01-20
Make drawString synchronous

marco-c:
This PR makes drawString synchronous, adds some new emoji that were needed by a MIDlet I was testing, stores the emoji in a few sprites instead of hundreds of different images.

I've used a black rect when an emoji was missing in the Twitter set.


midp/gfx.js
```diff
@@ -423,19 +426,24 @@ var currentlyFocusedTextEditor;
...
+
+    function withFont(font, c) {
         if (c.font != font.css) {
           c.font = font.css;
         }
-        return c.measureText(str).width | 0;
     }
 
     function withTextAnchor(g, c, anchor, x, y, str, cb) {
+        withFont(g.klass.classInfo.getField("I.currentFont.Ljavax/microedition/lcdui/Font;").get(g), c);
```
yxl: Move `withFont` into the block where `measureWith`  is called to reduce the times of calling `withFont`.

yxl:
It reduces the time of drawing string significantly  for one of my midlet :)

midp/gfx.js
marco-c: withFont is needed to change the font used before drawing the string

PR #890, opened 2015-01-20
warn once for noisy unimplemented natives

mykmelez:
Since dechameleonization, FormLFImpl and ItemLFImpl natives output a gazillion "not implemented" warnings, which pegs the log limit on Travis. So this branch adds an *UnimplementedNative* constructor whose instances output only a single warning, the first time they're called.

### DIFF
--- a/native.js
+++ b/native.js
@@ -1034,3 +1034,22 @@ Native["com/nokia/mid/impl/jms/core/Launcher.handleContent.(Ljava/lang/String;)V
     }));
 };
 
+function UnimplementedNative(signature, returnValue) {
+    var doNotWarn;
+
+    if (typeof returnValue === "function") {
+      doNotWarn = function() { return returnValue() };
+    } else if (typeof returnValue !== "undefined") {
+      doNotWarn = function() { return returnValue };
+    } else {
+      doNotWarn = function() { };
+    }
+
+    var warnOnce = function() {
+        console.warn(signature + " not implemented");
+        warnOnce = doNotWarn;
+        return doNotWarn();
+    };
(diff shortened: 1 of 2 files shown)


native.js
```diff
@@ -1034,3 +1034,22 @@ Native["com/nokia/mid/impl/jms/core/Launcher.handleContent.(Ljava/lang/String;)V
     }));
 };
 
+function UnimplementedNative(signature, returnValue) {
+    var doNotWarn;
+
+    if (typeof returnValue === "function") {
+      doNotWarn = function() { return returnValue() };
```
TimAbraldes: nit: We could probably make this `doNotWarn = returnValue;`

native.js
```diff
@@ -1034,3 +1034,22 @@ Native["com/nokia/mid/impl/jms/core/Launcher.handleContent.(Ljava/lang/String;)V
...
+    } else if (typeof returnValue !== "undefined") {
+      doNotWarn = function() { return returnValue };
+    } else {
+      doNotWarn = function() { };
+    }
```
TimAbraldes: nit: Isn't this anonymous function going to return `undefined` anyway? Maybe we could collapse this into the previous `else if` clause as just an `else`

mykmelez:
> nit: We could probably make this doNotWarn = returnValue;

> nit: Isn't this anonymous function going to return undefined anyway? Maybe we could collapse this into the previous else if clause as just an else

Good points, changed in https://github.com/andreasgal/j2me.js/commit/0b875c364ac20bebea60b030c377433b8d79b507.

PR #891, opened 2015-01-20
bump max different pixels for emoji strings on Mac

mykmelez:
On Mac, these tests produce larger numbers of differing pixels because of a different font weight. For example, here's DrawStringWithEmojiTest (reference on left, my result on right):

drawstringwithemojitest drawstringwithemojitest-result


marco-c:
Probably it's because of the quality of the emoji that has changed with #888.

PR #892, opened 2015-01-20
More fixes for chrome.

brendandahl:
-allows running pim test in chrome
-fixes infinite loop in onerror

### DIFF
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -188,12 +188,10 @@ var currentlyFocusedTextEditor;
                 setImageData(imageData, img.naturalWidth, img.naturalHeight, context);
 
                 URL.revokeObjectURL(img.src);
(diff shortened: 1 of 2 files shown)


midp/gfx.js
```diff
@@ -188,12 +188,10 @@ var currentlyFocusedTextEditor;
                 setImageData(imageData, img.naturalWidth, img.naturalHeight, context);
 
                 URL.revokeObjectURL(img.src);
-                img.src = '';
                 resolve();
             }
             img.onerror = function(e) {
                URL.revokeObjectURL(img.src);
-               img.src = '';
```
mykmelez: It's weird that Chrome would choke on these, but they shouldn't be necessary, as @yxl noted in another PR recently, so it's fine to remove them.

PR #893, opened 2015-01-20
baseline compiler fixes and optimizations

mykmelez:
This is @mbebenita's branch with a bunch of fixes and optimizations for the baseline compiler. It still doesn't do JIT compilation; he's working on that next.


context.ts
```diff
@@ -27,7 +27,8 @@ module J2ME {
   /**
    * Toggle VM tracing here.
    */
-  export var writers = WriterFlags.None; // WriterFlags.Link | WriterFlags.Load;
+  // export var writers = WriterFlags.None; // WriterFlags.Link; // WriterFlags.Link | WriterFlags.Load;
+  export var writers = WriterFlags.None; // WriterFlags.Link; // WriterFlags.Link | WriterFlags.Load;
```
mykmelez: Nit: the commented-out line just duplicates the next one.

context.ts
```diff
@@ -93,6 +133,11 @@ module J2ME {
       this.local[this.localBase + i] = value;
     }
 
+    incLocal(i: number, value: any) {
+      var j = this.localBase + i;
+      this.local[j] = this.local[j] + value | 0
```
mykmelez: Nit: missing semicolon.

jit/analyze.ts
```diff
@@ -76,14 +79,44 @@ module J2ME {
...
+
+
   };
 
-  export function isFinal(classInfo: ClassInfo): boolean {
-    if (classInfo.isFinal) {
-      return true;
+  export var yieldVirtualMap = {
+    // These can technically yield but are worth the risk.
+    // "java/lang/Object.equals.(Ljava/lang/Object;)Z": YieldReason.None
```
mykmelez: Or perhaps not worth the risk? Would be good to have a comment here explaining why this is commented out.

jit/analyze.ts
```diff
@@ -76,14 +79,44 @@ module J2ME {
...
+      result = classInfo.subClasses.length === 0;
     }
+    // console.log(classInfo.className + " is final class " + result);
+    return result;
+  }
+
+  export function isFinalMethod(methodInfo: MethodInfo): boolean {
     return false;
-    // TODO: Be more clever here.
+    var result = methodInfo.isFinal;
```
mykmelez: Erm, are we now able to be more clever here? We're still returning false unconditionally.

main.js
```diff
@@ -250,15 +250,35 @@ window.onload = function() {
...
-   J2ME.interpreterMethodCounter && J2ME.interpreterMethodCounter.clear();
-   J2ME.baselineMethodCounter && J2ME.baselineMethodCounter.clear();
-
+   clearCounters();
  };
+
+  setInterval(function () {
+    function numberWithCommas(x) {
+      return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+    }
```
mykmelez: This function should get declared outside the setInterval call, so it doesn't have to be redeclared each time the interval function is called. But does this really need to get called unconditionally? Seems like it should be disabled by default.

mykmelez:
In this interest of moving the compiler work forward, I've addressed the issues I raised with commits 46b23b3dcab8fbc0a9e279e880652ed718590da1, 62007bd2fc721335a78cf1a534f41e40149adcfe, and bc56b09f8ac55666eb53cdf722759957e5645545.

PR #894, opened 2015-01-20
clarify licenses harder

mykmelez:
This mostly just empties out the top-level LICENSE file in favor of directory-specific files.


java/midp/com/sun/midp/l10n/LocalizedStringsBase.java
```diff
@@ -1,3 +1,29 @@
+/*
```
marco-c: Doesn't java/midp/LICENSE apply to this file as well?

style/emoji/LICENSE
```diff
@@ -0,0 +1,3 @@
+Copyright 2014 Twitter, Inc and other contributors
```
marco-c: (This only applies to the images in the sprites, as the sprites were created by me)

java/midp/com/sun/midp/l10n/LocalizedStringsBase.java
mykmelez: > Doesn't java/midp/LICENSE apply to this file as well?

Yes, but all the rest of the files also have their own license headers, which all say that "You should have received a copy of the GNU General Public License along with this program". Thus I'm adding a copy of the license itself in the LICENSE file.

style/emoji/LICENSE
mykmelez: Hmm, I'm not sure how best to represent this, but I modified the language in https://github.com/andreasgal/j2me.js/commit/d2f7aad7c04c6140d5ada011845ecf0801d66d95 to say:

> Adapted from graphics copyright 2014 Twitter, Inc and other contributors

How does that sound?

java/midp/com/sun/midp/l10n/LocalizedStringsBase.java
marco-c: OK. So, in the future, whenever we add a file from the reference implementation we need to add a license header if it isn't already present?

style/emoji/LICENSE
marco-c: Sounds good to me :smile:

marco-c:
Please merge this yourself once you've made sure we've covered everything.

java/midp/com/sun/midp/l10n/LocalizedStringsBase.java
mykmelez: > OK. So, in the future, whenever we add a file from the reference implementation we need to add a license header if it isn't already present?

Yes, but they should already be present. The only reason this one didn't have a header is that it's a generated file.

mykmelez:
> Please merge this yourself once you've made sure we've covered everything.

Will do, thanks!

java/midp/com/sun/midp/l10n/LocalizedStringsBase.java
marco-c: Ah, I see. (I'll file an issue to automatically generate this file)

mykmelez:
> Please merge this yourself once you've made sure we've covered everything.

I'm not sure we've covered quite everything, but we've covered a bunch more stuff, so I'm going to merge this, then I'll open another branch if needed to cover the rest.

PR #896, opened 2015-01-21
Remove xy section

marco-c:
This is no longer useful


mykmelez:
Looks fine, but needs a rebase.

Issue #866, opened 2015-01-15
Allow apps to exit


TimAbraldes:
When the last foreground MIDlet stops running (i.e. the last Isolate calls `stop` as implemented in #744), we should close the app.

I'm not sure if FirefoxOS has a mechanism that allows apps to close themselves so this might not be possible.

marco-c:
IIRC calling *window.close* closes the app

TimAbraldes:
> IIRC calling window.close closes the app

When I tried that I got a "Scripts may not close windows that were not opened by script" error. I'll keep investigating though

mykmelez:
> When I tried that I got a "Scripts may not close windows that were not opened by script" error. I'll keep investigating though

Maybe main.html needs to send a message to index.html via DumbPipe to close the window.

marco-c:
Maybe you called it in the mozbrowser? I think you need to call it in the parent window, so with a DumbPipe

TimAbraldes:
Gah, of course. I had forgotten about our index/main distinction.

However, I just tried adding this to index.js:

    DumbPipe.registerOpener("close", function(message, sender) {
      window.close();
    });

and calling that from code in native.js and I got the same error.

marco-c:
Mmmmh, it used to work before (when I was playing with it to support updates).

TimAbraldes:
Maybe it would work in the emulator and on device? I'm finding that it doesn't work in the browser, but I suppose that there's no meaningful way to "close" the emulator in the browser

marco-c:
> Maybe it would work in the emulator and on device? I'm finding that it doesn't work in the browser, but I suppose that there's no meaningful way to "close" the emulator in the browser

I've tested in the simulator, so that's probably the reason!

marco-c:
Fixed by https://github.com/andreasgal/j2me.js/pull/898

Issue #744, opened 2014-12-18
Implement missing native: com/sun/cldc/isolate/Isolate.stop.(II)V


marco-c:

marco-c:
@TimAbraldes the way isolates are implemented might be changing with the JIT PR (is it changing, @brendandahl ?)

brendandahl:
Isolates really haven't changed much. We've changed the isolate starting code a bit, but not much.

PR #898, opened 2015-01-21
Allow app to exit

yxl:
On Firefox OS, the app could be closed by calling `window.close()` from index.html. On desktop, unless the the app window is opened by `window.open`, we could not close the app window and nothing happens.

A few missing native methods are added to avoid error when app exits.

The previous implementation suppressed `missing native method` exceptions and it causes weird issues, such as the app hangs during exiting. I changes the implementation to throw these exceptions to let us discover such kinds of bugs earlier.

### DIFF
--- a/midp/midp.js
+++ b/midp/midp.js
@@ -867,6 +867,7 @@ Native["com/sun/midp/util/isolate/InterIsolateMutex.unlock0.(I)V"] = function(id
 };
 
 MIDP.exit = function(code) {
+    $.pause("exit");
     DumbPipe.open("exit", null, function(message) {});
 };
 
--- a/midp/midp.js
+++ b/midp/midp.js
@@ -1046,7 +1046,7 @@ Native["com/sun/midp/main/CommandState.saveCommandState.(Lcom/sun/midp/main/Comm
 
 Native["com/sun/midp/main/CommandState.exitInternal.(I)V"] = function(exit) {
     console.info("Exit: " + exit);
-    asyncImpl("V", new Promise(function(){}));
+    DumbPipe.open("exit", null, function(message) {});
 };
 
 Native["com/sun/midp/suspend/SuspendSystem$MIDPSystem.allMidletsKilled.()Z"] = function() {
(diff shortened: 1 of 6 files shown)


runtime.ts
```diff
@@ -1127,7 +1127,7 @@ module J2ME {
         // Some Native MethodInfos are constructed but never called;
         // that's fine, unless we actually try to call them.
         return function missingImplementation() {
-          release || assert (false, "Method " + methodInfo.name + " is native but does not have an implementation.");
+          assert (false, "Method " + implKey + " is native but does not have an implementation.");
```
mykmelez: Hmm, I'd rather stick with the convention that *assert* calls get compiled out in release builds (which is what `release || assert` does) and use *throw* here if we want to throw unconditionally. But the old VM merely logged via *console.error* in this case, so perhaps we should do the same.

mykmelez:
Is there an example midlet I can use to test this?

midp/midp.js
```diff
@@ -1145,7 +1145,7 @@ Native["com/sun/midp/main/CommandState.saveCommandState.(Lcom/sun/midp/main/Comm
 
 Native["com/sun/midp/main/CommandState.exitInternal.(I)V"] = function(exit) {
     console.info("Exit: " + exit);
-    asyncImpl("V", new Promise(function(){}));
+    DumbPipe.open("exit", null, function(message) {});
```
mykmelez: Perhaps we should also halt the VM at this point.

mykmelez:
@marco-c and @TimAbraldes are working on related changes that may be affected by this, but presumably they'll be able to rebase their changes after this lands.

marco-c:
Looks good to me, but I'm not sure this actually fixes #866 for all MIDlets. It does allow closing a normal MIDlet, but I think it doesn't take into account the case where there's a background and a foreground MIDlet (for this we also need #744).

runtime.ts
yxl: In commit  8bd0359, I replace the `assert` with a log as the old VM.

midp/midp.js
yxl: done with commit  c241350

yxl:
@mykmelez I tested with `/index.html?jars=tests/tests.jar&midletClassName=HelloMid&nativeMenu=1`:
1. open the hidden drawer menu by clicking the on the left of the header bar
2. you can see the `quit` menu item in the drawer.

yxl:
@marco-c I also add the native method `Isolate.stop` to fix #744.

marco-c:
There's one last thing to fix now that you've implemented Isolate.stop. Before closing the app, we must download and apply any queued update (in CldcPlatformRequest.dispatchPlatformRequest in midp/midp.js).

yxl:
Could we fix this with a followup?

Do you mean we should not exit until the updating is completed. It seems we need a prompt to tell use why we don't close the app immediately or ask if the user wants to wait for updating.

marco-c:
The problem is that this PR breaks updates, because the app exits while the update is being downloaded or applied.

Updates work in the following way:
1) A MIDlet calls platformRequest with a URL to a JAD
2) The native implementation queues an update to be downloaded/applied when the MIDlets exits
3) The MIDlet automatically exits
4) The native implementation performs the update

> Do you mean we should not exit until the updating is completed. It seems we need a prompt to tell use why we don't close the app immediately or ask if the user wants to wait for updating.

The update process starts when the user presses on an "update" button and the MIDlet exits immediately afterwards, so we don't need to prompt the user (as the user has already been asked by the MIDlet).

yxl:
@marco-c Thanks for your clarification. The last commit acae6d9 fixes the updating issue.

marco-c:
@yxl, I sent you a pull request (https://github.com/yxl/j2me.js/pull/1) to fix the update test.

yxl:
Great! Thanks for the quick fix:-)

PR #899, opened 2015-01-21
Enable JIT Compilation

mbebenita:
This PR enables JIT compilation by default. The current JIT heuristic is: if the number of invocations + number of backwards branches is greater than 10 then the method is compiled, unless its bytecode size is larger than 2000 bytes.


mykmelez:
I am groot.

PR #900, opened 2015-01-21
Fix copyright regression

marco-c:


mykmelez:
On Travis this fails with 99 different pixels. On my Mac, it's 178.

PR #902, opened 2015-01-22
Even more chrome fixes.

brendandahl:
Allows us to get to the second screen of a midlet.


brendandahl:
This will also need some of the work in https://github.com/mbebenita/j2me.js/commits/fixchrome

Issue #870, opened 2015-01-15
Disable logging in release builds


marco-c:
I think logging is pretty expensive, we should disable it in release builds.

PR #903, opened 2015-01-22
improve configuration

mykmelez:
This does a few different things.

First, it loads four files from the config/ subdirectory: default.js, midlet.js, build.js, and urlparams.js, which get loaded in that order, and each of which can override configuration from files earlier in the load order.

Then, it appends config/app.js to config/midlet.js on `make app`, so the app can have custom configuration (fixing the regression from removal of URL params from the launch path in manifest.webapp).

Finally, it auto-generates config/build.js based on the RELEASE variable and uses that variable to configure the log level (if it isn't otherwise set), so we can disable logging of anything but errors via `RELEASE=1 make` (addressing #870, at least to some extent; unsure if there's more to do).

We might want to munge all the config files together at some point rather than loading them all individually.


Makefile
```diff
@@ -47,6 +48,9 @@ closure: build/j2me.js aot
 shumway: $(SHUMWAY_SRCS)
 	node tools/tsc.js --sourcemap --target ES5 shumway/references.ts --out build/shumway.js
 
+build:
```
yxl: We already have a directory named `build` when do `make j2me`. To avoid confliction, we need to choose a different target name. With the name `build`, `make build` won't update `config/build.js` unless we recreate the `build` directory.

Also the target should be added to the .PHONY list

mykmelez: Good points! Fixed in 2b393a4.

PR #904, opened 2015-01-22
Chrome fixes

mbebenita:


mykmelez:
For some bizarre reason, this is causing a bunch of these failures in the tests, both on Travis and on my local machine:

> [debug] [phantom] Navigation requested: url=http://localhost:8000/main.html?downloadJAD=http://localhost:8000/tests/Manifest1.jad&midletClassName=tests.jaddownloader.AMIDletUpdater&logConsole=web,page, type=Undefined, willNavigate=true, isMainFrame=false
> Traceback (most recent call last):
>   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 295, in _handle_request_noblock
>     self.process_request(request, client_address)
>   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 321, in process_request
>     self.finish_request(request, client_address)
>   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 334, in finish_request
>     self.RequestHandlerClass(request, client_address, self)
>   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 657, in __init__
>     self.finish()
>   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 716, in finish
>     self.wfile.close()
>   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 279, in close
>     self.flush()
>   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 303, in flush
>     self._sock.sendall(view[write_offset:write_offset+buffer_size])
> error: [Errno 32] Broken pipe

index.js
```diff
@@ -14,6 +14,10 @@
                          loadScript("tests/index.js"));
   }
 
+  if (navigator.userAgent.indexOf('Chrome')) {
+    loadingPromises.push(loadScript("chrome_polyfills.js"));
+  }
```
mykmelez: @brendandahl This loads chrome_polyfills.js in Firefox too, because navigator.userAgent.indexOf('Chrome') == -1, which is truthy, if "Chrome" isn't in the user-agent string.

I'm still not sure why that causes this to fail, but tests/contacts.js also sets navigator.mozContacts to a new object, so perhaps the Chrome polyfill overrides that in a way that busts some test.

So I think there are two things to do here:

1. correct the conditional, so we only load the polyfill in Chrome;
2. make the polyfill also conditional on `midletClassName !== "RunTests"`, so we don't override the test polyfill when running tests in Chrome (if that even works).

PR #907, opened 2015-01-22
JIT Optimizations

mbebenita:
Depends on #904

### DIFF
--- a/utilities.ts
+++ b/utilities.ts
@@ -971,6 +971,34 @@ module J2ME {
   }
 
   export module HashUtilities {
+    // https://github.com/garycourt/murmurhash-js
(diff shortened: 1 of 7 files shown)


mykmelez:
Four tests are consistently failing in test runs, both on Travis and on my local machine.

mykmelez:
It looks like this reverts the chrome fixes we recently landed.

utilities.ts
```diff
@@ -971,6 +971,34 @@ module J2ME {
   }
 
   export module HashUtilities {
+    // https://github.com/garycourt/murmurhash-js
```
mykmelez: I think we'll need to do something a bit different here to reference the license. I'll ping my license guy.

PR #917, opened 2015-01-24
Allow MIDlets to run in the background

marco-c:
This is how MIDlets with a BG server work now:
1) If the user opens the app, the BG MIDlet and the FG MIDlet will be started automatically
2) If the app is awakened by an alarm, the BG MIDlet is started and the FG MIDlet will be started in one of these two cases:
  2.1) When the user "opens" the app by tapping on its icon
  2.2) When the user taps on a notification opened by the BG MIDlet after some event has occurred

We always prefer opening the FG MIDlet via launchIEMIDlet (because the BG MIDlet is passing some information about the events via launchIEMIDlet).

There's a couple of things I'd like to do in this PR (cleaning up logging from launchIEMIDlet, handling startupNoteText in launchIEMIDlet, choosing a value for the alarm timeout)

There's some followup work that I'd like to do after this PR:
 - remove config.pushMidlet and config.pushConn as they're unused now;
 - avoid showing the "Loading MIDlet" screen when only the BG MIDlet is running (but show something meaningful in case the user long presses on the home button to see the apps currently running)
 - maybe close the FG MIDlet when the user switches to another app (either immediately or after a timeout)


marco-c:
Forgot to mention that this is also going to start the FG MIDlet earlier than what we were doing before, so it should improve startup perf (but we should make sure there are no issues).

java/custom/com/nokia/mid/s40/bg/BGUtils.java
```diff
@@ -1,13 +1,52 @@
 package com.nokia.mid.s40.bg;
 
+import com.sun.midp.main.MIDletSuiteUtils;
+
 public class BGUtils {
+    private static boolean launchMIDletCalled = false;
+
+    public static native int getFGMIDletNumber();
+    public static native String getFGMIDletClass();
+    public static native void waitUserInteraction();
```
TimAbraldes: Are these used outside of this class? If not, let's make them private

java/custom/com/nokia/mid/s40/bg/BGUtils.java
```diff
@@ -1,13 +1,52 @@
...
+      String midletClass = BGUtils.getFGMIDletClass();
+
+      MIDletSuiteUtils.execute(midletNumber, midletClass, null);
     }
 
-    public static boolean launchIEMIDlet(String string1, String string2, int int1, String string3, String string4) {
-        throw new RuntimeException("BGUtils.launchIEMIdlet(L...String;L...String;IL...String;L...String;)Z " +
-                                   "not implemented (" + string1 + ", " + string2 + ", " + int1 + ", " + string3 +
-                                   ", " + string4 + ")");
+    public static native void addSystemProperties(String args);
```
TimAbraldes: Same public/private comment as above

java/custom/com/nokia/mid/s40/bg/BGUtils.java
```diff
@@ -1,13 +1,52 @@
...
-                                   "not implemented (" + string1 + ", " + string2 + ", " + int1 + ", " + string3 +
-                                   ", " + string4 + ")");
+    private static native void addSystemProperties(String args);
+
+    public static boolean launchIEMIDlet(String midletSuiteVendor, String midletName, int midletNumber, String startupNoteText, String args) {
+        System.out.println("midletSuiteVendor: " + midletSuiteVendor);
+        System.out.println("midletName: " + midletName);
+        System.out.println("midletNumber: " + midletNumber);
+        System.out.println("startupNoteText: " + startupNoteText);
+        System.out.println("args: " + args);
```
mykmelez: Added for debugging?

index.js
```diff
@@ -540,4 +540,14 @@ DumbPipe.registerOpener("reload", function(message, sender) {
...
-});
\ No newline at end of file
+});
+
+DumbPipe.registerOpener("backgroundCheck", function(message, sender) {
+  if (navigator.mozAlarms) {
+    navigator.mozAlarms.add(new Date(Date.now() + 10000), 'ignoreTimezone', {});
+
+    navigator.mozSetMessageHandler('alarm', function() {
+      navigator.mozAlarms.add(new Date(Date.now() + 10000), 'ignoreTimezone', {});
```
mykmelez: Nit: would be worth specifying the timeout in a variable.

Also, any thoughts on why you picked 10 seconds? It sounds a bit fast to me. In any case, it'd be good to include a comment explaining the reasoning here.

midp/background.js
```diff
@@ -0,0 +1,50 @@
...
+    util.fromJavaString(args).split(";").splice(1).forEach(function(arg) {
+        var elems = arg.split("=");
+        MIDP.additionalProperties[elems[0]] = elems[1];
+    });
+};
+
+Native["com/nokia/mid/s40/bg/BGUtils.waitUserInteraction.()V"] = function() {
+  asyncImpl("V", new Promise(function(resolve, reject) {
+    // If the page is visible, just start the FG MIDlet
+    if (document.hidden === false) {
```
mykmelez: Nit: *document.hidden* should only ever be a boolean, so you should be able to do simply `if (document.hidden)` here and below.

main.js
```diff
@@ -19,15 +19,6 @@ if (config.jars) {
...
 
-if (config.pushConn && config.pushMidlet) {
-  MIDP.ConnectionRegistry.addConnection({
-    connection: config.pushConn,
-    midlet: config.pushMidlet,
-    filter: "*",
-    suiteId: "1"
-  });
-}
-
```
mykmelez: Nit: there are also references to pushConn and pushMidlet in config/urlparams.js that should get removed.

java/custom/com/nokia/mid/s40/bg/BGUtils.java
marco-c: Yes, removing logging is included in the set of things I wanted to do in this PR (see my first comment)

index.js
marco-c: I picked 10 seconds for testing, I'm not sure what value would be good. Maybe 1 minute?

mykmelez: One minute sounds about right. We can always tune more later.

marco-c:
Turns out we need to wait for the localmsg server to be ready before starting the FG MIDlet (I've noticed a race condition while my laptop was plugged in)

main.js
marco-c: I've removed them, I'm now using fgMidletClass. We might be able to remove the parameter altogether, will do in a followup.

index.js
marco-c: Done in 2074d20

marco-c:
Unfortunately I couldn't find any info about the startupNoteText parameter, I'll file a followup.

mykmelez:
When testing this with a midlet, I'm seeing this error on startup, and the app stalls on the "Loading [NAME]" screen, although the throbber keeps spinning, and I see messages being logged to the console:

```
Error: java/lang/IllegalArgumentException: MIDlet class cannot be null
 - java/lang/Throwable.<init>(), pc=undefined
 - java/lang/Exception.<init>(), pc=undefined
 - java/lang/RuntimeException.<init>(), pc=undefined
 - java/lang/IllegalArgumentException.<init>(), pc=undefined
 - com/sun/midp/main/AmsUtil.executeWithArgs(), pc=undefined
 - com/sun/midp/main/MIDletSuiteUtils.executeWithArgs(), pc=undefined
 - com/sun/midp/main/MIDletSuiteUtils.executeWithArgs(), pc=undefined
 - com/sun/midp/main/MIDletSuiteUtils.executeWithArgs(), pc=undefined
 - com/sun/midp/main/MIDletSuiteUtils.executeWithArgs(), pc=undefined
 - com/sun/midp/main/MIDletSuiteUtils.execute(), pc=undefined
 - com/nokia/mid/s40/bg/BGUtils.startMIDlet(), pc=undefined
 - com/nokia/mid/s40/bg/WaitUserInteractionThread.run(), pc=undefined
 - com/nokia/mid/s40/bg/WaitUserInteractionThread.ThreadStart0Synthetic(), pc=undefined
 - com/sun/midp/main/AmsUtil.executeWithArgs(), pc=20
 - com/sun/midp/main/MIDletSuiteUtils.executeWithArgs(), pc=44
 - com/sun/midp/main/MIDletSuiteUtils.executeWithArgs(), pc=21
 - com/sun/midp/main/MIDletSuiteUtils.executeWithArgs(), pc=16
 - com/sun/midp/main/MIDletSuiteUtils.executeWithArgs(), pc=14
 - com/sun/midp/main/MIDletSuiteUtils.execute(), pc=10
 - com/nokia/mid/s40/bg/BGUtils.startMIDlet(), pc=21
 - com/nokia/mid/s40/bg/WaitUserInteractionThread.run(), pc=6
 - com/nokia/mid/s40/bg/WaitUserInteractionThread.ThreadStart0Synthetic(), pc=5

 j2me.js:6838:18
```

marco-c:
Have you replaced the pushMidlet parameter in your URL with fgMidletClass?

mykmelez:
> Have you replaced the pushMidlet parameter in your URL with fgMidletClass?

I hadn't, and that was the problem. Thanks!

PR #918, opened 2015-01-24
Reimplement setZeroTimeout with Promise.

yxl:
Using `Promise` is little lighter than `postMessage`.

### DIFF
--- a/main.html
+++ b/main.html
@@ -18,6 +18,7 @@
   <!--<script type="text/javascript" src="build/shumway.js" defer></script>-->
   <!--<script type="text/javascript" src="libs/terminal.js" defer></script>-->
   <script type="text/javascript" src="libs/console.js" defer></script>
+  <script type="text/javascript" src="libs/promise-6.0.0.js" defer></script>
   <script type="text/javascript" src="legacy.js" defer></script>
   <script type="text/javascript" src="blackBox.js" defer></script>
   <script type="text/javascript" src="timer.js" defer></script>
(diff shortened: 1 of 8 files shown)


mykmelez:
> Using Promise is little lighter than postMessage.

More than just a little! I see a 5-7x improvement in the time it takes to call the callback, in this test on my nightly build of Firefox: http://jsfiddle.net/wPCN4/7/

mykmelez:
This also reduces the overloading of *postMessage*, which has worried me (*DumbPipe* has to be careful to distinguish between messages from the parent frame and those that are zero timeouts). So it has multiple benefits.

The only issue with the branch is on browsers without built-in *Promise* (f.e. FxOS 1.3), where we use libs/promise-6.0.0.js to polyfill it. There, we need to make sure that timer.js loads after libs/promise-6.0.0.js, which is currently the case in index.html but not main.html.

(There are some test pages in tests/fs/ that currently load timer.js without loading libs/promise-6.0.0.js, and they should load libs/promise-6.0.0.js too.)

yxl:
e533c4a ensures `promise-6.0.0.js` is loaded before `time.js`.

mykmelez:
It's hard to tell, because the numbers are noisy, but I think I see some improvement in the performance of the fs tests. That would make sense, given that the fs API is mostly asynchronous. I'm working on making them less so, in which case the impact will be reduced, but this should still be beneficial all around. Nice find!

Note the discussion of Promise in [bug 686201](https://bugzilla.mozilla.org/show_bug.cgi?id=686201). In particular, @avih says:

> Interestingly, Firefox's native Promise.resolve().then(f) scheduling (at Firefox default settings) behaves very much like Firefox's setTimeout(f, 0) (with dom.min_timeout_value=0). I.e. parallel .then's are all executed before the browser UI is updated, and as such many synchronous .then's loop will hang the browser, but sequential(/nested) .then's don't get the 4ms clipping on one hand, while still letting the browser response/update to user inputs between each invocation (and not in a sluggish way either - properly responsive, upto the duration which each chunk executes).

To which @bzbarsky responds:

> The exact interaction of Promise stuff with the event loop will likely get pinned down in the spec at some point here; what we're doing right now won't match that and we'll need to change it.  In particular, the "nested" behavior you observe may no longer happen because afaict the spec requires pending Promise callbacks to be serviced before the main event loop.

So we should watch out for changes to the behavior in future versions of Gecko. But this should work for now, given the current implementation and the way we use *setZeroTimeout*.

Issue #920, opened 2015-01-25
remove unnecessary synchronization


mykmelez:
The gfx Java code is full of synchronized statements and methods. There are a bunch in Graphics.java, which @TimAbraldes is removing in #882, but also in Display.jpp, Displayable.java, Canvas.java, and elsewhere.

But our VM isn't truly multi-threaded, so most of them are unnecessary, and they make the VM insert yielding code and do extra work. It isn't clear how expensive this is, but we should find out by removing the unnecessary ones, i.e. the ones around code that never calls into a native that yields.

andreasgal:
Even without performance downside yields can cause flickering if we don't manually double buffer the canvas.

andreasgal:
I wonder whether its worth to do compile time analysis instead of a manual removal. @mbebenita?

PR #921, opened 2015-01-25
Thread scheduler

yxl:
It is a simple thread scheduler to let the thread with higher priority run first.

### DIFF
--- a/runtime.ts
+++ b/runtime.ts
@@ -680,11 +680,10 @@ module J2ME {
       (<any>window).setZeroTimeout(function() {
         try {
           Runtime._runningQueue.dequeue().execute();
-        } catch (e) {
-          // Suppress the exception.
-        }
-        if (!Runtime._runningQueue.isEmpty()) {
-          Runtime.processRunningQueue();;
+        } finally {
+          if (!Runtime._runningQueue.isEmpty()) {
+            Runtime.processRunningQueue();
+          }
         }
       });
     }
--- a/runtime.ts
+++ b/runtime.ts
@@ -819,7 +819,7 @@ module J2ME {
     export interface Thread extends java.lang.Object {
       pid: number;
       alive: boolean;
-      $priority: number;
+      priority: number;
     }
 
     export interface Exception extends java.lang.Object {
--- a/runtime.ts
+++ b/runtime.ts
@@ -610,11 +610,19 @@ module J2ME {
       }
     }
 
-    enqueue(ctx: Context) {
+    /*
+     * @param jump If true, move the context to the first of others who have the
+     * same priority.
+     */
+    enqueue(ctx: Context, jump: boolean) {
       var priority = ctx.getPriority();
       release || assert(priority >= MIN_PRIORITY && priority <= MAX_PRIORITY,
                         "Invalid priority: " + priority);
-      this._queues[priority].push(ctx);
+      if (jump) {
+        this._queues[priority].unshift(ctx);
+      } else {
+        this._queues[priority].push(ctx);
+      }
       this._top = Math.max(priority, this._top);
     }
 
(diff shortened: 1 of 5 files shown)


mykmelez:
Our VM has done surprisingly well without any thread scheduling to date, but I've been nervous about it, and I wonder if it has anything to do with some reports of long UI pauses from a few testers. At least one midlet we're testing is setting priorities on the threads it creates, so this seems like a wise addition.

And the implementation looks reasonable to me. My only concern would be whether this algorithm, by consistently choosing the highest-priority thread to resume, could starve low-priority threads, given frequent-enough scheduling of high-priority threads.

But I'm not the expert here. @mbebenita, @brendandahl: what do y'all think?

andreasgal:
Lets look up what J2ME actually does.

mykmelez:
This article [Using Threads in J2ME Applications](http://www.oracle.com/technetwork/systems/threading2-156909.html) has a bit of info:

> Thread scheduling is a complicated affair over which the application has little control. The system (either the underlying operating system or the Java virtual machine, depending on how the threads are implemented) gives each ready thread a chance to run for a short time (the suspended threads are ignored), switching rapidly from one thread to another. This context switching can occur at any time. The only real influence the application has over thread scheduling is via the thread's priority level: **higher-priority threads execute more often than those with lower priority.** *[emphasis mine]*

yxl:
@mbebenita Added `priority` field to bindings with commit 96d38ea

yxl:
@mykmelez  The JVM specification doesn't define a single thread scheduler for all JVM implementation. Most JVMs depend on the platform or OS to provide thread scheduler. 

This article from javaworld.com gives an overview about the JVM thread scheduler:
http://www.javaworld.com/article/2071214/java-concurrency/java-101--understanding-java-threads--part-3--thread-scheduling-and-wait-notify.html

In general,  two threading scheduler models are provided for java:
 1. Native thread scheduling
 2. Green thread scheduling

The native thread scheduling algorithms are mostly timesharing. They are system dependent, provided by the Windows, Linux... This model could prevent low-priority threads from starving and proves to be efficient. But quite complex for js to implement.

The [green thread scheduling](http://en.wikipedia.org/wiki/Green_threads) looks to be a good choice for us. It emulates multiple threads on the single thread platform, which is quite similar to the environment of j2me.js. This model is also served as an option for the JVM of for oracle weblogic server: https://docs.oracle.com/cd/E13222_01/wls/docs60/perform/JVMTuning.html

The previous article from javaworld.com explains how it work:
>A JVM's thread scheduler schedules green threads according to priority—a thread's relative importance, which you express as an integer from a well-defined range of values. Typically, a JVM's thread scheduler chooses the highest-priority thread and allows that thread to run until it either terminates or blocks. At that time, the thread scheduler chooses a thread of the next highest priority. That thread (usually) runs until it terminates or blocks. If, while a thread runs, a thread of higher priority unblocks (perhaps the higher-priority thread's sleep time expired), the thread scheduler preempts, or interrupts, the lower-priority thread and assigns the unblocked higher-priority thread to the processor.

The final scheduling algorithm may be as following:
http://lass.cs.umass.edu/~shenoy/courses/fall01/labs/talab2.html

*Features:*

1. The JVM schedules using a preemptive , priority based scheduling algorithm.
2. All Java threads have a priority and the thread with he highest priority is scheduled to run by the JVM.
3. In case two threads have the same priority a FIFO ordering is followed.

*A different thread is invoked to run in case one of the following events occur:*

1. The currently running thread exits the Runnable state ie either blocks or terminates.
2. A thread with a higher priority than the thread currently running enters the Runnable state. The lower priority thread is preempted and the higher priority thread is scheduled to run.

*Time Slicing is dependent on the implementation:*

A thread can voluntarily yield control through the yield() method. Whenever a thread yeilds control of the CPU another thread of the same priority is scheduled to run. A thread voluntarily yielding control of the CPU is called Cooperative Multitasking.

TimAbraldes:
If we implement pre-empting at arbitrary points in Java execution, we can't also implement #920.

If we continue to only switch threads when a yield occurs, then we should be OK

mykmelez:
> If we implement pre-empting at arbitrary points in Java execution, we can't also implement #920.

Fortunately (or unfortunately?) this would be difficult to do and isn't on the radar.

> If we continue to only switch threads when a yield occurs, then we should be OK

We might be able to be more aggressive while still being ok, provided we choose when to switch and which methods/block to remove synchronization from wisely.

runtime.ts
```diff
@@ -592,10 +592,101 @@ module J2ME {
...
+        Runtime.processRunningQueue();
+      }
+    }
+
+    private static processRunningQueue() {
+      (<any>window).setZeroTimeout(function() {
+        try {
+          Runtime._runningQueue.dequeue().execute();
+        } catch (e) {
+          // Suppress the exception.
```
brendandahl: Won't this lead to silent failures? Could we just put lines 684-686 into a finally block?

runtime.ts
```diff
@@ -592,10 +592,101 @@ module J2ME {
...
+        }
+        if (!Runtime._runningQueue.isEmpty()) {
+          Runtime.processRunningQueue();;
```
brendandahl: Is there a benefit to executing the next context on the next event tick? We could just execute multiple contexts and drain the queue in a loop here.

runtime.ts
yxl: Fixed.

runtime.ts
yxl: It is not only a benefit, but a requirement. If we drain the queue, next time when a new context with highest priority runs into, we cannot make it run before other pending contexts.

yxl:
>If we implement pre-empting at arbitrary points in Java execution, we can't also implement #920.
Current implementation only pre-empt when creating a new thread(by calling `Thread.start()`), so that's not a problem.

mbebenita:
> Fortunately (or unfortunately?) this would be difficult to do and isn't on the radar.

Actually this wouldn't be that hard to do. We can emit code at loop headers that checks a timer and yields periodically. Not sure if it's worth it though.

mykmelez:
> Actually this wouldn't be that hard to do. We can emit code at loop headers that checks a timer and yields periodically. Not sure if it's worth it though.

Hmm, there's a known hang in a midlet that seems to have a big impact, and I suspect a long-running loop, so this may well be worth it.

PR #923, opened 2015-01-26
JIT Relooper Optimizations

mbebenita:
This PR uses Emscripten's relooper to restructure control flow in the baseline JIT. The quality of the generated code now pretty much as good as it gets without doing more aggressive Java level optimizations. Compiles all methods, and passes most tests including various midlets.

Known Issues:
There are some problems with console logging. The relooper code tries to log messages before `console.js` is property initialized.


mykmelez:
I read through this and tested it on device, then @mbebenita and I talked through the various bits, including the relooper, which is fairly mature code from emscripten.

This is the last major chunk of changes to the VM, after which there are still a number of small-scale optimizations to be made. File issues on those as you think of them. And we'll keep iterating on other perf improvements.

PR #924, opened 2015-01-26
make fs API mostly synchronous

mykmelez:
Here's an updated version of the branch on which I minimize asynchronous calls to the virtual filesystem. With these changes, the only asynchronous fs calls are those that open files, since we read file contents with *FileReader*, which only has an asynchronous API. (In theory, we could open synchronously and only read asynchronously.)

This should make fs calls faster, modulo the time it takes to read in the file records on startup, which is about 1.1 - 1.3s for ~1300 records in a filesystem populated by a file-happy midlet on a Flame. And I do see plenty of improvement on the fs tests themselves.

Nevertheless, I see slower startup times for that midlet on that Flame with this branch. Here are some samples, in seconds from tapping on the homescreen icon until the midlet interface appears:

* Before: 20, 22, 23, 22, 23
* After: 25, 24, 24, 25, 24

Discarding the outlier, it looks like the branch slows down startup of that midlet by ~2s, which means the slowdown is greater than just the cost of loading all the records from the datastore. I'm not sure why that is. Perhaps reduced yielding on background threads is starving the UI?


mykmelez:
Warning! I needed to update the IndexedDB version to optimize retrieval of records on startup, so if you test this patch, you'll need to jump through some hoops to revert your profile to an earlier version of the database. I highly recommend exporting the filesystem before testing it, via either the development UI in the browser or [WebIDE on an FxOS device](https://github.com/andreasgal/j2me.js/wiki/Exporting-Importing-the-File-System-on-FxOS).

mykmelez:
I did some more manual stopwatch timings of startup for that IO-happy midlet with a populated filesystem, and the numbers are noisy, but what I think they show is that this "sync-fs" branch by itself regresses startup time, and @yxl's "thread-scheduler" branch in #921 doesn't improve startup time on its own, but the two branches together improve startup time over master (and may also make it more consistent, although it's hard to extrapolate from these timings):

```
                  master: 22, 24, 21, 25, 25, 22, 25, 26
        thread-scheduler: 27, 24, 22, 22, 22, 22, 24, 24, 24, 26, 24, 24
                 sync-fs: 27, 27, 26, 25, 26, 26, 27
sync-fs-thread-scheduler: 21, 22, 21, 22, 21, 22, 21
```

marco-c:
Assuming the measurements are mostly correct, the sync-fs-thread-scheduler is statistically better than master (at 95% of confidence)

mykmelez:
@marco-c Shall we take this? It seems to provide some improvement in startup times (combined with @yxl's scheduler, now landed), and it should theoretically improve overall perf.

marco-c:
Yes, I'll review it later today or tomorrow morning

libs/fs.js
```diff
@@ -139,36 +170,31 @@ var fs = (function() {
...
-      };
-      transaction.oncomplete = (function() {
-        if (DEBUG_FS) { console.log("get " + key + " completed"); }
-        var value = req.result;
-        if (value === undefined) {
-          value = null;
-        }
-        this.map.set(key, value);
-        cb(value);
+      var then = performance.now();
```
marco-c: Nit: call performance.now() only if DEBUG_FS is true

marco-c:
Looks good to me. Could you update it?
I'll test it locally after you've updated it.

My only concern is that by reading the whole store at startup, we might end up using too much memory (even if the content of the files isn't going to be read at startup).

mykmelez:
> Looks good to me. Could you update it?
> I'll test it locally after you've updated it.

Updated!

> My only concern is that by reading the whole store at startup, we might end up using too much memory (even if the content of the files isn't going to be read at startup).

Hmm, file records are pretty small, modulo actual file content, just: { pathname: string, parentDir: string, isDir: boolean, mtime: number }. So I'm not worried about memory usage.

But I do worry that reducing yielding will enable background threads to starve the UI thread. I'd be more comfortable with this if we had a strategy for dealing with that generally, like @mbebenita's work to enable yielding at long-running loop headers.

PR #932, opened 2015-01-26
Fix font size

marco-c:
Using pixels instead of points fixes a problem I'm seeing with rendering.


marco-c:
I'll probably need to fix some gfx tests, will do after #933 is merged.

PR #936, opened 2015-01-27
If MCC and MNC are not available, return `null`

### LINKED ISSUE #550: java.lang.RuntimeException: RangeError: repeat count must be non-negative
This happens when you run j2me.js on a phone with no SIM.
The error is in the *pad* function in *util.js*, because the values that we're trying to pad (mobileInfo.network.mcc and mobileInfo.network.mnc) are undefined

TimAbraldes:
Fixes #550

### DIFF
--- a/native.js
+++ b/native.js
@@ -198,12 +198,20 @@ Native["java/lang/System.getProperty0.(Ljava/lang/String;)Ljava/lang/String;"] =
         value = "1.7";
         break;
     case "com.nokia.mid.mnc":
-        // The concatenation of the MCC and MNC for the ICC (i.e. SIM card).
-        value = util.pad(mobileInfo.icc.mcc, 3) + util.pad(mobileInfo.icc.mnc, 3);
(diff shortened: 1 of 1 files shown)


marco-c:
Actually, I think it'd be better to fix the library libs/mobileInfo.js to allow people with no sim card to use the mnc and mcc as url parameters (like on desktop).

TimAbraldes:
> Actually, I think it'd be better to fix the library libs/mobileInfo.js to allow people with no sim card to use the mnc and mcc as url parameters (like on desktop).

I'd be open to allowing mnc and mcc to be specified somehow, but I thought that we stopped using URL parameters in #851. Also, it seems unlikely that users will know what values to use for mnc and mcc; maybe there's some more friendly way we can expose this ability?

marco-c:
Sorry, I meant it would be good for testing :smile: (where you can still use url parameters)

TimAbraldes:
Ah that makes a lot more sense :smile:. Filed #944

PR #938, opened 2015-01-27
compiled method cache

mykmelez:
@mbebenita This is a work-in-progress compiled method cache. Doesn't yet do invalidation.

To clear your cache from the Web Console:

```
cd(frames[0]); // changes to main.html
CompiledMethodCache.clear();
```


mbebenita:
In this PR we still need to exceed the JIT compilation threshold before we link in the cached method. This means that we interpret a significant number of bytecodes before we ever decide to JIT (10 method invocations or backwards branches). We should also check the method cache in the `findCompiledMethod` function.

The mechanism for handling referenced classes is different there (it used to be per class, not per method) and we should fix that in a separate PR.

mykmelez:
> In this PR we still need to exceed the JIT compilation threshold before we link in the cached method. This means that we interpret a significant number of bytecodes before we ever decide to JIT (10 method invocations or backwards branches). We should also check the method cache in the findCompiledMethod function.

Ok, I've added checking of the cache in *findCompiledMethod*. I also added invalidation whenever the app or midlet version changes (and updating of the app version on *make all*, so the cache will be invalidated whenever you rebuild the project, since Java code could have changed). And I added a "Clear Compiled Method Cache" button to the debugging panel, so you can clear the cache manually.

mykmelez:
> We should create a TS interface for this external definition.

Added in 992495a.

And 49bcd92 grabs a change from @mbebenita that is needed to prevent an error during AOT compilation.

Ok, I think this is ready to go.

mykmelez:
This branch improves startup time on subsequent launch in a midlet I'm testing. Here are some numbers from stopwatch timings, in seconds:

> before: 22, 22, 22
> after: 20, 20, 20

mykmelez:
4990b42 cherry-picks a change from #952 to fix the intermittent test failure that keeps breaking the builds for this branch.

mykmelez:
I'm not sure how indicative the basic unit tests are, but they show a perf improvement in the Travis test run, which runs them twice:

> DONE: 71548 pass, 0 fail, 214 known fail, 0 unknown pass, 36489ms
> DONE: 71548 pass, 0 fail, 214 known fail, 0 unknown pass, 31859ms

mykmelez:
Whereas [this Travis run on master](https://travis-ci.org/mykmelez/j2me.js/builds/48576992) shows similar results for both runs of the basic unit tests:

> DONE: 71548 pass, 0 fail, 214 known fail, 0 unknown pass, 33149ms
> DONE: 71548 pass, 0 fail, 214 known fail, 0 unknown pass, 33493ms

I'm not sure how comparable the numbers are across different Travis runs, since the machines could be different (and Travis could be experiencing different load). Nevertheless, those numbers suggest that there's a perf hit on firstrun (which is expected, given that we have to persist the compiled functions to the datastore) and then a win on subsequent run.

Issue #867, opened 2015-01-15
Errors in natives are really difficult to debug


marco-c:
For example, this is what I see if I use an undefined variable (in waitForNativeEvent in midp.js):

    Error: assertion failed                                      j2me.js:223:18

marco-c:
This is now fixed.

PR #939, opened 2015-01-27
Show errors occurred in native methods. #867

yxl:
`RELEASE=0 make` show errors in native methods.
`RELEASE=1 make` hide the errors as the original implementation.


runtime.ts
```diff
@@ -1176,15 +1176,26 @@ module J2ME {
   }
 
   function findNativeMethodImplementation(methodInfo: MethodInfo) {
+    function wrap(method, key) {
```
mykmelez: Nit: I would call this something like reportError to make its purpose clearer.

Also, we should declare this outside the findNativeMethodImplementation function, so we don't have to redeclare it each time that function is called.

utilities.ts
```diff
@@ -20,8 +20,8 @@ var inBrowser = typeof console !== "undefined" && console.info;
 declare var putstr;
 declare var printErr;
 
-/** @const */ var release: boolean = true;
-/** @const */ var profile: boolean = true;
+/** @const */ var release: boolean = jsGlobal.config.release;
+/** @const */ var profile: boolean = !release;
```
mykmelez: @mbebenita mentioned that doing this will prevent Closure from compiling out the various assertions that are prefixed with `release ||`, so we should figure out some different way to apply the value of RELEASE here. Maybe we can move these declarations to a separate config.ts file that we include in j2me.js by adding it to references.ts.

mbebenita: Yes, there is quite a bit of code that is removed by closure if it can prove that `release` is false, it's not just assertion code.

runtime.ts
yxl: Fixed in 19539cb

Makefile
```diff
@@ -1,13 +1,23 @@
...
+PROFILE ?= $(if $(findstring 0,$(RELEASE)),1,0)
+
+toBool = $(if $(findstring 1,$(1)),true,false)
+PREPROCESS_DEFS = -D RELEASE=$(call toBool,$(RELEASE)) \
+                  -D PROFILE=$(call toBool,$(PROFILE))
+PREPROCESS_SRCS = $(shell find . -name "*.in")
+PREPROCESS_DESTS = $(PREPROCESS_SRCS:.in=)
 
 all: config-build java jasmin tests j2me shumway aot
+	echo $(PREPROCESS_SRCS)
```
mykmelez: Is this needed, or was it just for debugging?

Makefile
```diff
@@ -1,13 +1,23 @@
...
 
 test: all
-	tests/runtests.py
```
mykmelez: This change breaks running tests with `make test`.

Makefile
```diff
@@ -47,8 +57,13 @@ shumway: build/shumway.js
 build/shumway.js: $(SHUMWAY_SRCS)
 	node tools/tsc.js --sourcemap --target ES5 shumway/references.ts --out build/shumway.js
 
+CONFIG_BUILD_CONTENT := "config.release = ${RELEASE};"
+OLD_CONFIG_BUILD_CONTENT := "$(if $(wildcard config/build.js),$(shell cat config/build.js),)"
+
 config-build:
+ifneq ($(CONFIG_BUILD_CONTENT),$(OLD_CONFIG_BUILD_CONTENT))
 	echo "config.release = ${RELEASE};" > config/build.js
+endif
```
mykmelez: Hmm, if we're also going to modify the way that the *config-build* target produces config/build.js, we might as well start change it to use a preprocessed file (config/build.js.in).

tools/preprocess-1.1.0/README.txt
```diff
@@ -0,0 +1,412 @@
+preprocess.py -- a portable multi-language file preprocessor
+============================================================
+
+Download the latest preprocess.py packages from here:
+    (archive) http://preprocess.googlecode.com/files/preprocess-1.2.0.zip
```
mykmelez: Hunh, strange that this references a package for version 1.2.0, which doesn't exist.

tools/preprocess-1.1.0/lib/preprocess.py
```diff
@@ -0,0 +1,847 @@
+#!/usr/bin/env python
```
mykmelez: Thanks for integrating this, it's helpful! I was actually thinking about integrating a preprocessor when I first landed support for producing build/config.js.

However, to minimize the number of files in our repo, let's include just the necessary files from this third-party codebase:

* tools/preprocess-1.1.0/lib/preprocess.py
* tools/preprocess-1.1.0/LICENSE.txt
* tools/preprocess-1.1.0/README.txt

Makefile
yxl: That's for debugging and should be removed.

Makefile
yxl: Fixed by 6ba5f42

Makefile
yxl: Fixed in 87157a1

tools/preprocess-1.1.0/README.txt
yxl: It is a mistake of the original library and I fixed in 8c4fe34

tools/preprocess-1.1.0/lib/preprocess.py
yxl: Extra files are removed by ddcff13

Makefile
```diff
@@ -1,14 +1,32 @@
...
+PROFILE ?= $(if $(findstring 0,$(RELEASE)),1,0)
 
-all: config-build java jasmin tests j2me shumway aot
+# Create a checksum file to monitor the changes of the Makefile configuration.
+# If the configuration has changed, we update the checksum file to let the files
+# which depend on it to regenerate.
+
+CHECKSUM := "$(RELEASE)$(PROFILE)"
+OLD_CHECKSUM := "$(shell [ -f .checksum ] && cat .checksum)"
+$(shell [ $(CHECKSUM) != $(OLD_CHECKSUM) ] && echo $(CHECKSUM) > .checksum)
```
mykmelez: Ah, nice. I'd been wondering how we can do this. This is simple and elegant. Let's remember to update it when we add more build configuration variables.

runtime.ts
```diff
@@ -1142,16 +1142,27 @@ module J2ME {
     return null;
   }
 
+  function reportError(method, key) {
+    return function() {
+      try {
+        return method.apply(this, arguments);
+      } catch (e) {
+        stderrWriter.errorLn("Native " + key + " throws:" + e);
```
mykmelez: With release=0 (the default), I see a bunch of these calls, but *e* stringifies to `[object Object]` for Java exceptions, which isn't very informative. Can we stringify them to something more detailed? Something like this should work:

```
      } catch (e) {
        var msg = e;
        if (e.klass) {
          msg = e.klass.classInfo.className + " " +
                util.fromJavaString(CLASSES.getField(e.klass.classInfo, "I.detailMessage.Ljava/lang/String;").get(e));
        }
        stderrWriter.errorLn("Native " + key + " throws: " + msg);
        throw e;
      }
```

yxl: The Java exceptions are not unexpected error from the native method. They should be handled properly by the outer java functions.

So I fix this by not reporting the Java exceptions: 3b2991ce093.

mykmelez: > The Java exceptions are not unexpected error from the native method. They should be handled properly by the outer java functions.

Ah, good point, makes sense. I see a few of these in tests and filed #969 on fixing those.

PR #941, opened 2015-01-27
Coerce return values to int in MediaDownload::nGetFirstPacketSize and Me...

marco-c:
...diaDownload::nBuffering


brendandahl:
We could use the new checks from https://github.com/andreasgal/j2me.js/pull/943/files#diff-50bf7a156821b9fc15f03a06151db638R92

mykmelez:
> If you disable release mode (in utilities.ts), a checkReturnValue function in the interpreter can check for invalid return values. It doesn't test for integer return values, but it can easily be extended.
…
> We could use the new checks from https://github.com/andreasgal/j2me.js/pull/943/files#diff-50bf7a156821b9fc15f03a06151db638R92

Let's add type checking in a followup.

PR #950, opened 2015-01-28
Remove un-needed callback and timeout.

brendandahl:
The wakeup() code will never(and should never) throw exceptions so we no longer need to do the setTimeout. Also, removing the block() callbacks helped with reduce allocations in chrome's memory profiler.

### DIFF
--- a/context.ts
+++ b/context.ts
@@ -598,13 +598,12 @@ module J2ME {
       $.pause("block");
     }
 
-    unblock(obj, queue, notifyAll, callback) {
+    unblock(obj, queue, notifyAll) {
       while (obj[queue] && obj[queue].length) {
         var ctx = obj[queue].pop();
         if (!ctx)
           continue;
-        // Wait until next tick, so that we are sure to notify all waiting.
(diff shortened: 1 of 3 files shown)


brendandahl:
Looks like the extra pass is coming from https://github.com/andreasgal/j2me.js/blob/76a831c988b5e33d8c57e4b6a9c2a06ebd7fe3d0/tests/javax/microedition/media/TestMediaImage.java#L63
@marco-c @mykmelez Looks like you both have changed this file. Thoughts on if this should be happening?

mykmelez:
> @marco-c @mykmelez Looks like you both have changed this file. Thoughts on if this should be happening?

I'm guessing that line is only supposed to be called once.

marco-c:
You could try modifying TestMediaImage to synchronize like TestPlayerListener (I don't recall what events were supposed to be sent, you could modify the test before these changes and then make sure the same events are sent before and after the changes)

mykmelez:
There are some intermittent test failures after #955 and then this PR were merged. I filed this as #960.

PR #953, opened 2015-01-28
yield periodically while writing to FCOutputStream

mykmelez:
@marco-c , @mbebenita and I discussed this on IRC as a kind of hacky fix for the problem of a hot loop that writes large amounts of data to the filesystem without yielding. It yields every 10ms.

We could make this more precise: tracking the last time we wrote to each file descriptor, assuming that such loops are always writing to the same one. But it's reasonable to do, if we don't have a better way to pre-empt those loops.


mykmelez:
Closing this until we measure the wins from other fixes.

Issue #960, opened 2015-01-28
a variety of intermittent test failures after #955 and #950 were merged


mykmelez:
@mbebenita , @brendandahl The merge commit for #955 failed on Travis:

```
com.sun.midp.util.isolate.TestInterIsolateMutex-0: fail [Exception expected]
com.sun.midp.util.isolate.TestInterIsolateMutex: 2 pass, 1 fail, 0 unknown pass
…
gnu.testlet.vm.DupTest-0: fail
gnu.testlet.vm.DupTest-1: got (7), expected (5)
gnu.testlet.vm.DupTest-1: fail
gnu.testlet.vm.DupTest-2: got (8), expected (6)
gnu.testlet.vm.DupTest-6: fail
gnu.testlet.vm.DupTest-7: got (6), expected (4)
gnu.testlet.vm.DupTest-7: fail
gnu.testlet.vm.DupTest-8: got (7), expected (5)
gnu.testlet.vm.DupTest: 4 pass, 4 fail, 0 unknown pass
gnu.testlet.vm.NestedExceptionTest-3: fail [finallyReturn hasn't yet returned a value]
gnu.testlet.vm.NestedExceptionTest-6: fail [finallyThrow hasn't yet thrown an exception]
gnu.testlet.vm.NestedExceptionTest: 12 pass, 2 fail, 0 unknown pass
…
DONE: 74369 pass, 7 fail, 229 known fail, 0 unknown pass, 34222ms
…
FAIL 
  Expected 71548 pass. Got 74369
  Expected 0 fail. Got 7
  Expected 214 known fail. Got 229
```

- https://travis-ci.org/andreasgal/j2me.js/builds/48667391

And there are more failures in the merge commit for #950, which landed afterward:

```
gnu.testlet.vm.DupTest-0: fail
gnu.testlet.vm.DupTest-1: got (7), expected (5)
gnu.testlet.vm.DupTest-1: fail
gnu.testlet.vm.DupTest-2: got (8), expected (6)
gnu.testlet.vm.DupTest-6: fail
gnu.testlet.vm.DupTest-7: got (6), expected (4)
gnu.testlet.vm.DupTest-7: fail
gnu.testlet.vm.DupTest-8: got (7), expected (5)
gnu.testlet.vm.DupTest: 4 pass, 4 fail, 0 unknown pass
gnu.testlet.vm.NestedExceptionTest-3: fail [finallyReturn hasn't yet returned a value]
[...]

PR #961, opened 2015-01-28
Support selecting 'Sound' in the nokia.file-ui localmsg server

marco-c:


marco-c:
The failures are the same of #960.

PR #965, opened 2015-01-28
polyfill String.fromCodePoint and String.prototype.codePointAt for FxOS 1.3

mykmelez:
These are needed for FxOS 1.3. They come from these MDN pages:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt

I decided to create a new *polyfill* directory so we can distinguish between the various polyfills and libraries. We should move other polyfills there too (chrome_polyfills.js, libs/promise-6.0.0.js, etc.).

### DIFF
--- a/main.html
+++ b/main.html
@@ -19,6 +19,8 @@
   <!--<script type="text/javascript" src="libs/terminal.js" defer></script>-->
   <script type="text/javascript" src="libs/console.js" defer></script>
   <script type="text/javascript" src="libs/promise-6.0.0.js" defer></script>
+  <script type="text/javascript" src="polyfill/fromcodepoint.js" defer></script>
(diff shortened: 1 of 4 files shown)


mykmelez:
While testing a midlet's basic functionality, these were the only issues I found. There may of course be others.

marco-c:
We should probably merge these files in the packaging step as a followup.

mykmelez:
> We should probably merge these files in the packaging step as a followup.

Yes, along with all the other scripts, and then run them through the Closure compiler.

PR #966, opened 2015-01-29
Misc keyboard and canvas fixes

TimAbraldes:
Should prevent the canvas from being accidentally resized. Keeps the MIDlet's copies of the screen and keyboard size in sync with midp.js's copies.


midp/midp.js
```diff
@@ -1340,3 +1355,138 @@ addUnimplementedNative("com/sun/j2me/content/InvocationStore.get0.(Lcom/sun/j2me
...
+        MIDP.updateCanvas();
+        MIDP.sendNativeEvent({ type: MIDP.ROTATION_EVENT, intParam1: 0, intParam2: 0, intParam3: 0, intParam4: MIDP.displayId }, MIDP.foregroundIsolateId);
+    }
+};
+
+Native["com/nokia/mid/ui/VirtualKeyboard.getCustomKeyboardControl.()Lcom/nokia/mid/ui/CustomKeyboardControl;"] = function() {
+    throw $.newRuntimeException("VirtualKeyboard::getCustomKeyboardControl() not implemented");
+};
+
+Native["com/nokia/mid/ui/VirtualKeyboard.hideOpenKeypadCommand.(Z)V"] = function(bl) {
```
yxl: You could use `UnimplementedNative()` to simplify this.

midp/gfx.js
```diff
@@ -24,14 +24,7 @@ var currentlyFocusedTextEditor;
...
     function setFullScreen(isFullScreen) {
-        var sidebar = document.getElementById("sidebar");
-        var header = document.getElementById("drawer").querySelector("header");
-        var canvas = MIDP.Context2D.canvas;
-        sidebar.style.display = header.style.display =
-            isFullScreen ? "none" : "block";
-        var headerHeight = isFullScreen ? 0 : header.offsetHeight;
-        canvas.height = MIDP.ScreenHeight - headerHeight;
-        canvas.dispatchEvent(new Event('canvasresize'));
+        MIDP.setFullScreen(isFullScreen);
```
yxl: The `setFullScreen()` in gfx.js can be remove and call `MIDP.setFullScreen()` directly instead.

midp/midp.js
```diff
@@ -1340,3 +1355,138 @@ addUnimplementedNative("com/sun/j2me/content/InvocationStore.get0.(Lcom/sun/j2me
 addUnimplementedNative("com/sun/j2me/content/InvocationStore.getByTid0.(Lcom/sun/j2me/content/InvocationImpl;II)I", 0);
 addUnimplementedNative("com/sun/j2me/content/InvocationStore.resetFlags0.(I)V");
 addUnimplementedNative("com/sun/j2me/content/AppProxy.midletIsRemoved.(ILjava/lang/String;)V");
+
+
+var isFullScreen = false;
```
yxl: There is no closure for MIDP.js. To avoid potential variable collision, it'd better make `isFullScreen` be an attribute under `MIDP`

midp/midp.js
```diff
@@ -1340,3 +1355,138 @@ addUnimplementedNative("com/sun/j2me/content/InvocationStore.get0.(Lcom/sun/j2me
...
+        MIDP.ScreenWidth = window.innerWidth;
+        shouldSendRotate = true;
+    } else if (window.innerWidth < MIDP.ScreenWidth) {
+        console.warn("Window shrunk from initial width!");
+        MIDP.ScreenWidth = window.innerWidth;
+        shouldSendRotate = true;
+    }
+
+    if (window.innerHeight < MIDP.ScreenHeight) {
+        window.dispatchEvent(new Event("keyboardShown"));
```
yxl: Rotation can cause window.innerHeight < MIDP.ScreenHeight. We cannot assert keyboard is visible in this case.

midp/midp.js
```diff
@@ -1340,3 +1355,138 @@ addUnimplementedNative("com/sun/j2me/content/InvocationStore.get0.(Lcom/sun/j2me
...
+    } else if (window.innerHeight >= MIDP.ScreenHeight) {
+        if (window.innerHeight > MIDP.ScreenHeight) {
+            console.warn("Window grew beyond initial height!");
+            MIDP.ScreenHeight = window.innerHeight;
+            shouldSendRotate = true;
+        }
+        window.dispatchEvent(new Event("keyboardHidden"));
```
yxl: We cannot assert keyboard is invisible either in this case.

midp/midp.js
TimAbraldes: Good idea! Will do

midp/gfx.js
TimAbraldes: Fixed locally

midp/midp.js
TimAbraldes: Fixed locally. Also all various keyboard-related variables

midp/midp.js
TimAbraldes: I don't think we support rotation.

The current heuristic for deciding that the keyboard is visible is not perfect, but it seems to work pretty well so far. A new, better heuristic would be great to have; perhaps we can do this as a follow-up?

If we want to support rotation, we will definitely have to modify this heuristic and we will also have to come up with one for determining whether a rotation has occurred  so we can send rotation information to the MIDlet. That could be another follow-up

midp/midp.js
TimAbraldes: No; our detection of whether the keyboard is visible isn't perfect, but this patch isn't making it worse (it should be the same as before).

mykmelez:
Looks good, works well on device. Let's stay on top of #975 and #976 to make this more robust.

PR #972, opened 2015-01-29
Avoid leaking text editor elements

### LINKED ISSUE #797: We're never removing text editors
We assumed setParent would be called with *null* when a TextEditor had to be removed, but this doesn't appear to be the case.

Instead of attaching/detaching when setParent is called, we could try to do it in setFocus.

yxl:
This fixes #797.  Currently there is no way to release `TextEditor` object except the midlet calling `TextEditor.setParent()` explicitly. Unfortunately, at least two apps that I'm using don't do so. Since every `TextEditor` associates with a html content editable or input element, it leaks lots of html elements when you open and close a midlet page with `TextEditor` repeatedly.

I find that two major apps of mine will only show one `TextEditor` at a time. Therefore this allow me to do a workaround with leaking issue. Instead of creating  a html element for each `TextEditor`, we create only one html element for all `TextEditors` that share the same type html element. Currently we need two types of html element: one is the content editable `<div>` capable of displaying emojis, and the other is the password input.

When a midlet wants to show a `TextEditor`, we show the html element already created and adjust its appearance and content to match the `TextEditor`.

There needs some tricks about setting the focus. Most midlets makes the `TextEditor` visible before setting focus, but a few do the opposite. If the focus is set before the `TextEditor` is visible, we need to  postpone setting focus. Another issue with setting focus is that when calling `setVisible` and `setFocus` successively  without pause, the html element may not display in time to get focus. So we have to do a `setTimeout` to let the html element render properly.

Other small bugs fixed are:
1. Use `<object>` to display emoji instead of `` to not allow resizing emoji when typing in a text editor.
2. Use the same font style for the text in the text editor as that in the canvs to keep the text in the TextEditor and the canvas looks the same.


midp/text_editor.js
```diff
@@ -9,110 +13,138 @@ var TextEditorProvider = (function() {
...
-            return (this.styles && this.styles[styleKey]) || null;
-        },
-
         focus: function() {
             this.focused = true;
-            this.textEditorElem && this.textEditorElem.focus();
+            return new Promise(function(resolve, reject) {
+                if (currentVisibleEditor !== this ||
+                    document.activeElement === this.textEditorElem) {
+                    resolve()
```
mykmelez: Nit: missing semicolon.

midp/gfx.js
```diff
@@ -1600,22 +1582,25 @@ var currentlyFocusedTextEditor;
...
         if (shouldFocus && (currentlyFocusedTextEditor != this.textEditor)) {
-            this.textEditor.focus();
+            promise = this.textEditor.focus();
             currentlyFocusedTextEditor = this.textEditor;
         } else if (!shouldFocus && (currentlyFocusedTextEditor == this.textEditor)) {
-            this.textEditor.blur();
+            promise = this.textEditor.blur();
             currentlyFocusedTextEditor = null;
+        } else {
+            promise = Promise.resolve();
```
mykmelez: Natives that call asyncImpl don't have to do so in every case. The VM is smart enough to do the right thing if an "async native" that is listed in *yieldMap* sometimes returns a synchronous value. So this case should just return.

midp/text_editor.js
```diff
@@ -9,110 +13,138 @@ var TextEditorProvider = (function() {
...
+                if (currentVisibleEditor) {
+                    currentVisibleEditor.visible = false;
+                }
+                currentVisibleEditor = this;
+            } else {
+                currentVisibleEditor = null;
+            }
+
+            if (aVisible) {
+                // Add the "show" class to set opacity to 0.
```
mykmelez: This also sets the z-index (and the *blur* block below also sets the opacity).

midp/text_editor.js
```diff
@@ -9,110 +13,138 @@ var TextEditorProvider = (function() {
...
                 // Sometimes in Java, setVisible() is called after focus(), to make
                 // sure the native input won't lose focus, we change opacity instead
                 // of visibility.
```
mykmelez: Hmm, I'm confused, because this code comment suggests that we change opacity instead of visibility to handle the case where *setVisible* is called after *focus*. But in your description of this pull request, you suggest that we need to delay setting focus in that case:

> If the focus is set before the TextEditor is visible, we need to postpone setting focus. Another issue with setting focus is that when calling setVisible and setFocus successively without pause, the html element may not display in time to get focus. So we have to do a setTimeout to let the html element render properly.

Perhaps this code comment is incorrect or obsolete? If so, then I wonder if we should switch back to changing visibility rather than opacity.

midp/text_editor.js
```diff
@@ -137,9 +169,56 @@ var TextEditorProvider = (function() {
...
+            this.backgroundColor = color;
+            this._setStyle("backgroundColor", util.abgrIntToCSS(color));
+        },
+
+        getBackgroundColor: function() {
+            return this.backgroundColor;
+        },
+
+        setForegroundColor: function(color) {
+            this.foregroundColor =  color;
```
mykmelez: Nit: extra space.

mykmelez:
> I find that two major apps of mine will only show one TextEditor at a time. Therefore this allow me to do a workaround with leaking issue. Instead of creating a html element for each TextEditor, we create only one html element for all TextEditors that share the same type html element. Currently we need two types of html element: one is the content editable &lt;div> capable of displaying emojis, and the other is the password input.

Even if we find a midlet that shows two text editors at once, only one of them can be focused at any given time, so this strategy should still work, if we simulate the display of the other one.

But I'm happy to delay solving that problem until we actually find a midlet that shows two text editors.

midp/text_editor.js
yxl: This code comment is obsolete. We can switch back to using visibility and drop the hacking way of setting opacity and z-index.

yxl:
@mykmelez address all above comments with 624bf96

PR #977, opened 2015-01-30
ignore messages from different origin (fixes #909)

### LINKED ISSUE #909: make parent frame check origin of child frame when receiving message
To avoid the possibility that the child frame will somehow be navigated to a malicious page that abuses the APIs it accesses through the parent frame, the parent frame should check the origin of the child frame to ensure it is as expected when it receives a message from the child frame.

mykmelez:
Navigation in the child shouldn't happen, but this is defense in depth in case it does. Unfortunately, I can't check *event.target.origin*, as suggested in #909, because the child doesn't *postMessage* to the parent. So instead I track the child's location with *mozbrowserlocationchange* and ignore the message if the parent and child origins differ.

This one is difficult to test automatically, but I've confirmed that it works as expected with these manual test steps:

1. load some midlet in a browser tab, f.e. http://localhost:8000/index.html?midletClassName=midlets.InitMidlet&jars=tests/tests.jar
2. install [tcpsocketpup](https://github.com/mykmelez/tcpsocketpup/releases/latest) and use it to give the URL mozbrowser permissions
3. reload the tab
4. open Developer Tools and switch to the child frame: `cd(frames[0])`
5. evaluate `window.alert(JSON.stringify({"command":"open","type":"alert","message":"Hello, world!"}))`: an alert dialog should appear that says "Hello, world!"
6. navigate the child frame to another origin: `document.location = "http://www.mozilla.org/"`
7. evaluate `window.alert(JSON.stringify({"command":"open","type":"alert","message":"Hello, world!"}))` again: no alert dialog should appear

After step 6, you might also see this warning in the console: "on location change: child origin https://www.mozilla.org !== parent origin http://localhost:8000", and after step 7, you might see the error: ""on show modal prompt: child origin https://www.mozilla.org !== parent origin http://localhost:8000". I did see them once, but another time I didn't (which I suspect is an issue with Firefox's Web Console). Nevertheless, under no circumstances should you see the "Hello, world!" alert again.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,15 @@
   });
 })();
 
+var childOrigin;
+document.getElementById("mozbrowser").addEventListener("mozbrowserlocationchange", function(event) {
+  childOrigin = new URL(event.detail).origin;
+  var parentOrigin = document.location.origin;
+  if (childOrigin !== parentOrigin) {
+    console.warn("on location change: child origin " + childOrigin + " !== parent origin " + parentOrigin);
+  }
+});
+
 var DumbPipe = {
   // Functions that handle requests to open a pipe, indexed by type.
   openers: {},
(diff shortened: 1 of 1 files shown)


mykmelez:
You can also combine steps 6 and 7 by evaluating: `document.location='data:text/html,<script>window.alert(JSON.stringify({"command":"open","type":"alert","message":"Hello, world!"}))</script>'`

PR #978, opened 2015-01-30
Convert long to int for setMediaTime.

brendandahl:
I added real type checking for native return values and found we were returning a long and trying to divide a long by 1000. Separate PR coming to add real type checking.

### DIFF
--- a/midp/media.js
+++ b/midp/media.js
@@ -1025,7 +1025,7 @@ Native["com/sun/mmedia/DirectPlayer.nGetMediaTime.(I)I"] = function(handle) {
 
 Native["com/sun/mmedia/DirectPlayer.nSetMediaTime.(IJ)I"] = function(handle, ms) {
     var container = Media.PlayerCache[handle];
-    return container.player.setMediaTime(ms);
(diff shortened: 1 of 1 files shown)


andreasgal:
Isn't toDouble better here? In case its outside the int range

midp/media.js
```diff
@@ -1025,7 +1025,7 @@ Native["com/sun/mmedia/DirectPlayer.nGetMediaTime.(I)I"] = function(handle) {
 
 Native["com/sun/mmedia/DirectPlayer.nSetMediaTime.(IJ)I"] = function(handle, ms) {
     var container = Media.PlayerCache[handle];
-    return container.player.setMediaTime(ms);
+    return container.player.setMediaTime(ms.toInt());
```
yxl: `return container.player.setMediaTime(ms.toNumber()) | 0` is better.

brendandahl:
> Isn't toDouble better here? In case its outside the int range

Probably, but I would hope we wouldn't have any media that is over 24 days(2^31-1ms).

marco-c:
We should add a test for this, could you file a followup?

Issue #739, opened 2014-12-16
NS_ERROR_ILLEGAL_DURING_SHUTDOWN on parentWindow.openDialog in slLauncher.openBrowser


mykmelez:
I'm seeing a lot of intermittent test failures, like https://travis-ci.org/andreasgal/j2me.js/builds/44157722. It looks like the problem is in _TestAudioPlayer.read_:

> Update event: closed null
> java.io.IOException: 
> - java/lang/Throwable.<init>(), bci=8
> - java/lang/Exception.<init>(), bci=5
> - java/io/IOException.<init>(), bci=5
> - java/io/IOException.RaiseExceptionSynthetic(), bci=9
> - javax/microedition/media/TestAudioPlayer.read(), bci=4
> - javax/microedition/media/TestAudioPlayer.test(), bci=104
> - RunTests.runTest(), bci=115
> - RunTests.startApp(), bci=166
> - javax/microedition/midlet/MIDletTunnelImpl.callStartApp(), bci=4
> - com/sun/midp/midlet/MIDletPeer.startApp(), bci=12
> - com/sun/midp/midlet/MIDletStateHandler.startSuite(), bci=252
> - com/sun/midp/main/AbstractMIDletSuiteLoader.startSuite(), bci=54
> - com/sun/midp/main/CldcMIDletSuiteLoader.startSuite(), bci=11
> - com/sun/midp/main/AbstractMIDletSuiteLoader.runMIDletSuite(), bci=146
> - com/sun/midp/main/MIDletSuiteLoader.main(), bci=15

There are also some socket-related exceptions in the logs, like this one:

> java.io.IOException: error opening socket
> - java/lang/Throwable.<init>(), bci=8
> - java/lang/Exception.<init>(), bci=5
> - java/io/IOException.<init>(), bci=5
> - java/io/IOException.RaiseExceptionSynthetic(), bci=9
> - com/sun/midp/io/j2me/socket/Protocol.connect(), bci=123
> - com/sun/midp/io/j2me/socket/Protocol.open(), bci=125
> - com/sun/midp/io/j2me/socket/Protocol.openPrim(), bci=7
> - javax/microedition/io/Connector.openPrim(), bci=325
> - javax/microedition/io/Connector.open(), bci=6
> - javax/microedition/io/Connector.open(), bci=6
> - javax/microedition/io/Connector.open(), bci=5
[...]

PR #979, opened 2015-01-30
Thoroughly check return values.

brendandahl:
Use asserts.

### DIFF
--- a/interpreter.ts
+++ b/interpreter.ts
@@ -82,17 +82,14 @@ module J2ME {
    * Debugging helper to make sure native methods were implemented correctly.
    */
   function checkReturnValue(methodInfo: MethodInfo, returnValue: any) {
-    if (returnValue instanceof Promise) {
-      console.error("You forgot to call asyncImpl():", methodInfo.implKey);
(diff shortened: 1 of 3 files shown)


mykmelez:
Fascinatingly, this branch seems to be able to reproduce #739 consistently:

> [debug] [phantom] opening url: http://localhost:8000/index.html?logConsole=web,page, HTTP GETFAIL [Exception... "<no message>"  nsresult: "0x8046001e (<unknown>)"  location: "JS frame :: resource://slimerjs/slLauncher.jsm :: slLauncher.openBrowser :: line 186"  data: no]

We should use it to figure out the cause of that problem, especially if we can reproduce it locally.

Don't change this branch!

mykmelez:
> Fascinatingly, this branch seems to be able to reproduce #739 consistently:

Erm, or not, apparently, on third run.

mykmelez:
(Nor am I able to reproduce locally.)

mykmelez:
Here's another test run with a merge from master, and it's also successful:

https://travis-ci.org/mykmelez/j2me.js/builds/48973614

So I guess this doesn't trigger that SlimerJS bug consistently after all, at least not anymore.

PR #980, opened 2015-01-30
Fix non fullscreen display

yxl:
https://github.com/andreasgal/j2me.js/pull/966 makes canvas hold its position and not resize. It fixed the bug - black bar appears after opening text editor, but introduces two side effects when we set `autoresize=1`:

1. app, which is not in full screen, cannot display its whole content. The reason is `DisplayDevice.getScreenHeight0()` should return canvas height, but not the actual screen height. The app mistaken the screen height for the canvas height and draws out of the canvas area.

2. The system back button doesn't stay at the bottom but at the top.

The pull request reverts related changes.

As https://github.com/andreasgal/j2me.js/pull/972 already fixed the bug that causes canvas to change position and resize, so it's safe to do the reverting.

### DIFF
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -63,11 +63,11 @@ var currentlyFocusedTextEditor;
     };
 
     Native["com/sun/midp/lcdui/DisplayDevice.getScreenWidth0.(I)I"] = function(id) {
-        return MIDP.ScreenWidth;
+        return MIDP.Context2D.canvas.width;
     };
 
     Native["com/sun/midp/lcdui/DisplayDevice.getScreenHeight0.(I)I"] = function(id) {
-        return MIDP.ScreenHeight;
+        return MIDP.Context2D.canvas.height;
     };
 
     Native["com/sun/midp/lcdui/DisplayDevice.displayStateChanged0.(II)V"] = function(hardwareId, state) {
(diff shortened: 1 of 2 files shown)


TimAbraldes:
I think we could also alleviate these issues by  having `DisplayDevice.getScreenHeight0` return the actual screen height minus the height of the header, and positioning the canvas below the header on the screen (I thought I did this but I didn't actually test on a MIDlet that uses a header - I assumed that the tests would catch any issues).

I'm not familiar with the back button but it sounds like we could easily position it similarly to the canvas.

An alternative might be to position the canvas using some other positioning but somehow guarantee its height and width (I'm not sure if this is possible without fixed positioning)

yxl:
@TimAbraldes We could calculate the `DisplayDevice.getScreenHeight0` by calculation and positioning the back button by calculation as well, but doing these layout things manually make our code hard to understand and maintain. I tend to utilize the external css style file as much as possible other than manipulate css styles with js directly. That's why I want to remove the js code which change the css style properties.

For the back button, you could test using `index.html?jars=tests/tests.jar&midletClassName=gfx.CanvasWithHeaderTest`

This test has passed with `autosize=0`, but failed with `autosize=1`.  Any elegant way to fix it? :)

mykmelez:
Hmm, we should test both with and without autoresize for stuff that might break in the different modes. In the meantime, let's take this to resolve the regression.

PR #981, opened 2015-01-30
miscellaneous new text editor fixes

yxl:
Fix `getContentHeight()` to make it more accurate. 
There is a midlet that increases the TextEditor unlimitedly, but Nokia Asha's implementation doesn't allow that, so I set a limit of `getContentHeight()` to 130.

Also it fixes a mistake that when the type of TextEditor has changed, I forgot to set `oninputCallback` and cause TextEditorThread don't get input notification.

### DIFF
--- a/midp/text_editor.js
+++ b/midp/text_editor.js
@@ -25,6 +25,7 @@ var TextEditorProvider = (function() {
         id: -1,
         selectionRange: [0, 0],
         focused: false,
+        oninputCallback: null,
 
         // opaque white
         backgroundColor:  0xFFFFFFFF | 0,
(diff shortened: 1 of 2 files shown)


marco-c:
Are you sure there's a limit? I'm testing on a Nokia Asha and there doesn't seem to be a limit

yxl:
@marco-c It seems not `getContentHeight()` issue. I observe the html element of a message text field grows out of CanvasItem behind it, but it doesn't on the Nokia Asha 3110 device.

I removed the height limit from this pull request and let's fix that in another bug.

PR #996, opened 2015-01-31
Self hosting strings.

mbebenita:
This PR gets rid of our native String implementation.


marco-c:
I'm not sure UTF_16_Writer and UTF_16_Reader are going to be faster this way, they were modified because they were reading/writing one byte at a time from/in a buffer.
I would also expect ByteArrayInputBuffer to be faster in a native implementation (since it doesn't have to access Java properties), but I'd be happy if this wasn't the case.

mbebenita:
I think we can write better JS by going the Java -> JS route. We can optimize the implementation of UTF_16_Writer / Reader directly in Java.

Can you elaborate on why you think ByteArrayInputBuffer would be faster in a native implementation? Accessing Java properties is fast, and all Java objects should have a fixed shape which will keep the JS engine happy.

marco-c:
> Can you elaborate on why you think ByteArrayInputBuffer would be faster in a native implementation? Accessing Java properties is fast, and all Java objects should have a fixed shape which will keep the JS engine happy.

I guess it depends on the quality of the generated code. Also, I see ByteArrayInputBuffer is going to call a native (the System::arrayCopy function), while in JS we're going to copy the array directly using TypedArray functions. So, I think it depends on several factors and we should carefully measure which one is better (it's also possible that self hosting is better in some cases, e.g. small arrays, and natives are better in other cases, e.g. big arrays).

> I think we can write better JS by going the Java -> JS route. We can optimize the implementation of UTF_16_Writer / Reader directly in Java.

Sure. We should also consider that in JS we can use a TextDecoder, while in Java we have to implement decoding by hand. Here too, one approach might be faster than the other according to the size of the data.

mbebenita:
Did a quick test:

```

var dst = new Uint8Array(2048 * 16);
var src = new Uint8Array(2048 * 16);

for (var i = 0; i < 2048 * 16; i++) {
	dst[i] = i;
	src[i] = i;
}

function foo(len) {
	for (var i = 0; i < 10000; i++) {
		dst.set(src.subarray(0, len), 0)
	}
}

function copy(dst, dstOffset, src, srcOffset, length) {
	var i = 0;
	while (i < length) {
		dst[dstOffset] = src[srcOffset];
		dstOffset = dstOffset + 1 | 0;
		srcOffset = srcOffset + 1 | 0;
		i = i + 1 | 0;
	}
}

function bar(len) {
	for (var i = 0; i < 10000; i++) {
		copy(dst, 0, src, 0, len);
	}	
}

// Warmup

foo(256);
bar(256);

for (var k = 1; k < 14; k++) {
	var s = dateNow();
	for (var i = 0; i < 128; i++) { foo(1 << k); }
	var n = (dateNow() - s).toFixed(2);
	
	var s = dateNow();
	for (var i = 0; i < 128; i++) { bar(1 << k); }
	var j = (dateNow() - s).toFixed(2);

	print("Size: " + (1 << k) + ", native: " + n + ", js: " + j);
}
```

my results are:

```
Size: 2, native: 230.33, js: 4.08
Size: 4, native: 227.22, js: 7.60
Size: 8, native: 234.87, js: 13.56
Size: 16, native: 230.27, js: 27.13
Size: 32, native: 236.14, js: 54.58
Size: 64, native: 233.55, js: 104.89
Size: 128, native: 241.43, js: 210.65
Size: 256, native: 239.72, js: 408.29
Size: 512, native: 237.48, js: 805.71
Size: 1024, native: 263.08, js: 1607.18
Size: 2048, native: 297.47, js: 3333.43
Size: 4096, native: 352.22, js: 6576.60
Size: 8192, native: 473.39, js: 13012.67
```

When copying small buffers, under 128, js pulls out a head, after than native is way faster. It's kind of interesting that native has a sub linear curve.

mbebenita:
Measured the length in `System.arraycopy`, `frequency`: `length`. For the most part, we're usually under the 128 threshold.

```
     | 2387: 4 
     |  559: 10 
     |  527: 20 
     |  507: 40 
     |  504: 80 
     |  502: 160 
     |  386: 1 
     |  253: 0 
     |  168: 7 
     |  110: 3 
     |  102: 9 
     |  100: 2 
     |  85: 5 
     |  71: 8 
     |  70: 15 
     |  62: 11 
     |  54: 16 
     |  45: 6 
     |  44: 21 
     |  39: 12 
     |  33: 14 
     |  32: 13 
     |  30: 22 
     |  27: 19 
     |  24: 17 
     |  21: 28 
     |  17: 18 
     |  16: 29 
     |  15: 24 
     |  15: 23 
     |  11: 25 
     |  9: 34 
     |  9: 30 
     |  8: 26 
     |  8: 27 
     |  7: 42 
     |  7: 33 
     |  6: 32 
     |  6: 47 
     |  6: 41 
     |  6: 44 
     |  6: 31 
     |  6: 53 
     |  5: 52 
     |  5: 51 
     |  5: 116 
     |  5: 46 
     |  5: 48 
     |  4: 256 
     |  4: 45 
     |  4: 58 
     |  4: 151 
     |  4: 61 
     |  4: 149 
     |  4: 118 
     |  4: 39 
     |  4: 64 
     |  4: 71 
     |  4: 74 
     |  4: 38 
     |  3: 62 
     |  3: 102 
     |  3: 37 
     |  3: 222 
     |  3: 54 
     |  3: 285 
     |  3: 50 
     |  3: 43 
     |  2: 128 
     |  2: 132 
     |  2: 89 
     |  2: 60 
     |  2: 57 
     |  2: 100 
     |  2: 87 
     |  2: 113 
     |  2: 92 
     |  2: 63 
     |  2: 66 
     |  1: 90 
     |  1: 67 
     |  1: 94 
     |  1: 95 
     |  1: 96 
     |  1: 97 
     |  1: 68 
     |  1: 101 
     |  1: 70 
     |  1: 103 
     |  1: 105 
     |  1: 107 
     |  1: 56 
     |  1: 114 
     |  1: 115 
     |  1: 49 
     |  1: 75 
     |  1: 121 
     |  1: 123 
     |  1: 127 
     |  1: 76 
     |  1: 129 
     |  1: 77 
     |  1: 134 
     |  1: 135 
     |  1: 136 
     |  1: 140 
     |  1: 79 
     |  1: 35 
     |  1: 153 
     |  1: 157 
     |  1: 81 
     |  1: 174 
     |  1: 83 
     |  1: 224 
     |  1: 84 
     |  1: 85 
     |  1: 400 
     |  1: 638 
     |  1: 755 
     |  1: 86 
     |  1: 59 
     |  1: 36 
     |  1: 55 
     |  1: 88 
```

mbebenita:
@mykmelez @brendandahl How do I figure out why this is not passing all the tests. @brendandahl and I resolved an issue with a test file, where we now pass more than before. 

now:
```
1:1  | com.sun.cldc.i18n.TestUtfReaders: 72 pass, 0 fail, 0 unknown pass 
```

before:
```
1:1  | com.sun.cldc.i18n.TestUtfReaders: 63 pass, 0 fail, 0 unknown pass 
```

I adjusted the total pass count accordingly, however now it says that on my branch I have

```
1:1  | DONE: 71574 pass, 0 fail, 212 known fail, 0 unknown pass, 40032ms 
```

and on master it is

```
DONE: 71565 pass, 0 fail, 215 known fail, 0 unknown pass, 39664ms 
```

Where are the fewer unknown fails coming from?

mykmelez:
> Where are the fewer unknown fails coming from?

With #1040, I see this change in a diff of logs from master and this branch:

```
< 1:1  | com.sun.cldc.i18n.TestUtfReaders: 63 pass, 0 fail, 3 known fail, 0 unknown pass 
---
> 1:1  | com.sun.cldc.i18n.TestUtfReaders: 72 pass, 0 fail, 0 known fail, 0 unknown pass 
```

So it looks like they're from the same test class that is now passing more test assertions. Perhaps you fixed the known failures that we added in #343?

marco-c:
Have you also measured Java vs Java native?

If the JS implementation is faster than the native implementation, we could implement the Java natives with a threshold (use JS if < threshold, use native if > threshold).

Even if in the common case we're using small data, we should also make sure we're not making some things too slow (for example, when a MIDlet is downloading some data it could use a buffer of 4096 bytes, and from your numbers *native: 352.22, js: 6576.60*)

mbebenita:
> Have you also measured Java vs Java native?

I haven't done a complete analysis. Do you have a larger benchmark in mind, where I can test the entire stack not just a few methods.

marco-c:
I see you've merged bdahl's startup benchmark branch into this PR, have you run the measurements? How is startup time affected by this?

marco-c:
> I haven't done a complete analysis. Do you have a larger benchmark in mind, where I can test the entire stack not just a few methods.

Unfortunately not, we'd need to test manually

java/custom/com/sun/cldc/i18n/j2me/UTF_8_Reader.java
```diff
@@ -32,20 +32,62 @@
 
 /** Reader for UTF-8 encoded input streams. */
 public class UTF_8_Reader extends StreamReader {
```
marco-c: If you have restored this file to the original version, I think we can just remove it from java/custom/

java/custom/com/sun/cldc/i18n/j2me/UTF_8_Reader.java
```diff
@@ -56,28 +98,173 @@ public Reader open(InputStream in, String enc)
      * @exception IOException is thrown if the input stream 
      * could not be read for the raw unconverted character
      */
-    public native int readNative(char cbuf[], int off, int len);
     public int read(char cbuf[], int off, int len) throws IOException {
-        if (!initialized) {
```
marco-c: I think we should keep this optimization, otherwise we'd end up reading a few bytes at a time from the InputStream. In the case of a FCInputStream, we'd call the FS native a lot of times.

java/custom/com/sun/cldc/i18n/j2me/UTF_8_Writer.java
```diff
@@ -55,10 +54,61 @@
...
      */
     public void write(char cbuf[], int off, int len) throws IOException {
-        out.write(encodeUTF8(cbuf, off, len));
-    }
+        byte[] outputByte = new byte[4];     // Never more than 4 encoded bytes
+        int inputChar;
+        int outputSize;
+        int count = 0;
 
-    private native byte[] encodeUTF8(char cbuf[], int off, int len);
```
marco-c: There's the same problem as with UTF_8_Reader, we're calling the OutputStream::write method a bunch of times

java/custom/java/io/ByteArrayInputStream.java
```diff
@@ -0,0 +1,270 @@
...
+        if (pos >= count) {
+            return -1;
+        }
+        if (pos + len > count) {
+            len = count - pos;
+        }
+        if (len <= 0) {
+            return 0;
+        }
+        System.arraycopy(buf, pos, b, off, len);
```
marco-c: Can we use the org.mozilla.internal.Sys class here?

java/custom/java/lang/String.java
```diff
@@ -197,7 +200,7 @@ public String(char value[], int offset, int count) {
      * @since      JDK1.1
      */
     public String(byte bytes[], int off, int len, String enc)
-        throws UnsupportedEncodingException
+            throws UnsupportedEncodingException
```
marco-c: Could you avoid indentation changes in these files? It makes it easier to review and, in the future, to see what we've changed w.r.t. the reference implementation

java/custom/org/mozilla/internal/Sys.java
```diff
@@ -0,0 +1,52 @@
...
+
+package org.mozilla.internal;
+
+import java.io.*;
+import com.sun.cldchi.io.*;
+import java.security.*;
+import java.util.PropertyPermission;
+
+/**
+ * The <code>Sys</code> class contains several useful priviledged functions.
```
marco-c: Nit: typo "priviledged"

brendandahl:
I ran some benchmarks where I'm doing 10 rounds, waiting for 5s to start next round, forcing gc between rounds, and deleting both databases.

```
Current times: [1406.599403,2060.8294010000004,1461.4797370000001,1438.9177160000002,1643.4384960000002,1467.970296,1403.8400299999998,2864.864712,1420.435996,1590.409412]
Current mean : 1676ms
Baseline mean: 1495ms
+/-          : 181ms
%            : 12.07
INSIGNIFICANT RESULT
```

I also ran without deleting databases and had similarly insignificant results.
```
Current mean : 1251ms
Baseline mean: 1238ms
+/-          : 12ms
%            : 0.99
INSIGNIFICANT RESULT
```

andreasgal:
Does this change memory consumption overall?

java/custom/java/io/ByteArrayInputStream.java
mbebenita: Yes, will change.

java/custom/java/lang/String.java
mbebenita: My IDE does this strange thing, will revert it back.

mbebenita:
> Does this change memory consumption overall?

Hard to tell, @brendandahl is looking into getting his benchmark harness to report memory usage.

This PR should at least implement the Java's way of handling strings. We should only reflect Java strings into JS strings when we need to. This PR makes it possible to improve startup and memory usage by not converting Java UTF16 strings into JS strings during class loading. We can create Java string objects whose internal buffers are subviews of the classfile's constant pool buffer.

Since all the of Java string handling code is in Java, we need to make sure that our AOT compiler generates optimal code. We'll need to eliminate all the array bounds and store checks for these classes, see #1037.

tests/JITBenchmark.java
```diff
@@ -40,22 +44,125 @@ public static void writeByteArrayOutputStream() {
...
+    for (int i = 0; i < count - 1; i++) {
+      array[i % 1024] = a;
+      array[i % 1024] = b;
+    }
+  }
+
   public static void main(String[] args) {
     long start = 0;
 
     start = System.currentTimeMillis();
```
marco-c: In benchmarks, we should use JVM.monotonicTimeMillis (it uses performance.now() instead of Date.now())

marco-c:
I've added two benchmarks for ByteArrayInputStream, ByteArrayOutputStream, UTF_8_reader and UTF_8_Writer

marco-c:
I'm seeing the following results for the benchmarks bench/ByteArrayInputOutputStream and bench/UTF8Bench:

Before:
writeAndReadSingle: 1576
writeAndReadArray: 769
writeUtf8Data: 401
readUtf8Data: 401

After:
writeAndReadSingle: 92
writeAndReadArray: 854
writeUtf8Data: 2106
readUtf8Data: 2106

mbebenita:
My numbers for after are:

```
writeAndReadSingle: 90 
writeAndReadArray: 734 
writeUtf8Data: 444
readUtf8Data: 136
```

Did you measure your numbers in release builds?

marco-c:
The numbers are not really stable on my machine, now for *after* I'm seeing:
writeAndReadSingle: 96
writeAndReadArray: 986
writeUtf8Data: 1008 
readUtf8Data: 199 

(I'm building with *RELEASE=1 JSR_256=0 JSR_082=0 JSR_179=0 make*)

marco-c:
Do you think we could self-host the ByteArrayInputStream and ByteArrayOutputStream methods and use the Java version if the size of the data is < X and the native if the size is > X ?

For example, Sys.copyByteArray could use System.arraycopy if length > some threshold.

mbebenita:
> Do you think we could self-host the ByteArrayInputStream and ByteArrayOutputStream methods and use the Java version if the size of the data is < X and the native if the size is > X ?

Yes, definitely. I'd like to land this PR first and then iterate on it. Your recent patches to optimize Streams should be applied on top of this PR.

I have several patches in the queue that I'll upload shortly.

tests/JITBenchmark.java
```diff
@@ -30,32 +38,154 @@ public static void createPrimitiveArrays() {
...
+
+  public static void main(String[] args) {
+
+    // Sys.eval("J2ME.emitCheckArrayStore = false;");
+    // Sys.eval("J2ME.emitCheckArrayBounds = false;");
+
+    size = 1024;
+
+    begin();
     start = System.currentTimeMillis();
```
marco-c: We should use JVM.monotonicTimeMillis() that uses performance.now() instead of Date.now()

override.js
```diff
@@ -42,127 +42,12 @@ function asyncImpl(returnKind, promise) {
   $.pause("Async");
 }
 
-Override["java/lang/Math.min.(II)I"] = function(a, b) {
-  return Math.min(a, b);
+Override["com/ibm/oti/connection/file/Connection.decode.(Ljava/lang/String;)Ljava/lang/String;"] = function(string) {
```
marco-c: The two Connection natives have been removed with another PR.

marco-c:
I think the regression with ByteArrayInputStream and ByteArrayOutputStream is fine (because we can easily improve it later), but I'm still concerned about UTF_8_Writer and UTF_8_Reader (since they might be much slower than a native implementation).

Do we need to self host them in this PR? Couldn't we start by self-hosting String, StringBuffer, ByteArrayInputStream and ByteArrayOutputStream and take care of UTF_8_Reader and UTF_8_Writer in another PR?

marco-c:
Can we remove the ByteArrayOutputStream bindings in bindings.ts now?

tests/JITBenchmark.java
mbebenita: Ok, will change.

override.js
mbebenita: Looks like a bad merge.

mbebenita: Do I jut need to remove the `Connection.decode` method?

marco-c: *Connection.decode* and *Connection.encode*

mbebenita:
> and take care of UTF_8_Reader and UTF_8_Writer in another PR?
 
Seems to me that the main reason for the regression is the Java implementation of `read`, `write`. I'd like to understand what makes them slow and iterate on compiler improvements. If they are much worse algorithmically, then we can reapply the old optimizations. I don't know how those worked so I didn't reapply them.

marco-c:
The problem is that the RI implementations are reading/writing one byte at a time from the InputStream/OutputStream, while the custom versions are reading the whole content at the beginning (or writing the whole content at the end).
Reading/writing the whole content is definitely faster in our current implementation of the filesystem APIs.
I think the performance difference might be also higher on devices, where I/O is much worse than desktop.

So, it would be better to split this PR in two to avoid regressions with UTF_8_Reader and UTF_8_Writer, unless it's impossible/difficult to do so.

mbebenita:
My numbers after the last patch are:

```
writeUtf8Data: 341 
readUtf8Data: 43 
```

marco-c:
Nice find! Now the performance difference is way smaller (the UTF_8_Writer benchmark is 50% slower on my machine, but I think we can take this regression and improve it later)

runtime.ts
```diff
@@ -1657,27 +1663,38 @@ module J2ME {
...
+    var jsString = String(value);
     var object = <java.lang.String>newObject(Klasses.java.lang.String);
-    object.str = str;
+    var array = new Uint16Array(jsString.length);
+    var length = jsString.length;
+    for (var i = 0; i < length; i++) {
+      array[i] = jsString.charCodeAt(i);
+    }
+    object.value = array;
+    object.count = length;
```
marco-c: We could set the cached properties here as well

PR #1001, opened 2015-01-31
Reduce save/restore calls

marco-c:


marco-c:
I'm seeing some rendering artifacts.

PR #1013, opened 2015-02-02
make RecordStoreUtil.exists, RecordStoreFile.commitWrite, and RecordStoreFile.closeFile async again

mykmelez:
A midlet I'm testing currently fails on firstrun due to what appears to be a race condition between two threads that both access a record store. Making the RecordStoreUtil.exists, RecordStoreFile.commitWrite, and RecordStoreFile.closeFile natives async again (as they used to be, before we made most fs APIs synchronous) seems to work around the problem.


marco-c:
From http://docs.oracle.com/javame/config/cldc/ref-impl/midp2.0/jsr118/javax/microedition/rms/RecordStore.html:
> No locking operations are provided in this API. Record store implementations ensure that all individual record store operations are atomic, synchronous, and serialized, so no corruption will occur with multiple accesses. However, if a MIDlet uses multiple threads to access a record store, it is the MIDlet's responsibility to coordinate this access or unintended consequences may result. Similarly, if a platform performs transparent synchronization of a record store, it is the platform's responsibility to enforce exclusive access to the record store between the MIDlet and synchronization engine.

Let's also open a followup to fix this problem properly.

PR #1015, opened 2015-02-02
Fix text editor secrity issue

yxl:
Fix two bugs:
1. The styled text, such as <span>ABC</span>, from copy paste will be converted to plain text before insertion.
2. The text content inserted by `TextEditor.setContent()`, `TextEditor.insert()` and so on will be escaped, such as `"<span></span>"` is escaped to `"&lt;span&gt;&lt;/span&gt;"`.


mykmelez:
Nice catch.

PR #1016, opened 2015-02-02
Fix InterIsolateMutex bug

marco-c:


marco-c:
This fixes the problem for the first run case, but prevents subsequent runs from working...

midp/midp.js
```diff
@@ -861,40 +861,36 @@ Native["com/sun/midp/util/isolate/InterIsolateMutex.getID0.(Ljava/lang/String;)I
 };
 
 Native["com/sun/midp/util/isolate/InterIsolateMutex.lock0.(I)V"] = function(id) {
-    var ctx = $.ctx;
```
mykmelez: This could benefit from an explanation of why we aren't using *asyncImpl* like the other async natives.

PR #1024, opened 2015-02-03
Use Int8Array instead of Uint8Array throughout the codebase

marco-c:
This PR isn't modifying libs/fs.js because using a Int8Array instead of a Uint8Array causes a test failure (TestInputOutputStorage)

### DIFF
--- a/midp/codec.js
+++ b/midp/codec.js
@@ -66,7 +66,7 @@ DataEncoder.prototype.getData = function() {
 }
 
 var DataDecoder = function(data, offset, length) {
-  this.data = JSON.parse(util.decodeUtf8(new Uint8Array(data.buffer, offset, length)));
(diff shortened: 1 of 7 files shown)


mbebenita:
In the VM, `Uint8Array` are boolean arrays, while `Int8Array` are byte arrays. If you leak any reference to these arrays into the VM, you'll break things.

marco-c:
In this PR, I'm using Int8Array instead of Uint8Array where a Int8Array is expected ([B).

mykmelez:
> This PR isn't modifying libs/fs.js because using a Int8Array instead of a Uint8Array causes a test failure (TestInputOutputStorage)

I have a fix for the test failure and will submit a followup that makes this change in the vfs APIs too.

PR #1027, opened 2015-02-03
Add basic startup benchmark.

brendandahl:
There is currently only a startup benchmark that measures from the load time of benchmarks.js file to the first gfx.refresh0. This could easily be expanded to more benchmarks, but I tried to not build too much until we have use cases for more benchmarks.

### Overview
A multi-round benchmark with a t-test to determine if the result was significantly different than a baseline benchmark.

#### Build the baseline benchmark:
1. `git checkout master`
1. `RELEASE=1 BENCHMARK=1 make`
1. Open up midlet
  1. `cd frames[0]`
  1. `Benchmark.startup.run()`
1. Wait for the above to run and copy/save the array from web console next to `Current times:`

#### Comparing to the baseline
1. checkout your branch and remake `RELEASE=1 BENCHMARK=1 make`
1. replace `baseline` in benchmark.js with the array copied from above
1. Open up midlet again
  1. `cd frames[0]`
  1. `Benchmark.startup.run()`
1. When completed the benchmark will dump output to the console and you'll hopefully see `FASTER`, if you see `INSIGNIFICANT RESULT` that means the t-test had a p-value >= .05 so it is unlikely your changes made a difference. See the *misc* section for more info on adjusting the benchmarks.

#### Misc
You can adjust the number of rounds in benchmark.js(`numRounds`), it currently defaults to 5. You can also adjust the time until the next round is run (`roundDelay`). I'm still experimenting with this but you'll probably get more stable results if you change the delay to longer, but your benchmark will take longer to run.


mbebenita:
:+1: 

Can we get a button that runs `Benchmark.startup.run()`?

marco-c:
Instead of measuring up to the first refresh0 call, you could measure up to the call to gainedForeground0 (that is also what the phoneME Constants.MEASURE_STARTUP does).

brendandahl:
> Instead of measuring up to the first refresh0 call, you could measure up to the call to gainedForeground0 (that is also what the phoneME Constants.MEASURE_STARTUP does).

Refresh0 always happens after gainedForeground0 and the app is closer to a truly ready state.  Is there an advantage over refresh0?

Later, we also could add several startup measurements.

marco-c:
I was suggesting gainedForeground0 because it's where we switch from "Loading MIDlet..." to a MIDlet specific screen.

mykmelez:
This needs resolution of a trivial merge conflict and some docs in the README.

Otherwise, it works as expected, except for a few issues:

1. The Run Startup Benchmark button remains disabled after the baseline benchmark finishes building, and you have to reload the page to enable that button. This isn't a big deal, as the workaround is simple, and you'll probably reload the page anyway after making the change you want to benchmark. It's just unexpected, once the baseline is available. And unnecessary, since the button reloads the page; so it shouldn't be necessary to reload it yourself to press the button.
2. With benchmarking enabled, loading a midlet reports "startTimer called while benchmark not running" and "stopTimer called while benchmark not running", whether or not you're benchmarking.
3. Deleting the filesystem in conjunction with the *downloadJAD* configuration parameter is not going to give useful results, since it adds network requests of indeterminate length to every startup (to get the JAD and JAR files for the midlet). And *downloadJAD* is the new standard way to configure an instance of the app to run a midlet, since we aren't going to ship the app with a pre-packaged midlet. So I would disable *deleteFs* by default.
4. *deleteJitCache* is probably why you needed to bump the timeout from 500ms to 5000ms. The cache currently writes each method in a new transaction, which is expensive. #1076 will fix this, and afterward the benchmarking code should be made to flush the cache (and the filesystem) before reloading instead of waiting for a certain amount of time. We can do this in a followup, however.
5. I too would rather see the time to *gainedForeground0*, which is a one-time event that represents the moment at which the app first displays the midlet's interface.

brendandahl:
1. I can change this, but I left it this way more as a precaution against cached files and so baseline and the benchmark first round runs are closer i.e. the code is loaded once before the benchmark data is collected. I'm unsure if just doing window.location.reload will reload cached js files. Fine with changing it though.

2. Fixed

3. I've just been running with the jar manually specified in the url, the network request(local) doesn't seem to give much variation in the test runs. We could add another option to better support the downloadJad option.

4. I added the delete cache so a fresh run can be simulated for every round.

5. Done

mykmelez:
> I can change this, but I left it this way more as a precaution against cached files and so baseline and the benchmark first round runs are closer i.e. the code is loaded once before the benchmark data is collected. I'm unsure if just doing window.location.reload will reload cached js files. Fine with changing it though.

*location.reload* should reload cached files. If it doesn't, though, or if you want to make sure they're reloaded, then the solution is the same whether or not you enable the button immediately: force-reload the page before pressing the button.

> I've just been running with the jar manually specified in the url, the network request(local) doesn't seem to give much variation in the test runs. We could add another option to better support the downloadJad option.
> 
> I added the delete cache so a fresh run can be simulated for every round.

The more interesting benchmark is subsequent run, as that's the one users will experience the most and have greater expectations about (since we won't be explaining away slowness by referencing the need to download/optimize the midlet like on firstrun).

So unless you're benchmarking something that only happens on firstrun, it's more valuable to benchmark without deleting the filesystem and compiled method cache between runs (provided you ensure that both datastores are synced to persistent storage before restarting the app), discarding the initial run if needed to avoid skewing the results.

But we can do that in a followup.

README.md
```diff
@@ -143,6 +143,26 @@ Modelines for JS files:
...
+The startup benchmark measures from when the benchmark.js file loads to the call of `DisplayDevice.gainedForeground0`. Included in a benchmark build are helpers to build baseline scores so that subsequent runs of the benchmark can be compared. A t-test is used in the comparison to see if the changes were significant.
+
+To use:
+
+*It is recommended that a dedicated Firefox profile is used with the about:config preference of `security.turn_off_all_security_so_that_viruses_can_take_over_this_computer` set to true so garbage collection and cycle collection can be run in between test rounds*
+
+1. Checkout the version you want to be the baseline(usually mozilla/master).
+1. Build a benchmark build `RELEASE=1 BENCHMARK=1 make`
+1. Open the midlet you want to test and click `Build Benchmark Baseline`
+1. When finished the message `FINISHED BUILDING BASELINE` will show up in the console.
```
mykmelez: This message doesn't actually show up in the console if you set RELEASE=1.

mykmelez: So, you could tell readers to add "&logLevel=log" to the URL parameters (or set the log level in some other way). But even better would be to remove "RELEASE=1" from this set of instructions, since that flag is only peripherally related to benchmarking.

README.md
```diff
@@ -143,6 +143,26 @@ Modelines for JS files:
...
 
+## Benchmarks
+
+### Startup Benchmark
+
+The startup benchmark measures from when the benchmark.js file loads to the call of `DisplayDevice.gainedForeground0`. Included in a benchmark build are helpers to build baseline scores so that subsequent runs of the benchmark can be compared. A t-test is used in the comparison to see if the changes were significant.
+
+To use:
+
+*It is recommended that a dedicated Firefox profile is used with the about:config preference of `security.turn_off_all_security_so_that_viruses_can_take_over_this_computer` set to true so garbage collection and cycle collection can be run in between test rounds*
```
mykmelez: This is fine for now, although in the long run we should make tcpsocketpup give the page this capability.

marco-c:
We should also keep in mind that the Student's t-test may yield incorrect results if its assumptions aren't met.
We should probably add statistical tests for the assumptions (normally distributed population and equal variances). The normal distribution assumption would be more easily satisfied with a larger number of rounds (> 30).
(I've also read Welch's t-test would be better than Student's t-test if variances are unequal or when it is unknown whether variances are equal or not: http://m.beheco.oxfordjournals.org/content/17/4/688.full.pdf)

PR #1028, opened 2015-02-03
Optimize monitorEnter/monitorExit

marco-c:


mbebenita:
Looks great, on my micro benchmarks this improves perf nearly 8x.

brendandahl:
Should we be adding a this._lock = null to all js constructors?

mbebenita:
I was thinking about that, let's add it if it doesn't cause a regression.

PR #1033, opened 2015-02-04
Configure Makefile to download closure compiler

yxl:


mykmelez:
*wget* warns:

> wget -O tools/closure.jar -N https://github.com/mykmelez/closure-compiler/releases/download/v0.1/closure.jar
> WARNING: timestamping does nothing in combination with -O. See the manual
> for details.

The man page says:

> Use of -O is not intended to mean simply "use the name file instead of the one
> in the URL;" rather, it is analogous to shell redirection: wget -O file
> http://foo is intended to work like wget -O - http://foo > file; file will be
> truncated immediately, and all downloaded content will be written there.
> 
> For this reason, -N (for timestamp-checking) is not supported in combination
> with -O: since file is always newly created, it will always have a very new
> timestamp. A warning will be issued if this combination is used.

Instead of `-O tools/closure.jar` you could use `-P tools/`. But that won't necessarily have the intended effect either, since the *tools/closure.jar* target will only be made if it doesn't already exist.

If you really want the file to be updated whenever the remote file has a different timestamp from the local file (as opposed to just when the local file doesn't exist), then you'd have to invoke `wget -P tools/ -N https://github.com/mykmelez/closure-compiler/releases/download/v0.1/closure.jar` unconditionally for the *closure* target.

But that still won't handle updates completely automatically, since the URL is version-specific, so the remote file will never have a different timestamp from the local file until we update the URL manually. Ideally, we'd use a URL that points to the build of the latest release on GitHub, but there's no obvious URL like that, although [this StackOverflow question](http://stackoverflow.com/questions/24987542/is-there-a-link-to-github-for-downloading-a-file-in-the-latest-release-of-a-repo) describes workarounds.

For this, however, I think we're fine keeping the *tools/closure.jar* target, and using either *-O* or *-P* to put the file into the right place if it doesn't exist, but not trying to check its timestamp.

yxl:
@mykmelez Thanks for the detailed explanation. I update the patch to use -O without -N.

PR #1038, opened 2015-02-04
Improve experience of exiting 

yxl:
Stop all the running thread and show the following waiting screen before the app window is closed:
exit

### DIFF
--- a/context.ts
+++ b/context.ts
@@ -577,6 +577,10 @@ module J2ME {
               break;
             case VMState.Pausing:
               break;
+            case VMState.Stopping:
+              this.clearCurrentContext();
+              this.kill();
+              return;
           }
           U = VMState.Running;
           this.clearCurrentContext();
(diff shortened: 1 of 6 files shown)


context.ts
```diff
@@ -577,6 +577,10 @@ module J2ME {
               break;
             case VMState.Pausing:
               break;
+            case VMState.Stopping:
+            this.clearCurrentContext();
```
mbebenita: Nit, indentation here.

yxl: Fixed in e62a7c :-)

mykmelez:
With this branch, while testing a midlet suite that has both a foreground and a background midlet, I still see a "flash" of a different screen that has a drawer.

It's quite hard to capture this screen, since it flashes by so quickly, but I think I caught it in the screenshot below.

(And while I was capturing that screenshot, I also saw "PIM to-do list" appear briefly, even though it isn't in the screenshot. So it isn't clear that the problem is completely solved.)

yxl:
@mykmelez The "flash of drawer screen" issue is fixed by 3b05c01.
`DisplayDevice.setFullScreen0` wasn't implemented correctly. When a background display requires full screen mode, it should not change current display mode, but the old implementation failed to check if the display is in background. Before the foreground exits, a background midlet mistakenly changes the screen mode and shows an empty normal mode screen - the drawer screen, which will be soon be covered by the exit screen. So you saw a flash.

PR #1041, opened 2015-02-04
Use the same intern pool for compiler and interpreter.

brendandahl:
string.intern() and J2ME.newStringConstant were not using the same pool of strings which could lead to unexpected comparison results if using "==".
Unfortunately, we don't have a good test for this at the moment.


marco-c:
Could you add such a test? Or file a followup?

brendandahl:
I'll file a follow up, writing a test would require us to have some way to guarantee a method was compiled.

PR #1043, opened 2015-02-04
Remove appmanager classes

marco-c:


marco-c:
This reduces classes.jar size of ~100 kB

mykmelez:
Outside of the app manager code itself, the only references I see are in these files:

* java/midp/com/sun/midp/i18n/ResourceConstants.java
* java/midp/com/sun/midp/installer/InternalMIDletSuiteImpl.java
* java/midp/com/sun/midp/io/j2me/push/ConnectionRegistry.java
* java/midp/com/sun/midp/main/MIDletSuiteUtils.java
* java/midp/com/sun/midp/security/SecurityInitializer.jpp

Nothing looks particularly necessary, and it doesn't look like we hit any of the code paths that rely on these classes, so this should work.

PR #1044, opened 2015-02-04
WIP: Simplify the startup path (by removing some unneeded code and classes)

marco-c:


marco-c:
I've decided to this anyway. Even if it doesn't improve startup perf so much, it definitely improves memory usage.
I'm going to open another PR.

mykmelez:
> I've decided to this anyway. Even if it doesn't improve startup perf so much, it definitely improves memory usage.

Sounds good to me. A memory-only win is justification enough!

PR #1047, opened 2015-02-05
Track expected pass/fail per class file.

### LINKED ISSUE #684: Count the test results individually for each unit test
Currently we count the whole test results for all unit tests to ensure all test cases run as expected. We need to update the total passed and failed test case number every time we add or modify a unit test.
It is easy to conflict when two guys modify different unit tests.
If we could let each unit test count its own test result, it won't cause such conflict.

brendandahl:
Fixes #684


tests/automation.js
```diff
@@ -123,23 +116,18 @@ casper.test.begin("unit tests", 16 + gfxTests.length, function(test) {
...
-                    if (match[i+1] != expectedUnitTestResults[i].number) {
-                        msg += "\n\tExpected " + expectedUnitTestResults[i].number + " " + expectedUnitTestResults[i].name + ". Got " + match[i+1];
-                    }
-                }
-                if (!msg) {
+                var failed = match[2];
+                if (failed === "0") {
                     test.pass('main unit tests');
                 } else {
                     this.echo("data:image/png;base64," + this.captureBase64('png'));
```
yxl: @mykmelez  It seems `this.echo("data:image/png;base64," + this.captureBase64('png'));` is redundant? Everytime the unit tests fails, it logs a bunch of annoying base64 string on travis.

mykmelez: > @mykmelez It seems this.echo("data:image/png;base64," + this.captureBase64('png')); is redundant? Everytime the unit tests fails, it logs a bunch of annoying base64 string on travis.

Yes, those URLs were never very useful for the unit tests (unlike the gfx tests, where they're helpful), and they're even less useful now. We should stop logging them.

PR #1050, opened 2015-02-05
Support acceleration sensor

yxl:
Copy java files of JSR256 from the reference implementation and implement native methods to allow using acceleration sensor.

The unit test of 7c695cf shows how to use the sensor API to get accelerometer data.

c06a2014e7be adds a simple simulator to generate acceleration by mouse movement.

I also fix some java compiler warns that java 1.3 doesn't allow using `enum` as variable name.

Note: Please do `make clean` under `java` directory to regenerate java files from .jpp files if you see errors of no permission.


marco-c:
Do you think we could enable compiling and including these classes in classes.jar behind a flag? This way we wouldn't ship them for MIDlets that don't use them (and classes.jar would be smaller).

yxl:
>Do you think we could enable compiling and including these classes in classes.jar behind a flag? This way we wouldn't ship them for MIDlets that don't use them (and classes.jar would be smaller).

I guess I can.

yxl:
@marco-c I add a Makefile configuration to remove sensor related code by `JSR_256=0 make`.

See the following file about the changes:
` Makefile `
`main.html.in`
`tools/package.sh`

java/custom/com/nokia/mid/ui/lcdui/LCDUIUtils.java
```diff
@@ -3,6 +3,9 @@
 import javax.microedition.lcdui.Display;
 import com.nokia.mid.ui.lcdui.DisplayStateListener;
 
+import javax.microedition.lcdui.Display;
```
marco-c: Nit: this class is already imported a few lines above

midp/sensor.js
```diff
@@ -0,0 +1,300 @@
...
+            console.error("Invalid sensor number: " + sensorsNumber);
+        } else {
+            console.error("Invalid channel number: " + channelNumber);
+        }
+        return J2ME.newArray(J2ME.PrimitiveArrayClassInfo.B.klass, 0);
+    }
+
+    asyncImpl("[B", new Promise(function(resolve, reject) {
+        var result = AccelerometerSensor.readBuffer(channelNumber);
+        setTimeout(resolve.bind(null, result), 50);
```
marco-c: Maybe we should wait for a devicemotion event to occur before resolving this? (and, on desktop, use the timeout)

yxl: The devicemotion events are fired too frequent on the Firefox device and block the main thread on my app. The timeout is deliberately set to make `doMeasureData` slower :-)

java/custom/com/nokia/mid/ui/lcdui/LCDUIUtils.java
yxl: >Nit: this class is already imported a few lines above
Fixed by amending c47f0b1e1c4

Issue #1049, opened 2015-02-05
Race condition issue with clinit


yxl:
There is a race condition issue with <clinit> method if it does any asnyc operations, such as calling `Thread.sleep()`, using `synchronized` keyword.

The first time we visit a class of such async <clinit>, the VM will wait untill the async opeartions are done.
But if another thread visits the same class simultaneously, it won't wait any more.

The following is a test case to illustrate issue. 

```Java
package java.lang;

import gnu.testlet.Testlet;
import gnu.testlet.TestHarness;

class Monkey {
    public static final String name;

    static {
        try {
            Thread.sleep(100);
        } catch (InterruptedException e) {
        }
        name = "Monkey";
    }
}

public class TestStaticInitRaceCondition extends Thread implements Testlet {
    private TestHarness th;

    public void test(TestHarness th) {
        this.th = th;
        start();
        run();
    }

    public void run() {
        System.out.println(Monkey.name);
        th.check("Monkey".equals(Monkey.name));
    }
}
```
I make the &lt;clinit> of `Monkey` async by adding a `Thread.sleep()`. The main thread and a background thread acesses the static propertiy of `Monkey.name` at the same time. The main thread will waits until the &lt;clinit> finishes to get the correct monkey name, but the background thread won't wait and get a null name.

yxl:
cc @brendandahl @mbebenita It may related to the jit compiler.

brendandahl:
I haven't checked if it's a regression yet or if this has always been bad. With fully compiled code you'll actually end up with Monkey.name = null for the first access on both threads.

brendandahl:
If remember correctly we decided to not handle unwinding in static initializers because it didn't happen in any real midlets. We should re enable this check https://github.com/mozilla/j2me.js/blob/master/runtime.ts#L878 though so we can see the error.

yxl:
@brendandahl Thanks. This issue breaks a demo midlet (Cottage360) when I implement https://github.com/mozilla/j2me.js/pull/1050. That'll be good if we can fix it.

PR #1051, opened 2015-02-05
Enable check for unwinding during static initializers.

brendandahl:
Illuminates the problem in #1049 but does not fix it for now.


runtime.ts
```diff
@@ -875,15 +875,8 @@ module J2ME {
         });
         initWriter && initWriter.writeLn("Running Static Constructor: " + classInfo.className);
         $.ctx.pushClassInitFrame(classInfo);
-        // release || assert(!U);
-
-        //// TODO: monitorEnter
```
brendandahl: This is old irrelevant code from when we didn't use pushClassInitFrame.

PR #1052, opened 2015-02-05
Compiled code preemption.

brendandahl:


mbebenita:
Looks good.

mbebenita:
@brendandahl now that I think of it, we may want to inline the fast path for check preemption, which is:

```
preemptionSamples ++;
if ((preemptionSamples & preemptionSampleMask) === 0) {
  preempt();
}
```
 We could just bake in the `preemptionSampleMask` value.

@brendandahl we also need to file a follow up bug to experiment with preemption heuristics.

mbebenita:
Looks good.

PR #1053, opened 2015-02-06
Replace file connection

mykmelez:
@marco-c This still needs cleanup, but it passes tests (except for the FC TCK tests, which I've disabled in preparation for removing them as well; but I think those failures are due to a detail of the old implementation anyway, not actual bustage).

To Do:
* refactor duplicate code and remove debug logging in midp/fs.js
* remove obsolete methods from aot-methods.txt; update yieldmap
* add test for fs.truncate with *size* argument


java/midp/com/sun/cdc/io/j2me/file/DefaultFileHandler.java
```diff
@@ -0,0 +1,673 @@
+/*
```
marco-c: I guess this file is automatically generated from DefaultFileHandler.jpp, so shouldn't be added to the repo.

marco-c:
We should also remove the natives for the IBM implementation from midp/fs.js

marco-c:
> except for the FC TCK tests, which I've disabled in preparation for removing them as well; but I think those failures are due to a detail of the old implementation anyway, not actual bustage

What tests are failing?
Are we sure we need to remove those tests?

mykmelez:
I reenabled the FC TCK tests and got them passing to ensure we don't regress functionality. Next step is to clean up the code.

mykmelez:
> refactor duplicate code and remove debug logging in midp/fs.js
> remove obsolete methods from aot-methods.txt; update yieldmap
> add test for fs.truncate with size argument

Done.

> I guess this file is automatically generated from DefaultFileHandler.jpp, so shouldn't be added to the repo.

Yup, removed.

> We should also remove the natives for the IBM implementation from midp/fs.js

Indeed, removed.

> What tests are failing?

Tests were failing security checks for the permission to read/write files. In the old implementation, we commented out similar checks. I've commented out the checks here too.

After that, there were some other test failures, which I resolved with various fixes.

> Are we sure we need to remove those tests?

Unsure if we *need* to remove them, but we *should* remove them, although they are helpful.

mykmelez:
I think this is ready, although I may want to store the native file descriptor on the *DefaultFileHandler* instance instead of in a *Map*.

mykmelez:
Although it passes tests, I see slowness when writing followed by a hang during firstrun startup in a midlet on device.

mykmelez:
> Although it passes tests, I see slowness when writing followed by a hang during firstrun startup in a midlet on device.

Erm, the hang was apparently an unrelated issue with that specific device, as I was able to reproduce it on the master branch, but I couldn't reproduce it on another device, nor could I reproduce it on desktop, nor could @brendandahl and @marco-c reproduce it.

Unsure about the slowness when writing. I don't have exact numbers, but firstrun startup seems roughly equivalent. And we should be able to AOT-compile some of the methods to speed it up, like we did with a number of methods in the old implementation.

@brendandahl what's the best way to determine which methods to compile?

brendandahl:
I'm not sure we've come up with a great heauristic. I usually just use the runtimeCounter  and either raw call counts, or keep track of what methods get linked and compiled at runtime (https://github.com/mozilla/j2me.js/blob/master/interpreter.ts#L252).

midp/fs.js
```diff
@@ -267,375 +263,463 @@ Native["com/sun/midp/rms/RecordStoreRegistry.stopAllRecordStoreListeners.(I)V"]
...
+    var path = "/" + util.fromJavaString(fileName);
 
-    for (var i = 0; i < path.length; i++) {
-        if (path[i] <= 31 || invalid.indexOf(path[i]) != -1) {
-            return 0;
+    var ctx = $.ctx;
+    asyncImpl("I", new Promise(function(resolve, reject) {
+        function open() {
+            fs.open(path, function(fd) {
+                ctx.setAsCurrentContext();
```
marco-c: No need to setAsCurrentContext here, you could either setAsCurrentContext in the reject branch, or simply use ctx.runtime.newIOException(

midp/fs.js
```diff
@@ -267,375 +263,463 @@ Native["com/sun/midp/rms/RecordStoreRegistry.stopAllRecordStoreListeners.(I)V"]
...
-Native["com/ibm/oti/connection/file/Connection.existsImpl.([B)Z"] = function(path) {
-    return fs.exists(getAbsolutePath(path)) ? 1 : 0;
-};
+    if (readBytes.byteLength <= 0) {
+        return -1;
+    }
 
-Native["com/ibm/oti/connection/file/Connection.fileSizeImpl.([B)J"] = function(path) {
-    return Long.fromNumber(fs.size(getAbsolutePath(path)));
+    var subBuffer = buffer.subarray(offset, offset + readBytes.byteLength);
```
marco-c: Maybe you could use buffer.set here

midp/fs.js
```diff
@@ -267,375 +263,463 @@ Native["com/sun/midp/rms/RecordStoreRegistry.stopAllRecordStoreListeners.(I)V"]
...
+Native["com/sun/cdc/io/j2me/file/DefaultFileHandler.closeForRead.()V"] = function() {
+    var pathname = util.fromJavaString(this.$nativePath);
+    DEBUG_FS && console.log("DefaultFileHandler.closeForRead: " + pathname);
+    var fd = MIDP.openFileHandlers.get(this);
+    fs.close(fd);
+    MIDP.openFileHandlers.delete(this);
+};
 
-    var data = fs.read(fd, curpos, curpos+1);
+Native["com/sun/cdc/io/j2me/file/DefaultFileHandler.openForWrite.()V"] = function() {
```
marco-c: This is similar (or equal?) to openForRead, could you refactor the code in a single function?

midp/fs.js
```diff
@@ -267,375 +263,463 @@ Native["com/sun/midp/rms/RecordStoreRegistry.stopAllRecordStoreListeners.(I)V"]
...
+            // it represents, so a file can be opened by multiple handlers.
+            MIDP.openFileHandlers.set(fileHandler, fd);
+            resolve();
+        });
     }));
 };
 
-Native["com/ibm/oti/connection/file/FCOutputStream.openOffsetImpl.([BJ)I"] = function(jPath, offset) {
-    var path = getAbsolutePath(jPath);
+Native["com/sun/cdc/io/j2me/file/DefaultFileHandler.closeForWrite.()V"] = function() {
```
marco-c: Same here, it's similar (or equal?) to closeForRead

midp/fs.js
```diff
@@ -267,375 +263,463 @@ Native["com/sun/midp/rms/RecordStoreRegistry.stopAllRecordStoreListeners.(I)V"]
...
 };
 
-Native["com/ibm/oti/connection/file/FCOutputStream.writeImpl.([BIII)V"] =
-function(byteArray, offset, count, fd) {
-    fs.write(fd, byteArray.subarray(offset, offset+count));
+Native["com/sun/cdc/io/j2me/file/DefaultFileHandler.flush.()V"] = function() {
+    var pathname = util.fromJavaString(this.$nativePath);
+    DEBUG_FS && console.log("DefaultFileHandler.flush: " + pathname);
+    var fd = MIDP.openFileHandlers.get(this);
+    if (fd) {
```
marco-c: Why is this check needed here and not in the other natives?

midp/fs.js
```diff
@@ -267,375 +263,463 @@ Native["com/sun/midp/rms/RecordStoreRegistry.stopAllRecordStoreListeners.(I)V"]
...
 };
 
-Native["com/ibm/oti/connection/file/Connection.deleteDirImpl.([B)Z"] =
-  Native["com/ibm/oti/connection/file/Connection.deleteFileImpl.([B)Z"];
+addUnimplementedNative("com/sun/cdc/io/j2me/file/DefaultFileHandler.directorySize.(Z)J", Long.fromNumber(0));
 
-Native["com/ibm/oti/connection/file/Connection.isReadOnlyImpl.([B)Z"] = function(path) {
-    console.warn("Connection.isReadOnlyImpl.([B)Z not implemented (" + getAbsolutePath(path) + ")");
-    return 0;
+Native["com/sun/cdc/io/j2me/file/DefaultFileHandler.canRead.()Z"] = function() {
```
marco-c: We can probably avoid checking if the file exists here, looks like the Java implementation always check if the file exists before calling canRead (the same applies to canWrite)

midp/fs.js
mykmelez: > No need to setAsCurrentContext here, you could either setAsCurrentContext in the reject branch, or simply use ctx.runtime.newIOException(

Erm, I haven't actually touched this code. Git's default diff algorithm is just making it look like I did. I moved the new natives to the location of the old natives to make it more obvious which code I'm replacing.

midp/fs.js
mykmelez: > Maybe you could use buffer.set here

Ditto here. This is existing code that I haven't actually changed.

midp/fs.js
mykmelez: > This is similar (or equal?) to openForRead, could you refactor the code in a single function?

Yes, done in 9a835d43fd5df6311be3782bb3652613cefd27c9.

midp/fs.js
mykmelez: > Same here, it's similar (or equal?) to closeForRead

This too was done in 9a835d43fd5df6311be3782bb3652613cefd27c9.

midp/fs.js
mykmelez: > Why is this check needed here and not in the other natives?

Hrm, it shouldn't be needed, so I've removed it in 1f7b2e8.

midp/fs.js
mykmelez: > We can probably avoid checking if the file exists here, looks like the Java implementation always checks if the file exists before calling canRead (the same applies to canWrite)

*Protocol.openInputStream* checks *DefaultFileHandler.exists* before it calls *DefaultFileHandler.canRead*, but *Protocol.canRead* doesn't do that, and that method is public (an implementation of *FileConnection.canRead*), so it could be called by midlets.

mykmelez:
The test failure is bizarre. The first unit test run succeeds, then the second fails with:

> java.lang.TestThreadPriority-0: fail 
> java.lang.TestThreadPriority-1: got (9 5 7 ), expected (9 7 ) 
> java.lang.TestThreadPriority-1: fail 
> java.lang.TestThreadPriority-2: got (9 5 7 1 ), expected (9 7 5 1 ) 
> java.lang.TestThreadPriority: 0 pass, 2 fail, 0 known fail, 0 unknown pass 
> java.lang.TestThreadPriority: test expected 2 passes, got 0 
> java.lang.TestThreadPriority: test expected 0 failures, got 2

marco-c:
Are the tests now passing consistently after the revert of the preemption branch?

mykmelez:
> Are the tests now passing consistently after the revert of the preemption branch?

Yes, they pass consistently for me both locally and on Travis since I merged the "reversion" (actually disablement). Here are the last three builds of this branch on Travis:

* https://travis-ci.org/mozilla/j2me.js/builds/50164204
* https://travis-ci.org/mozilla/j2me.js/builds/50235038
* https://travis-ci.org/mozilla/j2me.js/builds/50235169

tests/native.js
```diff
@@ -150,3 +150,7 @@ Native["javax/microedition/media/TestAudioRecorder.convert3gpToAmr.([B)[B"] = fu
   result.set(converted);
   return result;
 };
+
+// Many tests create FileConnection objects to files with the "/" root,
+// so add it to the list of valid roots.
+MIDP.fsRoots.push("/");
```
marco-c: Let's file a followup to remove this, the less test-specific code we have the better it is.

mykmelez:
On firstrun of a file-happy midlet running on the master branch, the most popular native is StringBuffer.append:

>     |  41988: Native java/lang/StringBuffer.append.(C)Ljava/lang/StringBuffer; 

But on this branch a bunch of Vector and String methods (plus StringBuffer.append with a different signature) occur more frequently:

>     |  193209: Compiled java/util/Vector.addElement.(Ljava/lang/Object;)V 
>     |  192704: Compiled java/util/Vector.elementAt.(I)Ljava/lang/Object; 
>     |  160370: Compiled java/util/Vector.size.()I 
>     |  103029: Native java/lang/String.charAt.(I)C 
>     |  100527: Native java/lang/String.indexOf.(II)I 
>     |  97098: Native java/lang/String.equals.(Ljava/lang/Object;)Z 
>     |  66967: Native java/lang/String.substring.(II)Ljava/lang/String; 
>     |  65373: Native java/lang/StringBuffer.append.(Ljava/lang/String;)Ljava/lang/StringBuffer; 
>     |  64374: Compiled java/util/Vector.<init>.(II)V 
>     |  64352: Compiled java/util/Vector.<init>.(I)V 
>     |  64318: Compiled java/util/Vector.<init>.()V 
>     |  41988: Native java/lang/StringBuffer.append.(C)Ljava/lang/StringBuffer;

mykmelez:
I don't see many calls to FileSystemRegistry.listRoots, Protocol.listRoots, nor DefaultFileHandler.listRoots/list, but those seem like the likeliest culprits for Vector usage, and they're implemented inefficiently. I'll see if better implementations make a difference.

mykmelez:
There's more we can do here, but a55fd62f541fba0b9ae67da05059e57e29066aab, f695f98a3cfaebc09953344faf518262bf99c4b2, and 0046eae09f9c26616caa48a3c006b699a4dcc179 should claw back the perf hit, especially the first of those changes.

marco-c:
I'm seeing *java/io/IOException: Root is not accessible* while running a MIDlet.

mykmelez:
> I'm seeing java/io/IOException: Root is not accessible while running a MIDlet.

#790 created *MIDP.fsRoots* but neglected to include *Persistent/* in the list. That was ok under the old implementation, which didn't enforce the restriction that a filesystem root be on the list of known roots. But it fails under the new implementation, which does enforce that restriction. 14b3334 fixes the problem by adding *Persistent/* to *MIDP.fsRoots*.

mykmelez:
(Note that *Persistent/* is not a filesystem root on the reference device that we used to determine the roots in #790. But it does get used by a midlet I'm testing, which apparently assumes that it exists.)

PR #1056, opened 2015-02-06
Implement missing resetNativeEventQueue native and misc cleanup of nativ...

marco-c:
...e events handling

### DIFF
--- a/midp/midp.js
+++ b/midp/midp.js
@@ -975,30 +975,22 @@ MIDP.copyEvent = function(obj, isolateId) {
         if (e[field.name] === undefined) {
           return;
         }
+
         field.set(obj, e[field.name]);
     });
 }
 
-MIDP.deliverWaitForNativeEventResult = function(resolve, nativeEvent, isolateId) {
-    if (MIDP.nativeEventQueues[isolateId].length > 0)
-        MIDP.copyEvent(nativeEvent, isolateId);
-    resolve(MIDP.nativeEventQueues[isolateId].length);
-}
-
-MIDP.sendEvent = function(obj, isolateId) {
-    var e = { type: obj.klass.classInfo.getField("I.type.I").get(obj) };
-    obj.klass.classInfo.fields.forEach(function(field) {
-        e[field.name] = field.get(obj);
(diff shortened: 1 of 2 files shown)


midp/midp.js
```diff
@@ -968,37 +968,23 @@ MIDP.foregroundIsolateId;
...
-    obj.klass.classInfo.fields.forEach(function(field) {
-        e[field.name] = field.get(obj);
-    });
-    MIDP.sendNativeEvent(e, isolateId);
+MIDP.copyEvent = function(e, obj, isolateId) {
+    var keys = Object.keys(e);
+    for (var i = 0; i < keys.length; i++) {
+      obj[keys[i]] = e[keys[i]];
+    }
 }
```
mykmelez: This function doesn't need an *isolateId* parameter anymore.

midp/midp.js
```diff
@@ -1049,47 +1035,45 @@ Native["com/sun/midp/events/EventQueue.getNativeEventQueueHandle.()I"] = functio
...
 
 Native["com/sun/midp/events/EventQueue.sendNativeEventToIsolate.(Lcom/sun/midp/events/NativeEvent;I)V"] =
 function(obj, isolateId) {
-    if (!MIDP.nativeEventQueues[isolateId]) {
-      MIDP.nativeEventQueues[isolateId] = [];
-    }
+    var e = { type: obj.type };
+    obj.klass.classInfo.fields.forEach(function(field) {
+        e[field.name] = field.get(obj);
+    });
```
mykmelez: *Array.forEach* is slower than a *for* loop, and this code could be perf-sensitive, so I'd use the latter here.

midp/midp.js
marco-c: Done in 89c96e4.

midp/midp.js
marco-c: This shouldn't matter much because sendNativeEventToIsolate is used just a couple of times (to start a MIDlet and to set a MIDlet as the foreground MIDlet), but I've done it anyway in 20416a5.

PR #1060, opened 2015-02-06
Increase threshold for `getMediaTime`

TimAbraldes:
This fixes test failures that I'm seeing locally on master.

### DIFF
--- a/tests/javax/microedition/media/TestAudioPlayer.java
+++ b/tests/javax/microedition/media/TestAudioPlayer.java
@@ -13,7 +13,7 @@ public class TestAudioPlayer implements Testlet, PlayerListener {
     public int getExpectedKnownFail() { return 0; }
     TestHarness th;
 
-    private static final long TIME_TOLERANCE = 100;
+    private static final long TIME_TOLERANCE = 175;
 
      /**
      * PlayerListener interface's method.


mykmelez:
@TimAbraldes This is worth it if needed to pass the test on your system, although we're really stretching the utility of the test when we tolerate a 175ms delta on a 500ms timer. Any idea why the values of *mediaTime* and *actualTimeSlept* are so different for you?

TimAbraldes:
Hmm. I really don't know! The values I was seeing for `actualTimeSlept` were about 505ms every time, and the values I saw for `mediaTime` were between 340ms and 365ms. Could there be something hardware-specific about my setup that causes these values to be so different from what others are seeing?

mykmelez:
> Hmm. I really don't know! The values I was seeing for actualTimeSlept were about 505ms every time, and the values I saw for mediaTime were between 340ms and 365ms. Could there be something hardware-specific about my setup that causes these values to be so different from what others are seeing?

Hmm, unsure. There could be, and cc: @yxl and @marco-c for their thoughts. But I'll take this anyway in the meantime, as I don't want unreliable tests to block development or make it less efficient.

yxl:
We changed the implementation of the audio player from using the Web Audio API to HTMLMediaElement. The media time got form HTMLMediaElement is not so accurate as that from the Web Audio API, especially for the short audio file.

PR #1061, opened 2015-02-09
Make Bluetooth support optional

marco-c:
Disabling Bluetooth support reduces the classes.jar size from 2,2 MB to 2,0 MB.


Makefile
```diff
@@ -6,9 +6,14 @@ RELEASE ?= 0
 VERSION ?=$(shell date +%s)
 PROFILE ?= 0
 
+# Sensor support
 JSR_256 ?= 1
 export JSR_256
 
+# Bluetooth support
+JSR_082 ?= 1
```
yxl: I guess we could turn JSR_082 off by default?

PR #1067, opened 2015-02-10
Raise EventQueue threads priority

marco-c:


marco-c:
I'm seeing the following error while running the tests:
*Error: Expected undefined return value during unwind, got 0 in java/util/Vector.size.()I*

mbebenita:
This is usually caused by some code that attempts to yield, but there is no unwinding code for it. At a later time, a function is called that is followed by unwind code, which reports this assertion, so:
```
result = foo() // Unwinds, by setting the global U property to 1.
// Missing Unwind Code

result = bar() // Doesn't unwind, but the global U property is still 1.
if (U) { ... } // asserts that result is undefined, and it is not in this case because bar() return successfuly
```

The bug is most likely caused by the preemption code, @brendandahl will submit a patch to disable it.

marco-c:
Now, I don't see errors while running the tests, but Firefox hangs.
The tests pass if I disable JIT compilation.

Issue #976, opened 2015-01-29
Improve heuristic for determining that keyboard is visible/hidden


TimAbraldes:
The current heuristic has false negatives and false positives, especially during rotation

TimAbraldes:
I've got something locally for this; will create a PR when I've unbroken tests

TimAbraldes:
Closed via #1073

PR #1073, opened 2015-02-10
Improve keyboard handling. Implement rotation.

### LINKED ISSUE #975: Support device rotation (landscape in addition to portrait)
When the device is rotated, we can send rotation events that will be processed by J2ME and the running MIDlet

TimAbraldes:
Fixes #975 and #976.

Instead of creating our own thread on which to notify the listener of
keyboard visibility events, pass the VIRTUAL_KEYBOARD event to j2me and
process it in the `Display` class by notifying the listener there.

Improve heuristic for determining whether the keyboard is visible; we
now check against the expected inner and outer dimensions of the screen
(which we are able to determine on app launch) instead of against the last
known inner dimensions.

Handle device rotation by forwarding new canvas dimensions to j2me as a
native rotation event upon window resize. Note that the MIDlet will
never believe that it is rotated - only that it's dimensions have
changed. To have the MIDlet actually know about rotation, I think we
would have to apply a rotation to our canvas whenever the device
orientation changes.


midp/midp.js
```diff
@@ -541,28 +500,72 @@ Native["com/sun/midp/util/ResourceHandler.loadRomizedResource0.(Ljava/lang/Strin
...
+        return false;
+    } else if (window.innerHeight < expectedHeightWithNoKeyboard) {
+        return true;
+    } else {
+        console.warn("window is taller than expected in isVKVisible!");
+        return false;
+    }
+};
+
+MIDP.isVKVisibleDebug = function() {
```
mykmelez: Nit: it wasn't immediately obvious to me that this is the implementation of isVKVisible when autosize is disabled, so it'd help to have a comment here explaining that.

mykmelez: Since these are only ever used in one place, you could even refactor them to be declared where they're used.

java/custom/javax/microedition/lcdui/Display.jpp
```diff
@@ -3666,6 +3670,13 @@ public class Display {
                 }
             }
             // #endif ENABLE_CHAMELEON
+
+            KeyboardVisibilityListener listener = Display.this.getKeyboardVisibilityListener();
```
marco-c: We could make com.nokia.mid.ui.VirtualKeyboard listen for this event, so we wouldn't need to modify Display.jpp.

PR #1079, opened 2015-02-11
Remove DisplayEventListener modifications

marco-c:


java/custom/com/nokia/mid/ui/gestures/GestureRegistrationManager.java
```diff
@@ -13,41 +15,77 @@ public ZoneRegistration(Object container, GestureInteractiveZone zone) {
...
-            listenerRegistrations.remove(container);
+    static {
+        EventQueue eventQueue = EventQueue.getEventQueue();
+        eventQueue.registerEventListener(EventTypes.GESTURE_EVENT, GestureRegistrationManager.getRegistrationManagerInstance());
+    }
+
+    private static GestureRegistrationManager gestureRegistrationManager = null;
+    private static GestureRegistrationManager getRegistrationManagerInstance() {
+        if (gestureRegistrationManager == null) {
+            gestureRegistrationManager = new GestureRegistrationManager();
```
TimAbraldes: This method isn't thread-safe. Is it possible that multiple threads might call `getRegistrationManagerInstance` at the same time? If so, we should add `synchronized` to its declaration

java/custom/com/nokia/mid/ui/gestures/GestureRegistrationManager.java
```diff
@@ -13,41 +15,77 @@ public ZoneRegistration(Object container, GestureInteractiveZone zone) {
...
 
             GestureListener listener = (GestureListener)listenerRegistrations.get(zoneReg.container);
             if (listener != null) {
-                listener.gestureAction(zoneReg.container, zone, event);
+                listener.gestureAction(zoneReg.container, zone, gestureEvent);
             }
         }
     }
 
+    public static void setListener(Object container, GestureListener listener) throws IllegalArgumentException {
```
TimAbraldes: Why did this move? Can we move it back so it doesn't show up in the diff? Feel free to move it again in a follow-up patch

TimAbraldes:
Overall LGTM. Just a couple nits pointed out as line notes

java/custom/com/nokia/mid/ui/gestures/GestureRegistrationManager.java
marco-c: I'm unsure why git shows this as moved, I haven't moved it (I've just added code on top of it).

java/custom/com/nokia/mid/ui/gestures/GestureRegistrationManager.java
marco-c: The getRegistrationManagerInstance method is only called by the static initializer (that is thread-safe and is called just once on class initialization).

Issue #1065, opened 2015-02-10
Investigate using new Function() instead of eval to load code.


mbebenita:
@tschneidereit mentioned that `eval()` is considerably slower than `new Function()`. I've tried this in the past and that was not the case, perhaps something has changed since then. We should investigate doing this.

tschneidereit:
As a bit of background, Shumway changing from `new Function()` to `eval` caused the huge regression on February 7 that can be seen in [these graphs](http://arewefastyet.com/#machine=28&view=breakdown&suite=shumway).

tschneidereit:
I [tested this](https://gist.github.com/tschneidereit/3c3a247b086a84eb9db9), with mostly definitive results on the JITBenchmarks: while `hashtable` got almost 10% faster, `arrayTypeCheck` got about 200% worse. This is with roughly current SpiderMonkey Nightly, so for all I know numbers in other engines could be wildly different. It seems worthwhile to, at some point, look into what causes these differences, but it won't be a useful change for j2me.js right now.

mbebenita:
Thanks for looking into it. I have a slightly modified version of the test harness now that call the GC between runs. I'll apply your patch and rerun the benchmark.

PR #1085, opened 2015-02-11
Use new Function() instead of eval() in linkMethod().

sstangl:
From experience with Shumway, https://github.com/mozilla/shumway/commit/81d53f0ee1f375a35835825e40be4eadcee60940, "new Function()" is significantly faster than "eval()", the latter being a huge deopt point in the JS engine. Some internal refactoring could likely get rid of the anonymous wrapper too.

The majority of the functions produced by j2me.js are still through eval() after this patch, in emitKlassConstructor(). I attempted to make a similar change to that function, but it results in a "too much recursion" error that I don't understand.

### DIFF
--- a/runtime.ts
+++ b/runtime.ts
@@ -1565,13 +1565,14 @@ module J2ME {
   export function linkMethod(methodInfo: MethodInfo, source: string, referencedClasses: string[]) {
     jitWriter && jitWriter.writeLn("Link method: " + methodInfo.implKey);
 
-    enterTimeline("Eval Compiled Code");
-    // This overwrites the method on the global object.
-    (1, eval)(source);
-    leaveTimeline("Eval Compiled Code");
+    enterTimeline("Link Compiled Code");
+    // Still not as bad as eval!
+    var fn = new Function('return ' + source)();
+    leaveTimeline("Link Compiled Code");
 
     var mangledClassAndMethodName = methodInfo.mangledClassAndMethodName;
-    var fn = jsGlobal[mangledClassAndMethodName];
+
+    jsGlobal[mangledClassAndMethodName] = fn;
     methodInfo.fn = fn;
     methodInfo.state = MethodState.Compiled;
 


marco-c:
See also #1065, where @tschneidereit run the benchmarks before and after moving to *new Function*, and before was faster.

tschneidereit:
Note that in Shumway the comparison was between direct eval and `new Function`, whereas j2me.js uses indirect eval. I tried doing the same in Shumway, where it was slightly slower than `new Function`, but not nearly by as much as direct eval. It'd still be nice to get a better understanding of why some things are faster with one form, some with the other.

marco-c:
I've tried to run the startup benchmarks on desktop after applying this change, unfortunately the result was insignificant:
| startupTime |       1,976ms | 1,863ms | -113ms | -5.73 | INSIGNIFICANT | 1,684ms | 2,029ms |

marco-c:
On a Flame:
| startupTime   |      15,443ms | 15,469ms |   25ms |   0.16 | INSIGNIFICANT | 14,498ms | 17,367ms | 
| totalSize     |      31,387kb | 31,316kb |  -71kb |  -0.23 | INSIGNIFICANT | 28,320kb | 32,258kb | 
| domSize       |          93kb |     93kb |    0kb |  -0.20 |        BETTER |     92kb |     93kb | 
| styleSize     |         328kb |    328kb |    0kb |  -0.08 | INSIGNIFICANT |    327kb |    329kb | 
| jsObjectsSize |      16,317kb | 16,374kb |   56kb |   0.35 | INSIGNIFICANT | 16,099kb | 17,093kb | 
| jsStringsSize |       1,632kb |  1,074kb | -558kb | -34.18 |        BETTER |  1,055kb |  1,083kb | 
| jsOtherSize   |      12,620kb | 13,054kb |  434kb |   3.44 | INSIGNIFICANT |  9,754kb | 14,051kb | 
| otherSize     |         397kb |    394kb |   -3kb |  -0.75 | INSIGNIFICANT |    380kb |    409kb |

mykmelez:
> The majority of the functions produced by j2me.js are still through eval() after this patch, in emitKlassConstructor(). I attempted to make a similar change to that function, but it results in a "too much recursion" error that I don't understand.

Note that I'm working on eliminating the eval() call from emitKlassConstructor over in #1293.

> Note that in Shumway the comparison was between direct eval and new Function, whereas j2me.js uses indirect eval. I tried doing the same in Shumway, where it was slightly slower than new Function, but not nearly by as much as direct eval. It'd still be nice to get a better understanding of why some things are faster with one form, some with the other.

Indeed. Who might be able to explain these results? Perhaps someone on the SpiderMonkey team?

FWIW, direct eval beats both indirect eval and `new Function` on [this jsperf benchmark](http://jsperf.com/eval-vs-indirect-eval-vs-new-function) when I run it in various versions of Firefox for desktop. So it would be interesting to benchmark switching from indirect to direct eval.

TimAbraldes:
Are we still pursuing this? We need a rebase if so

mykmelez:

PR #1086, opened 2015-02-12
Fix static initialization code.

### LINKED ISSUE #1049: Race condition issue with clinit

brendandahl:
Fixes #1049

### DIFF
--- a/context.ts
+++ b/context.ts
@@ -482,8 +482,7 @@ module J2ME {
     }
 
     pushClassInitFrame(classInfo: ClassInfo) {
-      if (this.runtime.initialized[classInfo.className] ||
-          this.runtime.pending[classInfo.className]) {
+      if (this.runtime.initialized[classInfo.className]) {
         return;
       }
       var needsInitialization = true;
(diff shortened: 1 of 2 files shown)


brendandahl:
Arg....don't merge. I should make add some synchronization code so we wait for both threads.

brendandahl:
It should be noted that if the code is compiled this will still fail. I'm working on a fix for this, but it's part of larger fix to re-enable preemption.

brendandahl:
I still think there's another issue, closing this PR for now and will open a new one.

Issue #875, opened 2015-01-16
Don't directly draw to screen, but draw in an offscreen canvas and on screen only when refresh is called


marco-c:

marco-c:
I was trying to do this in https://github.com/marco-c/j2me.js/tree/offscreen_canvas.
Looks faster on desktop, but unfortunately not on mobile.

marco-c:
Looks like the Display class is already drawing in an offscreen canvas, so this won't probably improve performance. It does help in removing the flickering.

mykmelez:
> Looks like the Display class is already drawing in an offscreen canvas, so this won't probably improve performance.

Would it help to override the Display class natives responsible for flushing that buffer?

> It does help in removing the flickering.

Hmm, that's useful.

marco-c:
I'm closing this since the Display class is already doing this and another PR fixed the flickering.

We might try overriding the Display methods though, since they're used a lot.

PR #1093, opened 2015-02-13
Buffer draws. Implement `DisplayDevice.refresh0`.

TimAbraldes:
MIDlets expect that they can draw to the screen `Graphics` object all
they want without it appearing on screen until `DisplayDevice.refresh`
has been called. This patch makes our implementation more like what
MIDlets actually expect, and in my testing drastically reduces
flickering.


marco-c:
I've tried doing this in #875, but it was noticeably slower. We've implemented a few gfx optimizations since then, so the situation might be different now.
I'll test again on my Flame.

TimAbraldes:
> I've tried doing this in #875, but it was noticeably slower. We've implemented a few gfx optimizations since then, so the situation might be different now.
> I'll test again on my Flame.

Thanks Marco!

It seemed like things were about the same speed on my Flame device after applying this patch, but they sometimes *felt* slower because it wasn't very obvious that anything was happening (since drawing operations weren't being immediately flushed to the screen). I think the overall experience is probably better without the extra drawing, but please let me know how your testing goes!

TimAbraldes:
Ping. How did testing this patch go? Did you see any regressions, perf or otherwise?

marco-c:
I've just re-tested this, I think it's like you said in your comment.
Without this patch, things seem faster because we're progressively drawing something, while with this patch we're drawing everything at the end.

I think we should do this, but I think we should wait until we're a little bit faster.

TimAbraldes:
We're eventually going to have to implement this for the sake of correctness; MIDlets could conceivably do just about anything to the canvas while they believe it's not flushing to the screen - indeed some apps redraw the entire screen multiple times before it gets flushed.

In the meantime, it might be harder to work on drawing perf if we keep the existing implementation: Operations that are hot or slow in the existing implementation might have very little impact in the new implementation or vice versa.

Maybe we should land an implementation of this and revert it if it's slowing down our development?

marco-c:
> We're eventually going to have to implement this for the sake of correctness

The standard doesn't actually require double-buffering (indeed, in the RI there are different implementations of refresh0, some are actually refreshing the screen, some are no-ops), so correctness shouldn't be a problem.

(Some MIDlets also implement double-buffering manually.)

We should figure out whether double-buffering is enabled on the platform we're trying to emulate, that is a Nokia Asha. I'll try to do it with my phone.

Anyway, given that the standard doesn't require double-buffering, I think we should go with the option that is faster (or, in this case, looks faster :smile:). We should also test on a lower end phone than the Flame, where the deferred flushing could be even more noticeable.

mykmelez:
> We should figure out whether double-buffering is enabled on the platform we're trying to emulate, that is a Nokia Asha. I'll try to do it with my phone.

@marco-c Any results yet?

marco-c:
Double-buffering is enabled on a Nokia Asha, so it would be better to enable it in j2me.js as well, but it isn' t mandatory to do so.
So I think we should either defer implementing it after we're fast enough (because right now with double-buffering MIDlets sometimes feel less responsive) or implementing it with a way to disable it with a configuration parameter.

TimAbraldes:
This is made obsolete by the patch in #1198

PR #1096, opened 2015-02-13
Convert MIDP to module pattern

TimAbraldes:
This PR is separated into two distinct patches. If you want to review this PR, please look at the diffs of the two patches separately. The first patch is intended to have as a few differences as possible with master to make it more readable. The second patch is a whitespace-only change.

Relevant info from first patch:
  "MIDP has been converted from an object to a module. `MIDP.Context2D` has been converted from a module to the member variable `MIDP.context2D`. Invocations of `MIDP.sendNativeEvent` have been replaced with `MIDP.send*Event` function invocations. This provides a dual benefit: Those functions now do a little bit of validation that we weren't doing before, and we don't have to expose `MIDP.sendNativeEvent` or `MIDP.foregroundIsolateId`"


mykmelez:
Needs a rebase. And this one'll get out-of-date quickly. But fortunately it should be easy to review as well.

TimAbraldes:
I'm not sure who wants to take this review but I'm kind of dreading a future merge if someone makes extensive changes to midp.js. Please let me know if this patch is simply too much code churn; I'd rather make that decision explicitly so I don't end up rebasing it a bunch of times :)

mykmelez:
@marco-c Does this look good to you now?

marco-c:
Yes!

PR #1097, opened 2015-02-13
finish clarifying licenses

mykmelez:
This branch moves the GPL v2 license file from java/LICENSE to the top-level LICENSE file, since that's the license for the project as a whole.

It isn't ready to merge yet, because we still need to figure out how to identify the code that originally came from the node-jvm project, which is in the following files:

* classfile/classfile.ts
* classfile/reader.ts
* classRegistry.ts
* context.ts
* jvm.ts
* native.js (just the TAGS object)
* opcodes.js
* runtime.ts

cc: @gerv


mykmelez:
> It isn't ready to merge yet, because we still need to figure out how to identify the code that originally came from the node-jvm project, which is in the following files:

@gerv previously suggested moving such code to a subdirectory with its own LICENSE file, so I've done that in 6c633a6 (and 7b43a9a). I'd like to hear from him that this is what he recommends in this particular case, however, so this PR remains a WIP.

Note: I moved the *TAGS* structure out of native.js and into vm/tags.js so I wouldn't have to move all of native.js into the subdirectory, since *TAGS* is the only part of that file that comes from node-jvm.

I also removed opcodes.js, which isn't being used anymore. There's a data structure in tests/opcodeCoverage.py that looks like it might be a copy of a structure in opcodes.js (rewritten from JS to Python), but it's hard to know, and in any case we would reimplement it identically, so it doesn't seem like the kind of thing that would require a reference to node-jvm. @gerv should confirm, though.

gerv:
Hey. Subdirectories with their own clear LICENSE file are manna from heaven :-) Always a good idea. Don't worry about the opcodes stuff. Ship it :-)

PR #1099, opened 2015-02-13
Fix calculations in `setClip` and `clipRect`

TimAbraldes:
I ported this logic from Java and, upon looking at it, realized that
it could have been a little cleaner. I don't think this changes
behavior, but it make the logic easier to follow


marco-c:
To avoid bugs in the future, I'd keep the RI's behavior whenever possible (that is, unless we're improving perf or fixing a bug).

TimAbraldes:
I agree that we should try to keep the RI's behavior.

In this case we've got some JavaScript that was converted from Java and the conversion was messy. Some of the code is strange (i.e. makes no sense in JavaScript but made sense in Java) and some of the code is wrong (e.g. in `setClip` we fail to ever reset the `clipped` member of the `Graphics` object if its clipping rectangle gets reset). To properly keep the RI's behavior, I think we need to modify the JS to look more like JS.

I'll try to point out in line comments what this patch attempts to fix about the current implementation.

midp/gfx.js
```diff
@@ -1039,82 +996,24 @@ var currentlyFocusedTextEditor;
...
-        translatedX1 = x + g.transX;
-        translatedY1 = y + g.transY;
-
-        // Detect overflow
-        if (translatedX1 < 0) {
-            translatedX1 = (x < 0 || g.transX < 0) ? 0 : g.maxWidth;
-        }
-        if (translatedY1 < 0) {
-            translatedY1 = (y < 0 || g.transY < 0) ? 0 : g.maxHeight;
-        }
```
TimAbraldes: This section of code doesn't make sense in JavaScript; we're not going to "overflow" a JavaScript `Number` and, if we do, we're not going to get a negative number - we'll get `Number.POSITIVE_INFINITY`

midp/gfx.js
```diff
@@ -1039,82 +996,24 @@ var currentlyFocusedTextEditor;
...
+        } else {
+            g.clipX1 = Math.max(g.clipX1, x + g.transX) & 0x7fff;
+            g.clipY1 = Math.max(g.clipY1, y + g.transY) & 0x7fff;
 
-        // Detect Overflow
-        if (translatedX2 < 0) {
-            translatedX2 = (x < 0 || g.transX < 0) ? translatedX1 : g.maxWidth;
-        }
-        if (translatedY2 < 0) {
-            translatedY2 = (y < 0 || g.transY < 0) ? translatedY1 : g.maxHeight;
```
TimAbraldes: This section is also attempting to "detect overflow" in a way that doesn't make sense in JS.

midp/gfx.js
```diff
@@ -896,74 +896,31 @@ var currentlyFocusedTextEditor;
     }
 
     function setClip(g, x, y, width, height) {
```
TimAbraldes: This function doesn't ever set `g.clipped` to false so if a `Graphics` object ever becomes clipped, we will always think it is clipped

TimAbraldes:
> In this case we've got some JavaScript that was converted from Java and the conversion was messy. Some of the code is strange (i.e. makes no sense in JavaScript but made sense in Java) and some of the code is wrong (e.g. in setClip we fail to ever reset the clipped member of the Graphics object if its clipping rectangle gets reset). To properly keep the RI's behavior, I think we need to modify the JS to look more like JS.

I realize now that this makes my original statement:

> I don't think this changes behavior, but it make the logic easier to follow

Completely false :)

marco-c:
Thank you, this makes more sense now :smile:

midp/gfx.js
marco-c: This will also improve performance, because in many natives we don't need to call save() if the graphics element isn't clipped!

marco-c:
I wonder if we could improve perf by not always saving/restoring the canvas state in each native, but setting the canvas clip in setClip (I'll file an issue).

marco-c:
Could you write a few gfx tests with clipping? Currently, we only have two very basic tests.

TimAbraldes:
> I wonder if we could improve perf by not always saving/restoring the canvas state in each native, but setting the canvas clip in setClip (I'll file an issue).

Yeah this would be really great to be able to do. In `clipRect` we could definitely shrink the clipped region. Things get complicated when we have a `setClip` call that is trying to expand the clipped region (or reset it entirely). There's supposed to be a `resetClip` function - https://html.spec.whatwg.org/multipage/scripting.html#dom-context-2d-resetclip - but we don't seem to have it available to us.

TimAbraldes:
> Could you write a few gfx tests with clipping? Currently, we only have two very basic tests.

Filed #1104. I'll start working on those.

TimAbraldes:
@marco-c I think I found a way to set our canvas clip in `setClip` instead of each native. Let me know if this is along the lines of what you were thinking

marco-c:
Yes! I think this will reduce calls to save and restore a lot, because many Graphic objects are clipped!

marco-c:
P.S.: Do you want to work on the tests as part of this PR or do you want to merge this PR first?

TimAbraldes:
I'd prefer to merge sooner than later; less rebasing effort :)

marco-c:
Unfortunately, I'm seeing a MIDlet calling setClip a lot of times (like, ~300 times for a simple interaction), so saving/restoring the canvas when calling setClip could actually introduce a regression instead of improving perf.

We should investigate how many times the MIDlet is trying to set the clip to the same old value (because in this case we could optimize by just returning early and without saving/restoring the state).
If most of the times the clip is actually changing, then the old behavior might be better.

TimAbraldes:
I've updated the patch to only save/restore the context if the clipping rectangle has grown. If the clipping rectangle stays the same size (and position) or shrinks, we do not save/restore the canvas context.

I would expect the new behavior to be better than the old behavior in every reasonable situation. I presume that for each `setClip` call there is going to be at least one drawing operation, and I imagine that in many cases there will be multiple drawing operations for a single `setClip` call. I'll see if I can get data to this effect.

TimAbraldes:
OK I tested in a real MIDlet with some logging around `save` and `restore` calls. Without the patch, a common interaction in the MIDlet called `save` around 450 times. With the patch, the same interaction called `save` around 65 times.

TimAbraldes:
Although... don't merge yet. I think I'm seeing a visual artifact and I'd like to see if I can find out what's causing it.

marco-c:
Shouldn't we save/restore also when the rectangle shrinks?

TimAbraldes:
OK I fixed the visual artifact I was seeing. I guess those extra tests really are important :)

> Shouldn't we save/restore also when the rectangle shrinks?

I don't think so. When we further reduce the area that we are clipped to, the canvas `clip` call will work correctly. We only need to save/restore if we're trying to expand the clipping rectangle.

TimAbraldes:
Gah, just found another issue. Please don't merge yet

marco-c:
> I don't think so. When we further reduce the area that we are clipped to, the canvas clip call will work correctly. We only need to save/restore if we're trying to expand the clipping rectangle.

Ah, ok, I misunderstood what you meant (I thought you weren't calling *clip* when the area was smaller, instead you're just not calling save/restore)

TimAbraldes:
Fixed the remaining issue that I was aware of. I'm surprised that that one wasn't caught by existing tests :-/

PR #1106, opened 2015-02-16
Self host GestureEventImpl and allow running benchmarks using SpiderMonkey

marco-c:


marco-c:
This also makes it possible to run benchmarks using SpiderMonkey.

TimAbraldes:
This looks good to me. Please fix the typo and I'll merge

PR #1107, opened 2015-02-17
SHA-256 test and benchmark

marco-c:
Depends on the changes in #1106 (to be able to run the benchmark using SpiderMonkey)


bench/Makefile
```diff
@@ -3,7 +3,7 @@ SRCS=$(shell find . -name "*.java")
 benchmark.jar: $(SRCS)
 	rm -rf build
 	mkdir build
-	javac -source 1.3 -target 1.3 -encoding UTF-8 -bootclasspath ../java/classes.jar -extdirs "" -d ./build $(SRCS) > /dev/null
+	javac -source 1.3 -target 1.3 -encoding UTF-8 -bootclasspath ../java/classes.jar:../tests/tests.jar -extdirs "" -d ./build $(SRCS) > /dev/null
```
mykmelez: If benchmark.jar is going to depend on tests.jar, then the *benchmarks* target should depend on the *tests* target (or at least the *tests/tests.jar* target). And while you're fixing that, you should also make it depend on the *java* (or *java/classes.jar*) target, since it needs classes.jar.

(Unrelated, but you might as well make the change in the process: *tests* should depend on *jasmin*.)

PR #1117, opened 2015-02-18
Implement streaming SHA-1 hash

marco-c:
The benchmark introduced with #1115 takes ~5 seconds on master, ~300 ms with this branch.


mbebenita:
Any idea why we're seeing such a large perf improvement?

marco-c:
The old implementation was "simulating" progressive hashing, so it was particularly slow (https://github.com/mozilla/j2me.js/pull/1117/files#diff-ad08ecd00b1b5c89795ffccbf3c657f5L23)

mykmelez:
I'm unsure how to properly identify a file with code from two other files that have somewhat different licenses, but I suspect we shouldn't give it two copyright/license headers, which is what this changeset does to java/custom/com/sun/midp/crypto/SHA.java. cc: @gerv for input.

marco-c:
Actually, I think there's nothing left of the old implementation (only the comments that explain the interface of the functions are the same), so I've just removed the old license header.

mykmelez:
The difference is even starker on device:

Before:

"SHA-1: 37318"
"SHA-1: 34517"
"SHA-1: 25757"

After:

"SHA-1: 1948"
"SHA-1: 1806"
"SHA-1: 1958"

PR #1119, opened 2015-02-18
WIP: Add async attribute to the libs/relooper.js script to let Firefox cache ...

marco-c:
...its compilation

Loading from the cache is slower on desktop (160-180 ms vs ~50 ms with compilation), but on mobile it should be faster (compilation takes ~500 ms).

### DIFF
--- a/main.html.in
+++ b/main.html.in
@@ -33,7 +33,7 @@
   <script type="text/javascript" src="util.js" defer></script>
 
   <script type="text/javascript" src="libs/compiled-method-cache.js" defer></script>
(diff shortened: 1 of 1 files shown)


EricRahm:
Testing on flame I found that loading from the cache took 1050ms on average vs compiling at 470ms on average. The numbers might be different on less substantial devices though. My guess is that the `async` requirement is pushing logic off the main thread and slowing down due to IPC overhead/scheduling. I'll follow up w/ the JS team to see if I can find more out.

PR #1120, opened 2015-02-18
Benchmarks for DataInputStream and DataOutputStream

marco-c:


bench/DataInputOutputStreamBench.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import com.sun.cldchi.jvm.JVM;
+
+public class DataInputOutputStreamBench {
+  void writeUTF(OutputStream out) throws IOException {
+      DataOutputStream dos = new DataOutputStream(out);
+
+      for (int i = 0; i < 25000; i++) {
+          dos.writeUTF("abcdefghilmnopqrstuvzABCDEFGHILMNOPQRSTUVZabcdefghilmnopqrstuvzABCDEFGHILMNOPQRSTUVZ");
```
mykmelez: Any particular reason you left out "jk" and "wxy" from these sequences of alphabetic characters? And wouldn't it give us just as reasonable a measurement (albeit not identical) if we wrote a single sequence of a-zA-Z 50k times?

marco-c: I just copied & pasted the Italian alphabet :smile: 

I don't recall why I made a 25k loop instead of a 50k loop, so I've changed it with 17dd80b.

PR #1123, opened 2015-02-19
Self-hosting DataInputStream

marco-c:


marco-c:
The results before-after are basically the same on my machine, so the custom implementation is no longer needed.
It'd be great if someone could validate my results.

marco-c:
(This PR includes the commit from #1120, where I'm adding benchmarks for these two classes)

mykmelez:
> The results before-after are basically the same on my machine, so the custom implementation is no longer needed.
> It'd be great if someone could validate my results.

Are you testing on device? On my Flame with FxOS 2.0, readUTF is a bit faster, but writeUTF is significantly slower, with these changes:

DataInputOutputStreamBench Before:

DataOutputStream::writeUTF: 1120, 1117, 1177
DataInputStream::readUTF: 1686, 1633, 1650

DataInputOutputStreamBench After:

DataOutputStream::writeUTF: 2153, 1827, 1944
DataInputStream::readUTF: 1536, 1420, 1410

DataInputOutputStreamFileBench Before:

DataOutputStream::writeUTF in file: 2563, 2328, 2448
DataInputStream::readUTF from file: 2482, 2231, 2202

DataInputOutputStreamFileBench After:

DataOutputStream::writeUTF in file: 3168, 3973, 3156
DataInputStream::readUTF from file: 2134, 2213, 2055

marco-c:
No, I haven't run the benchmark on device (I was expecting them to have the same behavior).
The results are interesting, we should re-run the benchmarks before/after #996 on the Flame.

mykmelez:
> The results are interesting, we should re-run the benchmarks before/after #996 on the Flame.

Indeed. Can you do that and open an issue if that change caused a regression?

marco-c:
> Indeed. Can you do that and open an issue if that change caused a regression?

Done and filed #1149.

Issue #1104, opened 2015-02-14
Add gfx tests for clipping


TimAbraldes:
For testing #1099 and similar

PR #1127, opened 2015-02-19
Let context handle translation

TimAbraldes:
We've been holding onto the translation (x,y) amounts and adding them to
the positions in various `Graphics` operations. Instead, let's translate
the canvas context when the `Graphics` is translated so we don't have to
include translation amounts in our own calculations.

### DIFF
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -414,9 +414,6 @@ var currentlyFocusedTextEditor;
     var BASELINE = 64;
 
     function withAnchor(g, c, anchor, x, y, w, h) {
-        x += g.transX;
-        y += g.transY;
-
         if (anchor & RIGHT) {
             x -= w;
         } else if (anchor & HCENTER) {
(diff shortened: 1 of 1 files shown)


midp/gfx.js
```diff
@@ -686,26 +683,26 @@ var currentlyFocusedTextEditor;
...
     };
 
     Native["javax/microedition/lcdui/Graphics.getClipY.()I"] = function() {
-        return this.clipY1 - this.transY;
+        return this.clipY1;
     };
 
     Native["javax/microedition/lcdui/Graphics.getClipWidth.()I"] = function() {
-        return this.clipX2 - this.clipX1;
+        return this.clipX2;
```
marco-c: Why is this needed? Shouldn't the clipping area still be clipX2 - clipX1?

(The same applies to getClipHeight)

marco-c:
Looks good to me! Could you add a couple of gfx tests?

Currently, we don't have any gfx test that calls getClip* or translate.

The getClip* functions could be tested as part of #1104.

For translate, it would be great to have at least a basic test and a test that ensures that the effect of subsequent calls to translate are cumulative.

TimAbraldes:
> Looks good to me! Could you add a couple of gfx tests?
> 
> Currently, we don't have any gfx test that calls getClip* or translate.
> 
> The getClip* functions could be tested as part of #1104.
> 
> For translate, it would be great to have at least a basic test and a test that ensures that the effect of subsequent calls to translate are cumulative.

Yeah, I'm starting to realize that I should have created a test for the `Graphics` object after porting it to JS. I'll modify #1104 to be about adding tests for all the functions in the `Graphics` object

midp/gfx.js
TimAbraldes: Oops!! Good catch!

PR #1131, opened 2015-02-19
Download SlimerJS and XULRunner on make test

marco-c:


marco-c:
Oops, I forgot to supporto other OSes, will update this.

Makefile
```diff
@@ -40,8 +42,48 @@ PREPROCESS_DESTS = $(PREPROCESS_SRCS:.in=)
 
 all: config-build java jasmin tests j2me shumway aot benchmarks
 
-test: all
-	tests/runtests.py
+$(shell mkdir build_tools)
```
mykmelez: This should be `mkdir -p build_tools`, otherwise it'll complain `mkdir: build_tools: File exists` after the first time you build.

PR #1132, opened 2015-02-19
Enable socket tests

marco-c:
Depends on #1131.


mykmelez:
Looks good, just needs its dependencies to land first.

PR #1135, opened 2015-02-20
Don't store the classes' data in a map

marco-c:
We basically have three representations of the same data at any time:
1) The JAR ArrayBuffer
2) This map where we store the deflated classes
3) The *classes* map where we store ClassInfo objects, built from the class data

This classFiles map isn't really useful, I've seen it used just for three data files in a MIDlet I'm testing. Caching the deflated data for thousands of classes when we only need to load only three files twice seems overkill.


mykmelez:
Looks good, just needs a rebase.

PR #1138, opened 2015-02-20
Use Class.java static initialization code.

### LINKED ISSUE #1049: Race condition issue with clinit

brendandahl:
Fixes #1049

We've had a number of issues with our static initialization code and this will hopefully fix them. I was hoping to enable preemption in this pull request, but I'm still running into issues and I thought it would be better to get this going before the PR gets any bigger.

A few notes:
- Static initializers are no longer automatically run if you do $.$SomeClassName. I had to do this so we could get the js class(function) without running the static init.  To initialize you have to do $.$SomeClassName.classObject.intialize(); Though it's best to use J2ME.classInitCheck if this is needed.
- Tried a few different was of running the initializers, such as stubbing .initialize() once it was ready, but found just inlining the initialization check was the fastest.
- Modified Isolate.java by changing private static int  _API_access_ok = 1; so we don't have to use some hacky special case for the class.
- Modifed Class.forName so it is broken in to two parts so the unwinding is automatically taken care of by interpreter/compiled code.
- The preinited classes are now properly initialized per runtime
- Moved some of the isolate starting code so unwinds can occur and still be handled
- Removed the isolate test as its output order was non-deterministic. There's a testlet that more or less does the exact same test.
- Running the startup benchmark I see no difference in this vs old way.


mykmelez:
> We've had a number of issues with our static initialization code and this will hopefully fix them. I was hoping to enable preemption in this pull request, but I'm still running into issues and I thought it would be better to get this going before the PR gets any bigger.

Good thinking! Is there any way to make it smaller? I.e. parts we can break off and land separately?

> Modified Isolate.java by changing private static int _API_access_ok = 1; so we don't have to use some hacky special case for the class.

Does this mean all Isolates have API access, or are we removing access from the second and subsequent isolates via some other means?

> Removed the isolate test as its output order was non-deterministic. There's a testlet that more or less does the exact same test.

Hmm, it didn't used to be. Was that indeterminacy introduced by this branch? And which is the testlet that runs the same test?

brendandahl:
> Good thinking! Is there any way to make it smaller? I.e. parts we can break off and land separately?

I can try, but I think the only thing I can really pull out is the Class.forName stuff.

> Does this mean all Isolates have API access, or are we removing access from the second and subsequent isolates via some other means?

Previously we were setting this private static field to 1 for all isolates during the static initializer on every runtime. Always settings this to 1 should be no different than what we were doing before(correct or not).

> Hmm, it didn't used to be. Was that indeterminacy introduced by this branch? And which is the testlet that runs the same test?

The test still runs in a deterministic way in this branch, but the output order in general is not something we should rely on. This could change with any priority changes or when we enable premeption. 

gnu.testlet.vm.TestIsolate ensures a isolate static field is not changed when a new isolate is started.  There does seem to be a few more tests in this old tests such as checking the isolate id and making sure the waitForExit code works.  I can move these to a testlet test where we don't rely on output order.

marco-c:
> Previously we were setting this private static field to 1 for all isolates during the static initializer on every runtime. Always settings this to 1 should be no different than what we were doing before(correct or not).

Currently, we're setting it to 1 only for the first isolate.

I agree we should remove the test that relies on order, it was added at the beginning just to help development of the Isolate stuff.
One thing we should keep is testing that static values are different in different isolates (right now, in the TestIsolate testlet we're just testing that a static variable in the main isolate has the correct value, we should also test the value of the static variable in the child isolate).

brendandahl:
> Currently, we're setting it to 1 only for the first isolate.

I may be missing something, but we run the code in clinit, which will happen once per isolate/runtime. https://github.com/mozilla/j2me.js/blob/master/native.js#L285

brendandahl:
I attempted to rewrite the isolate test using LinkPortal, but soon discovered that this isn't fully implemented. There's not a great way to get information back from an isolate so I've just modified the expected output (we now give the main runtime a thread, which is why we have to increment the ids in the expected output).

I tried to tease apart the Class.forName, but this would have to change in both PR's so I don't really think there's a point to pulling it apart. The PR minus the new custom Java files is pretty small ~+250/-195 lines.

marco-c:
One way to do this would be adding test-only natives in tests/native.js (setting a value in the child isolate, checking it in the parent isolate)

interpreter.ts
```diff
@@ -26,6 +26,14 @@ module J2ME {
     traceWriter.writeLn(toDebugString(array) + "[" + index + "] (" + toDebugString(array[index]) + ")");
   }
 
+  function classInitAndUnwindCheck(classInfo: ClassInfo, pc: number) {
+    classInitCheck(classInfo);
+    if (U) {
```
marco-c: classInitCheck is already doing this for all but array classes.
Since classInitCheck is only used here, I think inlining it would make classInitAndUnwindCheck clearer.

native.js
```diff
@@ -275,65 +275,56 @@ Native["java/lang/Object.notifyAll.()V"] = function() {
...
-            frames.push(classInitFrame);
-        }
-    }
-    if (frames.length) {
-        $.ctx.executeFrames(frames);
+        $.ctx.executeFrames([Frame.create(clinit, [], 0)]);
     }
 };
 
+Native["java/lang/Class.invoke_verify.()V"] = function() {
```
marco-c: Nit: Use "addUnimplementedNative"

tests/java/lang/TestStaticInitRaceCondition.java
```diff
@@ -0,0 +1,38 @@
...
+public class TestStaticInitRaceCondition extends Thread implements Testlet {
+    private TestHarness th;
+    public int getExpectedPass() { return 2; }
+    public int getExpectedFail() { return 0; }
+    public int getExpectedKnownFail() { return 0; }
+
+    public void test(TestHarness th) {
+        this.th = th;
+        start();
+        run();
```
marco-c: run is called by start

vm/classRegistry.ts
```diff
@@ -62,6 +62,7 @@ module J2ME {
        * because they don't have any static state.
        */
       var classNames = [
+        "com/sun/cldc/isolate/Isolate", // Not used frequently, but needs setup before we start the isolates.
```
marco-c: This is already listed a few lines after

vm/runtime.ts
```diff
@@ -466,10 +465,42 @@ module J2ME {
...
+     * to a class initializer because there can be multiple calls into initialize from
+     * different threads that need trigger the Class.initialize() code so they block.
+     */
+    setClassInitialized(runtimeKlass: RuntimeKlass) {
+      var className = runtimeKlass.templateKlass.classInfo.className;
+      this.initialized[className] = true;
+    }
+
+    getRuntimeKlass(klass: Klass): RuntimeKlass {
+      release || assert(!(klass instanceof RuntimeKlass));
```
marco-c: Nit: add a message to these assertions (or remove them)

interpreter.ts
marco-c: You've added other calls to classInitCheck, so discard this comment :smiley:

PR #1140, opened 2015-02-22
Don't precompile asm.js libraries, because it's slower than recompiling

marco-c:
To actually use caching, we'd need to add the *async* attribute to the script tags, but in #1119 we've seen the loading from the cache is actually slower than recompiling.


mykmelez:
Makes sense, while we figure out why loading from the cache is slower (and fix that problem, if it's possible to do so on our side). Just needs conflict resolution.

PR #1141, opened 2015-02-22
Initialize classes as soon as possible

marco-c:
Instead of waiting for all the initialization steps to finish before initializing the built-in classes, we can initialize them as soon as we've finished loading the J2ME JAR (and while we're waiting for the FS IndexedDB transaction to finish).


main.js
```diff
@@ -35,6 +35,13 @@ var getMobileInfo = new Promise(function(resolve, reject) {
 
 var loadingPromises = [initFS, getMobileInfo];
 
+loadingPromises.push(load("java/classes.jar", "arraybuffer").then(function(data) {
+  CLASSES.addPath("java/classes.jar", data);
+  CLASSES.initializeBuiltinClasses();
+  // Load the Isolate class, that is needed to start the VM (see jvm.ts)
+  CLASSES.getClass("com/sun/cldc/isolate/Isolate");
```
mykmelez: *initializeBuiltinClasses* both loads and links classes, but *getClass* only loads them. Can we link *com/sun/cldc/isolate/Isolate* here?

PR #1160, opened 2015-02-25
JARStore

marco-c:
This PR introduces a new library to manage JARs.
1) The MIDlet JAR is downloaded and stored in a separate store and not in the FS.
2) We aren't storing the JAR data itself, but an already "parsed" object (a map { filename -> data })
3) When loading data from a stored JAR, we don't need to create a new buffer (because each file from the stored JAR has its own buffer)
4) Once we've loaded a class file from a stored JAR, we can release the memory associated with it (as before, because each file from the stored JAR has its own buffer)

Possible followups are:
- Investigate storing java/classes.jar in the JARStore too (since it's better for memory consumption and could provide better performance as well)
- Investigate storing a decompressed version of the MIDlet JAR instead of a compressed one (could be better for performance, worse for memory)
- Investigate compressing java/classes.jar (could be better for memory, worse for performance), right now we're not compressing it
- Periodically release classes that haven't been read (this would require class loading to be asynchronous, because some times we will need to reload a freed class from the DB)


libs/zipfile.js
```diff
@@ -322,7 +322,7 @@ function inflate(bytes) {
 
 var arrays = J2ME.ArrayUtilities.makeArrays(256);
 
-function ZipFile(buffer) {
+function ZipFile(buffer, extract) {
```
marco-c: The *extract* argument name is a bit misleading. It doesn't actually extract files from the JAR, it just creates different ArrayBuffers for each file (instead of different views on the same ArrayBuffer).

mykmelez:
@marco-c Is this still a WIP, and if so, why? It isn't clear from the commentary.

marco-c:
It is complete, but I need to test if it affects performance on mobile.

marco-c:
I've run the startup benchmark (with some problems, so I'm not totally sure about its results), and this seems to be slower (the difference between the means is ~1 second).
I think it's still worth it because we're going to use much less memory and because with some followup improvements we could get startup perf to be better (for example with #1168, we could load the JAR before the FS and start the MIDlet before the FS is completely loaded).

libs/jarstore.js
```diff
@@ -0,0 +1,196 @@
...
+    if (entry.compression_method === 0) {
+      bytes = entry.compressed_data;
+    } else if (entry.compression_method === 8) {
+      bytes = inflate(entry.compressed_data);
+    } else {
+      return null;
+    }
+
+    if (jar.isBuiltIn) {
+      return bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength)
```
EricRahm: Why the slice? Can't you just return bytes.buffer?

mykmelez:
On my https://github.com/mykmelez/j2me.js/tree/store-midlet-in-metastore branch, I created a "metastore" library to consolidate the various IndexedDB object stores (fs, compiled-method-cache) into a single database. We should do something like that here, so we stop having to recreate a bunch of database management code (and create multiple databases) every time we need a new store.

libs/jarstore.js
marco-c: This is pre-existing code, I can't return bytes.buffer because the code that uses this function is expecting an ArrayBuffer of a single file (while bytes.buffer for built-in JARs is the ArrayBuffer of the entire JAR)

marco-c: (I have another patch to make this possible, but it involves several changes throughout the codebase)

marco-c:
@mykmelez can we investigate the "metastore" in a separate PR? (I guess your branch is almost ready, so we'd only need to integrate this other store)

P.S.: I'm not totally convinced about it, because the upgrade code becomes more complex and the code in common between the different stores isn't actually too much. The "metastore" would allow us to have a single database and multiple object stores, but I don't know whether it's better for performance/memory/storage space or not.

marco-c:
> Investigate storing java/classes.jar in the JARStore too

I've done some measurements. In a MIDlet I'm testing, this would allow releasing ~800 kB of memory (soon after startup).

> Investigate compressing java/classes.jar

I've opened https://github.com/mozilla/j2me.js/pull/1163 for this. I'll run the startup benchmark to see whether it's significantly slower.

jsc.ts
```diff
@@ -65,8 +65,12 @@ jsGlobal.config = {
   args: "",
 };
 
+function Promise() {
+  // ...
```
mykmelez: …?

marco-c: jsc.ts is executed in the shell, so we need to define objects that are not available in the shell environment.

libs/jarstore.js
```diff
@@ -0,0 +1,196 @@
...
+  function installJAR(jarName, jarData, jadData) {
+    return openDatabase.then(function() {
+      return new Promise(function(resolve, reject) {
+        var zip = new ZipFile(jarData, true);
+
+        var transaction = database.transaction(OBJECT_STORE, "readwrite");
+        var objectStore = transaction.objectStore(OBJECT_STORE);
+        var request = objectStore.put({
+          jar: zip.directory,
+          jad: jadData ? jadData : null,
```
mykmelez: Nit: this could be simply `jadData || null`.

libs/jarstore.js
```diff
@@ -0,0 +1,196 @@
...
+
+  function loadJAR(jarName) {
+    return openDatabase.then(function() {
+      return new Promise(function(resolve, reject) {
+        var transaction = database.transaction(OBJECT_STORE, "readonly");
+        var objectStore = transaction.objectStore(OBJECT_STORE);
+        var request = objectStore.get(jarName);
+
+        request.onerror = function() {
+          console.error("Error loading: " + request.error.name);
```
mykmelez: Nit: I would include *jarName* in this error message.

libs/jarstore.js
```diff
@@ -0,0 +1,196 @@
...
+  var VERSION = 1;
+  var OBJECT_STORE = "files";
+
+  var database;
+  var jars = new Map();
+  var jad;
+
+  var upgrade = {
+    "0to1": function(database, transaction, next) {
+      database.createObjectStore(OBJECT_STORE);
```
mykmelez: It would be better to use inline keys by specifying a *keyPath* attribute in the *optionalParameters* argument to *createObjectStore*, as fs.js and compiled-method-cache.js do. Otherwise, it's expensive to retrieve all records. I know we don't currently do that here, and perhaps we never will, but the difference is so significant, the fix so simple, and the change so painful to make after-the-fact, that it would be better to do it now.

mykmelez:
> @mykmelez can we investigate the "metastore" in a separate PR? (I guess your branch is almost ready, so we'd only need to integrate this other store)

Yes, we can do that, although it will make that migration more complex.

> P.S.: I'm not totally convinced about it, because the upgrade code becomes more complex and the code in common between the different stores isn't actually too much.

I'm not sure what you mean by "more complex" here. The function to open a database seems similar in complexity to the existing ones, and there would only be one of them.

It's true that the code in common between the stores isn't that much, as it's "just" the code to open the database. But that code is significant, given that it can trigger the complex upgrade process.

And the simple object store built into the metastore obviates the need to create yet another object store the next time we need to store data like this.

> The "metastore" would allow us to have a single database and multiple object stores, but I don't know whether it's better for performance/memory/storage space or not.

I don't think it has a significant impact on those metrics, but it does reduce the risk of bugs in code that interacts with IndexedDB databases.

jsc.ts
mykmelez: > jsc.ts is executed in the shell, so we need to define objects that are not available in the shell environment.

Makes sense; can you put that in the comment?

marco-c:
> I'm not sure what you mean by "more complex" here. The function to open a database seems similar in complexity to the existing ones, and there would only be one of them.

I mean, in the upgrade functions we'd need to handle upgrading different object stores that will independently evolve.

mykmelez:
> I mean, in the upgrade functions we'd need to handle upgrading different object stores that will independently evolve.

Hmm, that seems to me like the same amount of complexity, but all in one place rather than scattered throughout the codebase. Of course it wouldn't work for independent modules that don't share state, but our non-third-party modules are quite interdependent.

Nevertheless, like I said, we can do it later, for some pain.

mykmelez:
Needs conflict resolution, and a better comment in jsc.ts's Promise stub.

PR #1162, opened 2015-02-25
replace transient paths with ignored files

mykmelez:
To save memory (and improve perf), we should completely ignore files that midlets write and never read, replacing "transient paths" with "ignored files."

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -24,7 +24,7 @@ var config = {
   // writes to them and return no data from reads.  It's useful for log files
   // to which midlets write frequently but never read.  Specify one by setting
   // a string pathname key to the boolean true value.
-  ignoredFiles: new Map(),
+  ignoredFiles: new Set(),
 };
 
 // The base directory of the app, relative to the current page.  Normally this
--- a/config/default.js
+++ b/config/default.js
@@ -22,8 +22,8 @@ var config = {
 
   // Ignored files are files that always exist and are empty, so we discard
   // writes to them and return no data from reads.  It's useful for log files
-  // to which midlets write frequently but never read.  Specify one by setting
-  // a string pathname key to the boolean true value.
+  // to which midlets write frequently but never read.  Specify one by adding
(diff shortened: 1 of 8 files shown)


config/default.js
```diff
@@ -19,6 +19,12 @@ var MIDlet = {
 var config = {
   main: "com/sun/midp/main/MIDletSuiteLoader",
   midletClassName: "RunTests",
+
+  // Ignored files are files that always exist and are empty, so we discard
+  // writes to them and return no data from reads.  It's useful for log files
+  // to which midlets write frequently but never read.  Specify one by setting
+  // a string pathname key to the boolean true value.
+  ignoredFiles: new Map(),
```
marco-c: Since we only need to add values and check if a value is present in *ignoredFiles*, we could use a *Set* instead of a *Map*.

mykmelez: > Since we only need to add values and check if a value is present in ignoredFiles, we could use a Set instead of a Map.

Ah, good point, I've switched to a set in 3628455.

config/default.js
```diff
@@ -19,6 +19,12 @@ var MIDlet = {
 var config = {
   main: "com/sun/midp/main/MIDletSuiteLoader",
   midletClassName: "RunTests",
+
+  // Ignored files are files that always exist and are empty, so we discard
+  // writes to them and return no data from reads.  It's useful for log files
+  // to which midlets write frequently but never read.  Specify one by setting
+  // a string pathname key to the boolean true value.
```
marco-c: Nit: the last statement no longer applies :smiley:

mykmelez: > Nit: the last statement no longer applies :smiley:

Nice catch! Fixed in d4a9d6f.

marco-c:
Looks like Travis is under maintenance, feel free to merge it yourself once a run is complete.

marco-c:
Travis is working on other PRs, but it's somehow "stuck" with this one. Since the first run was green and the other changes were trivial (and tests are passing locally), I'm going to merge this anyway.

PR #1163, opened 2015-02-25
WIP: Compress classes.jar file

marco-c:
This should save 900 kB of memory. It's a WIP because I want to test how much slower it is.


marco-c:
I've run the startup benchmark twice. The first time with 10 rounds, and the difference was not significant. The second time with 30 rounds, but the difference was still not significant.

PR #1166, opened 2015-02-25
Support custom emoji images

marco-c:


marco-c:
This should improve startup time (because we won't need to load images) and save some memory (~400 kB) for MIDlets that use custom emoji images.

libs/emoji.js
```diff
@@ -233,9 +233,940 @@ var emoji = (function() {
...
+    "1f1ee-1f1f7": { "sheet": 12, "x": 36 },
+    "1f1ef-1f1f4": { "sheet": 12, "x": 37 },
+    "1f1f0-1f1ff": { "sheet": 12, "x": 38 },
+    "1f1f1-1f1e7": { "sheet": 12, "x": 39 },
+    "1f1f5-1f1ea": { "sheet": 12, "x": 40 },
+    "1f1f5-1f1f9": { "sheet": 12, "x": 41 },
+    "1f1f8-1f1fe": { "sheet": 12, "x": 42 },
+    "1f1fa-1f1e6": { "sheet": 12, "x": 43 },
+    "1f1fa-1f1fe": { "sheet": 12, "x": 44 },
+    "1f1fd-1f1ea": { "sheet": 12, "x": 45 },
```
mykmelez: I know this is pre-existing, but why are all the "flag" emoji blank in this sheet?

native.js
```diff
@@ -601,6 +601,15 @@ Native["com/sun/cldc/io/ResourceInputStream.open.(Ljava/lang/String;)Ljava/lang/
         obj = J2ME.newObject(CLASSES.java_lang_Object.klass);
         obj.data = new Uint8Array(data);
         obj.pos = 0;
+
+        // Instead of loading the emoji images at startup, we load them
+        // as soon as MIDlets load them.
+        if (MIDlet.customEmojiImageRegex) {
```
mykmelez: I think we should put configuration parameters like this into the *config* parameter, to better isolate the changes that get made to configure the project to run a particular midlet.

libs/emoji.js
marco-c: The sheets are generated with the Twitter set, which doesn't contain these emoji (yet).

PR #1169, opened 2015-02-26
Reuse contexts on `ImageData` objects

TimAbraldes:
Reuse the existing canvas on `ImageData` objects instead of creating a
new one each for each operation.

For operations that only require a canvas temporarily, use have a
static canvas that we reuse over and over again rather than create one
per operation.

This definitely reduces the number of times we create canvases but I have to do some memory profiling to see if it actually reduces the amount of memory we're using.


marco-c:
Probably memory will be allocated when you set the width/height properties, deallocated when you set them to 0.
If this is true, we'll use the same amount of memory as before.

About perf, do you think using the same canvas and setting its width/height is going to be faster than creating new canvases?

midp/gfx.js
```diff
@@ -278,18 +273,40 @@ var currentlyFocusedTextEditor;
         converterFunc(rgbData, abgrData, width, height, 0, width);
 
         context.putImageData(ctxImageData, 0, 0);
-
-        setImageData(imageData, width, height, context);
     };
 
+    Native["javax/microedition/lcdui/ImageData.finalize.()V"] = function() {
```
marco-c: The finalize method is, in the standard JVM, called by the garbage collector before releasing an object. In our VM, we're never calling the finalize methods.

TimAbraldes: I suspected that might be the case; I'll just remove this impl from this PR

TimAbraldes:
> Probably memory will be allocated when you set the width/height properties, deallocated when you set them to 0.
> If this is true, we'll use the same amount of memory as before.

Yeah; if the old canvases are garbage-collected right away after the `ImageData` no longer holds a reference to them, then this patch won't save us memory. Intuitively it seems like reusing contexts will reduce the potential for leaking canvases and contexts, but I don't know how much of an issue this is in practice.

> About perf, do you think using the same canvas and setting its width/height is going to be faster than creating new canvases?

It'll definitely create less work for the garbage collector, and we get to avoid function calls in the `ImageDataFactory` methods, but this PR also converts some `ImageData` methods to hand-written JS, which I'm told is slower than compiled Java. With the patch applied there's no noticeable performance difference so I'll have to profile to determine whether perf has been meaningfully affected

marco-c:
> Yeah; if the old canvases are garbage-collected right away after the ImageData no longer holds a reference to them, then this patch won't save us memory. Intuitively it seems like reusing contexts will reduce the potential for leaking canvases and contexts, but I don't know how much of an issue this is in practice.

I suspect we're going to release the memory eventually anyway, but this PR will most probably release it earlier (as soon as we set width/height to 0).

marco-c:
> It'll definitely create less work for the garbage collector, and we get to avoid function calls in the ImageDataFactory methods, but this PR also converts some ImageData methods to hand-written JS, which I'm told is slower than compiled Java. With the patch applied there's no noticeable performance difference so I'll have to profile to determine whether perf has been meaningfully affected

Yeah, I think we could obtain the same effect (reuse contexts) without creating a custom ImageData.

midp/gfx.js
```diff
@@ -278,10 +268,27 @@ var currentlyFocusedTextEditor;
...
+    function setImageDataDimensions(imageData, w, h) {
+        imageData.klass.classInfo.getField("I.width.I").set(imageData, w);
+        imageData.context.canvas.width = w;
+
+        imageData.klass.classInfo.getField("I.height.I").set(imageData, h);
+        imageData.context.canvas.height = h;
+    }
+
+    Override["javax/microedition/lcdui/ImageData.<init>.()V"] = function() {
+        this.context = document.createElement("canvas").getContext("2d");
```
marco-c: Is overriding the constructors of ImageData necessary? Couldn't we keep the old behavior of creating the canvas in the ImageDataFactory natives?
I mean, we're creating the canvas once anyway, why creating it in the ImageData constructor?

TimAbraldes: I had done a `grep -r "new ImageData"` to see where we're constructing `ImageData` objects. One of our tests - "tests/midlets/favouriteartists/Control.java" - constructs its own `ImageData` object so it seemed like it might be legitimate to try to construct an `ImageData` outside of the `ImageDataFactory`. If that test is invalid, I can move construction of `ImageData` objects into the `ImageDataFactory` methods.

marco-c: The ImageData class that the FavouriteArtists MIDlet is using is not javax/microedition/lcdui/ImageData but a custom class that they've written (it's a private class defined in the Control.java file itself).

The ImageData constructor has package visibility, so it's only accessible to classes in the javax/microedition/lcdui package (and the only class that creates ImageData objects is ImageDataFactory).

TimAbraldes: Oops, I probably should have caught that! I'll update this patch

PR #1170, opened 2015-02-26
Fix isFinal for method infos.

brendandahl:


brendandahl:
Removes 78 uneeded unwinds.

jit/analyze.ts
```diff
@@ -90,7 +90,7 @@ module J2ME {
       return true;
     }
     // XXX Determine whether we can start using the code in this function.
-    return false;
+    return methodInfo.isFinal;
     var result = methodInfo.isFinal;
```
marco-c: Could we now use the code in this function? Maybe it wasn't working before a1fd332 and it's working now?

brendandahl: Not yet, that code is only valid if every class in every jar is loaded. However, mbx and I have been chatting on enabling better analysis.

marco-c: Ok, could you, just for clarity, comment it out?

marco-c:
Looks good to me, just needs to be updated.

jit/analyze.ts
```diff
@@ -89,26 +89,26 @@ module J2ME {
...
-              methodInfo.signature === subClassMethodInfo.signature) {
-            result = false;
-            break;
-          }
-        }
-      }
-    }
-    return result;
+    return methodInfo.isFinal;
+    // XXX The following can only be used if every class in all jars is loaded.
```
marco-c: Nit: An enclosing /* */ would have worked too :smile:

Issue #1150, opened 2015-02-23
Optimize ByteArrayInputStream and ByteArrayOutputStream (regression from #996)


marco-c:
Up to a size of 128 (128 not included), the self-hosted versions are on par with the old native versions.

BEFORE
writeSingle: 439 - 306 - 327
readSingle: 179 - 156 - 153
writeArray64: 1485 - 1995 - 2307
readArray64: 1351 - 1269 - 1282
writeArray128: 1390 - 1406 - 1500
readArray128: 826 - 1019 - 896
writeArray256: 1246 - 1230 - 1212
readArray256: 670 - 710 - 701
writeArray: 2996 - 3481 - 3395
readArray: 2731 - 3024 - 3137

AFTER
writeSingle: 462 - 301 - 275
readSingle: 172 - 163 - 157
writeArray64: 1904 - 2389 - 1964
readArray64: 1213 - 1366 - 1129
writeArray128: 2010 - 2098 - 2008
readArray128: 1110 - 1285 - 1059
writeArray256: 1802 - 1955 - 1865
readArray256: 1200 - 1049 - 1153
writeArray: 4602 - 5045 - 4762
readArray: 4468 - 4733 - 4550

(This is a regression from #996.)

marco-c:
We've reverted #996.

PR #1171, opened 2015-02-26
revert #996 (fixes #1149 and #1150)

### LINKED ISSUE #1149: Revert changes to UTF_8_Writer and UTF_8_Reader from #996
On desktop performance was almost the same, on mobile the self-hosted implementations are definitely slower.

BEFORE
write: 1272 - 1731 - 1502
read: 344 - 110 - 110

AFTER
write: 6867 - 9906 - 6765
read: 714 - 525 - 507

(This is a regression from #996.)

mykmelez:
@marco-c, @mbebenita: This reverts #996 to fix #1149 and #1150.


mykmelez:
Since #1172 was reverted for causing bustage, we should instead take this complete reversion of #996 to resolve both perf regressions. Then we can reland the changes in #996 in incremental chunks that don't cause such regressions.

PR #1172, opened 2015-02-26
revert UTF_8_* changes from #996 (fixes #1149)

### LINKED ISSUE #1149: Revert changes to UTF_8_Writer and UTF_8_Reader from #996

mykmelez:
@marco-c, @mbebenita: This reverts just the UTF_8_Writer and UTF_8_Reader changes from #996 to fix #1149.  It doesn't work yet; unsure why.


mykmelez:
Erm, or I guess it does work. Tests were failing locally, but I suppose that could be unrelated.

PR #1178, opened 2015-02-27
Fix emoji bug

marco-c:
Turns out some MIDlets actually use emoji before loading their custom emoji sheets, so we need to load them earlier.
We can avoid loading them at startup though, and load them when the MIDlet gains the foreground (since MIDlets aren't going to draw anything before).


libs/emoji.js
```diff
@@ -1179,33 +1132,29 @@ var emoji = (function() {
...
+            continue;
+          }
 
-            promises.push(new Promise(function(resolve, reject) {
-              images[i].onload = resolve;
-            }));
+          images[i] = new Image();
+          var num = i.toString(16);
+          if (config.customEmojiImageFormat) {
+            var fileName = config.customEmojiImageFormat.replace("NUM", num);
```
mykmelez: We should document *config.customEmojiImageFormat* in config/default.js, even if we don't define it there.

mykmelez: > We should document config.customEmojiImageFormat in config/default.js, even if we don't define it there.

This is more of a general nit: we're defining config.* attributes in a variety of places, and we should document the set being used. But I'm going to merge this and then tackle that in a followup.

PR #1183, opened 2015-02-28
Add baseline jit for new multi array.

brendandahl:
Sitting around in stash for awhile...Now all methods in classes.jar can compile (one whole more method!) and one more method in a midlet.


marco-c:
Looks good to me. We already have some tests for multianewarray (although probably we're not testing them with the baseline JIT).

brendandahl:
> (although probably we're not testing them with the baseline JIT

Yeah, for a quick test I added a loop in ArrayTest.java to make sure it worked in the jit.

PR #1188, opened 2015-02-28
Lazy Class Parsing

mbebenita:
This PR changes class loading to be lazy and memory efficient. There will be follow up PRs to make things even more lazy.


mbebenita:
After repeatedly running the same midlet, on master and this branch I see the following numbers in chrome:

Before:

screen shot 2015-03-05 at 12 40 55 am

After:

screen shot 2015-03-05 at 12 40 14 am

I can't really explain the large reduction in code size.

mbebenita:
As a side note, we need to look into reducing the number of allocated JS strings. Here are the counts of the total lengths of strings in some midlets. 

```
296862: ClassInfo::implKey 
119463: ClassInfo::signature 
45850: ClassInfo::name 
31052: FieldInfo::signature 
28721: FieldInfo::name 
```

This is a lot and we could solve this by not reflecting Java strings in JS. A quick analysis of several class files indicates that constant pools have no string duplicates. This means we can use constant pool name indices to look up methods instead of strings. We do however need to figure out what the constant pool index is for a given lookup string. Instead of:

```
indexOfMethod(name: string, signature: string, isStatic: boolean): number {
      var methods = this.methods;
      if (!methods) {
        return -1;
      }
      for (var i = 0; i < methods.length; i++) {
        var method = this.getMethodByIndex(i);
        if (method.name === name && method.signature === signature && method.isStatic === isStatic) {
          return i;
        }
      }
      return -1;
    }
``` 
we could use:

```
indexOfMethod(name: string, signature: string, isStatic: boolean): number {
      var nameIndex = this.indexOfString(name); // Lookup the name index first.
      var methods = this.methods;
      if (!methods) {
        return -1;
      }
      for (var i = 0; i < methods.length; i++) {
        var method = this.getMethodByIndex(i); // We wouldn't need to parse the method info here since we can get the index from an offset.
        if (method.nameIndex === nameIndex) {
          return i;
        }
      }
      return -1;
    }
```

marco-c:
TestThreadJoin is failing because we've removed the synthetic code in Thread.start0.()V and we're no longer calling notifyAll at the end of the execution of a thread.

marco-c:
About using indices instead of strings to lookup methods, we would need to use both the name index and the descriptor index (because methods are uniquely identified by name and descriptor).

So, something like:

```
indexOfMethod(name: string, signature: string, isStatic: boolean): number {
      var nameIndex = this.indexOfString(name); // Lookup the name index first.
      var descriptorIndex = this.indexOfString(signature);
      var methods = this.methods;
      if (!methods) {
        return -1;
      }
      for (var i = 0; i < methods.length; i++) {
        var method = this.getMethodByIndex(i); // We wouldn't need to parse the method info here since we can get the index from an offset.
        if (method.nameIndex === nameIndex && method.descriptorIndex == descriptorIndex) {
          return i;
        }
      }
      return -1;
    }
```

mbebenita:
Yes, although we wouldn't have to parse the method info fully to check if the method name index and method signature index matches.

jit/analyze.ts
```diff
@@ -65,6 +65,7 @@ module J2ME {
     "java/lang/Thread.yield.()V": YieldReason.Root,
     "java/lang/Thread.start0.()V": YieldReason.Root,
     "java/lang/Class.forName0.(Ljava/lang/String;)V": YieldReason.Root,
+    "java/lang/Class.newInstance1.(Ljava/lang/Object;)V": YieldReason.Root,
```
mykmelez: We should also remove "java/lang/Class.newInstance.()Ljava/lang/Object;" from this list, right?

parser.ts
```diff
@@ -0,0 +1,1102 @@
...
+    CONSTANT_String = 8,
+    CONSTANT_Integer = 3,
+    CONSTANT_Float = 4,
+    CONSTANT_Long = 5,
+    CONSTANT_Double = 6,
+    CONSTANT_NameAndType = 12,
+    CONSTANT_Utf8 = 1,
+    CONSTANT_Unicode = 2,
+    CONSTANT_Any = 13 // NON-STANDARD
+  }
```
mykmelez: I extracted this structure from native.js and moved it into vm/tags.js because it originally came from a node-jvm file, so it should be attributed to that project, per @gerv, and the vm/ subdirectory contains the correct LICENSE file for it.

If you want to move it into this file, you should also copy the copyright notice from vm/tags.js and the license from vm/LICENSE into this file, put it into a comment above the structure, and then put an "end of imported section" comment below the structure, like we did for the *hashBytesTo32BitsMurmur* function in utilities.ts.

parser.ts
```diff
@@ -0,0 +1,1102 @@
...
+    static Z: PrimitiveArrayClassInfo;
+    static C: PrimitiveArrayClassInfo;
+    static F: PrimitiveArrayClassInfo;
+    static D: PrimitiveArrayClassInfo;
+    static B: PrimitiveArrayClassInfo;
+    static S: PrimitiveArrayClassInfo;
+    static I: PrimitiveArrayClassInfo;
+    static J: PrimitiveArrayClassInfo;
+  }
+}
```
mykmelez: Nit: add newline to last line of file.

vm/classRegistry.ts
```diff
@@ -149,10 +156,12 @@ module J2ME {
...
-      var classes = classInfo.classes;
-      classes.forEach(function (c, n) {
-        classes[n] = self.loadClass(c);
-      });
+      //var classes = classInfo.getClasses();
+      // FIXME
+      //classes.forEach(function (c, n) {
+      //  debugger;
+      //  classes[n] = self.loadClass(c.className);
+      //});
```
mykmelez: What's broken about it?

parser.ts
mykmelez: Actually, it looks like a bunch of other code in parser.ts comes from files that originally came from node-jvm, so you should instead move this entire file into vm/ and put the node-jvm copyright header at the top of it.

mbebenita: The enum declarations look just like they appear in the JVM spec, I don't see how this can be attributed to any one project.

What other code are you referring to in parser.ts? We pretty much rewrote the entire file from scratch.

mykmelez: > The enum declarations look just like they appear in the JVM spec, I don't see how this can be attributed to any one project.

I might be missing something, but the only representation of the constant pool tags in the JVM is [this non-code table](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.4-140). There's no such enum declaration (not even in a different language).

Nevertheless, I've wondered whether it's necessary to attribute code that we would rewrite in exactly the same fashion, if we hadn't copied it (or converted it from another language). In https://github.com/mozilla/j2me.js/pull/1097, @gerv said not to worry about the *table* list in tests/opcodeCoverage.py, which appeared to have come from opcodes.js (which doesn't exist anymore).

So if he says the same about this structure, then I'm fine with not attributing it. Otherwise, we should err on the side of caution, because attributing the code now doesn't harm us, and fixing it later is a pain in the ass (witness https://github.com/mozilla/j2me.js/pull/883, https://github.com/mozilla/j2me.js/pull/894, and https://github.com/mozilla/j2me.js/pull/1097).


> What other code are you referring to in parser.ts? We pretty much rewrote the entire file from scratch.

I haven't audited all the code, but parts of it clearly come from vm/reader.ts, which has been removed. That code may or may not have originally come from node-jvm, but I think they're now licensed under its license, since they were added to a file that was copied from that project.

If you want to use a different license for parser.ts (which can only be GPL, then, or another permissive license that is compatible with it), then you could carefully split out the new code from the existing code, putting them into separate files. But I don't see why that's worth the effort. It would be much simpler to just maintain the license for the code you're copying, licensing the new code under the same license.

mbebenita:
Travis fails with: 
```
java.lang.TestThreadJoin-0: fail
```
This started happening after the last merge.

mykmelez:
> This started happening after the last merge.

@marco-c has PR https://github.com/mbebenita/j2me.js/pull/22 against your branch to fix it.

gerv:
Hey: is there a specific question for me here? :-)

Gerv

mykmelez:
> Hey: is there a specific question for me here? :-)

Erm, no, I don't think so.

We copied [this JavaScript object](https://github.com/mozilla/j2me.js/blob/3ad39aaffeb2869b99ea00dcaa520b4dd10f395a/vm/tags.js#L11-L24) from node-jvm, then converted into [this TypeScript enum](https://github.com/mbebenita/j2me.js/blob/parse/parser.ts#L189-L203), and there was some question about whether or not to retain the attribution and license, since it's a representation of [this table of constant pool tags](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.4-140), and we'd implement it identically if we were doing so from scratch.

But it turned out that we copied other code with the same attribution/license into the same file, so that file needs the original attribution/license anyway.

brendandahl:
Comparing master with master+this branch:.

On desktop:
```
Current mean : 996ms
Baseline mean: 950ms
+/-          : 46ms
%            : 4.88
SLOWER
```
On a flame:
```
Current mean : 17435ms
Baseline mean: 17740ms
+/-          : -305ms
%            : -1.72
INSIGNIFICANT RESULT
```

PR #1196, opened 2015-03-04
Don't allow thread preemption when resuming a thread.

yxl:
Current thread scheduler allows thread preemption when resuming an old thread from the current thread. This preemption behavior doesn't conform with the mainstream JVM at least in the following two cases:
1. When a new higher priority thread starts, it will preempt the current thread unconditionally.
2. When the current thread leaves a `synchronized` block and wakes up a higher priority thread being blocked, the higher priority thread will preempt the current thread.

The following code illustrates the above two cases.

```Java
import java.lang.*;

/* Name of the class has to be "Main" only if the class is public. */
class TestThreadPriority
{
	private static String result = "Main";
	
	static class Prioritized extends Thread {
	    public Prioritized() {
	        setPriority(Thread.MAX_PRIORITY);
	    }
	
	    public void run() {
	        result = "Prioritized";
	    }
	}
	
	static class Synchronized extends Thread {
	    public Synchronized() {
	        setPriority(Thread.MAX_PRIORITY);
	    }
	
	    public void run() {
	        synchronized(TestThreadPriority.class) {
	            result = "Synchronized";
	        }
	    }
	}
	    
	public static void main (String[] args) throws java.lang.Exception
	{
	    Thread p = new Prioritized();
	    p.start();
	    // With our implementation, p will preempt the main thread to set `result` to be "Prioritized".
	    // While the mainstream implementation, s won't and keep `result` to be "Main"
	    System.out.println(result);
	
	    try {
	        Thread.sleep(10);
	    }
	    // Wait p to set `result` to be "Prioritized".
	    System.out.println(result);
	
	    synchronized(TestThreadPriority.class) {
	        Thread s = new Synchronized();
	        s.start();
	        try {
	            Thread.sleep(10);
	        }
	    }
	    // With out implementation, s will preempt the main thread to set `result` to be "Synchronized".
	    // While with the mainstream implementation, s won't and keep `result` to be "Prioritized".
	    System.out.println(result);
	}
}
```

Our implementation is supposed to output the result of 
```
Prioritized
Prioritized
Synchronized
```
But there is a bug and we get 
```
Prioritized
Prioritized
Prioritized
```

The [mainstream implementation](http://ideone.com/tuOM1I) outputs
```
Main
Prioritized
Prioritized
```

This pull request drops the support of the unconditional thread preemption when resuming an old thread and makes our VM behaves the same as the mainstream one with the above test code.

### DIFF
--- a/vm/runtime.ts
+++ b/vm/runtime.ts
@@ -682,19 +682,11 @@ module J2ME {
       }
     }
 
-    /*
-     * @param jump If true, move the context to the first of others who have the
-     * same priority.
-     */
-    enqueue(ctx: Context, jump: boolean) {
+    enqueue(ctx: Context) {
       var priority = ctx.getPriority();
       release || assert(priority >= MIN_PRIORITY && priority <= MAX_PRIORITY,
                         "Invalid priority: " + priority);
-      if (jump) {
-        this._queues[priority].unshift(ctx);
-      } else {
-        this._queues[priority].push(ctx);
-      }
+      this._queues[priority].push(ctx);
       this._top = Math.max(priority, this._top);
     }
 
@@ -735,14 +727,7 @@ module J2ME {
      *      higher priority thread is scheduled to run.
      */
     static scheduleRunningContext(ctx: Context) {
-      // Preempt current thread if the new thread has higher priority
-      if ($ && ctx.getPriority() > $.ctx.getPriority()) {
-        Runtime._runningQueue.enqueue($.ctx, true);
-        Runtime._runningQueue.enqueue(ctx, false);
-        $.pause("preempt");
-      } else {
-        Runtime._runningQueue.enqueue(ctx, false);
-      }
+      Runtime._runningQueue.enqueue(ctx);
       Runtime.processRunningQueue();
     }
 
--- a/vm/runtime.ts
+++ b/vm/runtime.ts
@@ -715,16 +715,13 @@ module J2ME {
 
 
     /*
-     * The thread scheduler uses green thread algorithm, which a preemptive,
+     * The thread scheduler uses green thread algorithm, which a non-preemptive,
      * priority based algorithm.
      * All Java threads have a priority and the thread with he highest priority
      * is scheduled to run.
      * In case two threads have the same priority a FIFO ordering is followed.
-     * A different thread is invoked to run only if
-     *   1. The current thread blocks or terminates.
-     *   2. A thread with a higher priority than the current thread enters the
-     *      Runnable state. The lower priority thread is preempted and the
-     *      higher priority thread is scheduled to run.
+     * A different thread is invoked to run only if the current thread blocks or
+     * terminates.
      */
     static scheduleRunningContext(ctx: Context) {
       Runtime._runningQueue.enqueue(ctx);
--- a/tests/java/lang/TestThreadPriority.java
+++ b/tests/java/lang/TestThreadPriority.java
@@ -4,7 +4,7 @@ import gnu.testlet.Testlet;
 import gnu.testlet.TestHarness;
 
 public class TestThreadPriority implements Testlet {
-    public int getExpectedPass() { return 2; }
+    public int getExpectedPass() { return 1; }
     public int getExpectedFail() { return 0; }
     public int getExpectedKnownFail() { return 0; }
     private static String result = "";
@@ -24,10 +24,6 @@ public class TestThreadPriority implements Testlet {
         for (int i = 0; i < priorities.length; i++) {
             new Prioritized(priorities[i]).start();
         }
-
-        // The priority of the main thread is 5. Threads with higher priorities
-        // should preempt the main thread to run first.
-        th.check(result, "9 7 ");
 
         String expected = "9 7 5 1 ";
         try {


vm/runtime.ts
```diff
@@ -706,14 +698,7 @@ module J2ME {
      *      higher priority thread is scheduled to run.
      */
     static scheduleRunningContext(ctx: Context) {
```
mykmelez: Nit: the comment on this function still says a thread with a higher priority preempts the current thread, which is no longer the case, so that comment should be updated.

mykmelez:
Other than the outdated comment, this looks good and seems like the right fix for correctness.

yxl:
@mykmelez Thanks. I rebased the patch and updated the comment with b428506

mykmelez:
To ensure we're emulating behavior on the target device as closely as possible, I asked @marco-c to test an example midlet on his device. I'm waiting to hear back from him before merging this.

marco-c:
Running yxl's code I'm getting the same result (Main, Prioritized, Prioritized) both on my Linux machine and on the Nokia Asha simulator (unfortunately I'm having problems installing MIDlets on my real Nokia Asha phone).

I've written another test:
```Java
import java.lang.*;

class TestThreadPriority
{
	private static String result = "Main";
 
	static class NonPrioritized extends Thread {
	    public NonPrioritized() {
	        setPriority(Thread.MIN_PRIORITY);
 
	        Thread p = new Prioritized();
		p.start();
		    // With our implementation, p will preempt the main thread to set `result` to be "Prioritized".
		    // While the mainstream implementation, s won't and keep `result` to be "Main"
		    System.out.println(result);
 
		    try {
		        Thread.sleep(10);
		    }
		    // Wait p to set `result` to be "Prioritized".
		    System.out.println(result);
 
		    synchronized(TestThreadPriority.class) {
		        Thread s = new Synchronized();
		        s.start();
		        try {
		            Thread.sleep(10);
		        }
		    }
		    // With out implementation, s will preempt the main thread to set `result` to be "Synchronized".
		    // While with the mainstream implementation, s won't and keep `result` to be "Prioritized".
		    System.out.println(result);
	    }
 
	    public void run() {
	        result = "Prioritized";
	    }
	}
 
	static class Prioritized extends Thread {
	    public Prioritized() {
	        setPriority(Thread.MAX_PRIORITY);
	    }
 
	    public void run() {
	        result = "Prioritized";
		System.out.println("RUNNING");
	    }
	}
 
	static class Synchronized extends Thread {
	    public Synchronized() {
	        setPriority(Thread.MAX_PRIORITY);
	    }
 
	    public void run() {
	        synchronized(TestThreadPriority.class) {
	            result = "Synchronized";
                    System.out.println("RUNNING SYNCHRONIZED");
	        }
	    }
	}
 
	public static void main (String[] args) throws java.lang.Exception
	{
	    Thread p = new NonPrioritized();
	    p.start();
	}
}
```

The output alternates between:
```
Main
RUNNING
Prioritized
Prioritized
RUNNING SYNCHRONIZED
```

and (most often):
```
RUNNING
Main
Prioritized
Prioritized
RUNNING SYNCHRONIZED
```

So looks like the Thread with the highest priority is sometimes scheduled to run before the Thread with the lowest priority, sometimes not.
The *result* value is always *Main* perhaps because of a synchronization problem, and not because of the scheduling order of the threads.

marco-c:
Running yxl's code I'm always seeing the same result (Main, Prioritized, Prioritized) both on my desktop machine and on the Nokia Asha simulator (unfortunately I'm unable to install MIDlets on my real Nokia Asha phone at the moment).

I've written another test:
```Java
import java.lang.*;

class TestThreadPriority
{
	private static String result = "Main";
 
	static class NonPrioritized extends Thread {
	    public NonPrioritized() {
	        setPriority(Thread.MIN_PRIORITY);
	    }
 
	    public void run() {
	        Thread p = new Prioritized();
		p.start();
		// With our implementation, p will preempt the main thread to set `result` to be "Prioritized".
		// While the mainstream implementation, s won't and keep `result` to be "Main"
		System.out.println(result);
 
		try {
		    Thread.sleep(10);
		}
		// Wait p to set `result` to be "Prioritized".
		System.out.println(result);
 
		synchronized(TestThreadPriority.class) {
		    Thread s = new Synchronized();
		    s.start();
		    try {
		        Thread.sleep(10);
		    }
		}
		// With out implementation, s will preempt the main thread to set `result` to be "Synchronized".
		// While with the mainstream implementation, s won't and keep `result` to be "Prioritized".
		System.out.println(result);
	    }
	}
 
	static class Prioritized extends Thread {
	    public Prioritized() {
	        setPriority(Thread.MAX_PRIORITY);
	    }
 
	    public void run() {
	        result = "Prioritized";
		System.out.println("RUNNING");
	    }
	}
 
	static class Synchronized extends Thread {
	    public Synchronized() {
	        setPriority(Thread.MAX_PRIORITY);
	    }
 
	    public void run() {
	        synchronized(TestThreadPriority.class) {
	            result = "Synchronized";
                    System.out.println("RUNNING SYNCHRONIZED");
	        }
	    }
	}
 
	public static void main (String[] args) throws java.lang.Exception
	{
	    Thread p = new NonPrioritized();
	    p.start();
	}
}
```

The output alternates between:
```
Main
RUNNING
Prioritized
Prioritized
RUNNING SYNCHRONIZED
```
and (most often):
```
RUNNING
Main
Prioritized
Prioritized
RUNNING SYNCHRONIZED
```

So, looks like the Thread with the highest priority is sometimes scheduled to run before the Thread with the lowest priority, sometimes after.
Perhaps the output is "Main" because of a synchronization problem, and not because of the scheduling order.

marco-c:
This is the output of our VM:
```
RUNNING
Prioritized
Prioritized
RUNNING SYNCHRONIZED
Synchronized
```

yxl:
@marco-c I think the reason you get random results with your test is because the desktop Java VM uses native threads which are scheduled by a time sharing algorithm. Every thread runs in turns with a certain slice of time and that slice of time is not accurately fixed.

I tested your test with this pull request and the result is:
```
Main
RUNNING
Prioritized
Prioritized
RUNNING SYNCHRONIZED
```

The same with the first case of your desktop machine and device.

yxl:
@marco-c I rewrote you test as following to test on real Asha device:
```Java
package com.nokia.example;

import javax.microedition.lcdui.Display;
import javax.microedition.lcdui.Displayable;
import javax.microedition.lcdui.Form;
import javax.microedition.lcdui.StringItem;
import javax.microedition.midlet.MIDlet;

public class HelloWorldMIDlet extends MIDlet {

    public HelloWorldMIDlet() {
    }

    public void startApp() {
        Displayable current = Display.getDisplay(this).getCurrent();
        if (current == null) {
            HelloScreen helloScreen = new HelloScreen(this);
            Display.getDisplay(this).setCurrent(helloScreen);
        }
    }

    public void pauseApp() {
    }

    public void destroyApp(boolean unconditional) {
    }
}

class HelloScreen extends Form {

	static HelloScreen s;

	void println(String string) {
		StringItem helloText = new StringItem("", string);
		append(helloText);
	}

	static class TestThreadPriority {
		private static String result = "Main";

		static class NonPrioritized extends Thread {
			public NonPrioritized() {
				setPriority(Thread.MIN_PRIORITY);
			}

			public void run() {
				Thread p = new Prioritized();
				p.start();
				// With our implementation, p will preempt the main thread to
				// set `result` to be "Prioritized".
				// While the mainstream implementation, s won't and keep
				// `result` to be "Main"
				s.println(result);

				try {
					Thread.sleep(10);
				}
				// Wait p to set `result` to be "Prioritized".
				s.println(result);

				synchronized (TestThreadPriority.class) {
					Thread s = new Synchronized();
					s.start();
					try {
						Thread.sleep(10);
					}
				}
				// With out implementation, s will preempt the main thread to
				// set `result` to be "Synchronized".
				// While with the mainstream implementation, s won't and keep
				// `result` to be "Prioritized".
				s.println(result);
			}
		}

		static class Prioritized extends Thread {
			public Prioritized() {
				setPriority(Thread.MAX_PRIORITY);
			}

			public void run() {
				result = "Prioritized";
				s.println("RUNNING");
			}
		}

		static class Synchronized extends Thread {
			public Synchronized() {
				setPriority(Thread.MAX_PRIORITY);
			}

			public void run() {
				synchronized (TestThreadPriority.class) {
					result = "Synchronized";
					s.println("RUNNING SYNCHRONIZED");
				}
			}
		}

		public static void main() {
			Thread p = new NonPrioritized();
			p.start();
		}
	}

	public HelloScreen(HelloWorldMIDlet midlet) {

		super("HelloWorld"); // added for 3.0

		s = this;
		TestThreadPriority.main();
	}
}
```

The result is interesting:
```
Main
Main
RUNNING
Prioritized
RUNNING SYNCHRONIZED
```

It shows that the Prioritized thread event doesn't have enough time to run. This turns out that the starting a new thread is not preemptive. If we increase the time of the first `sleep` to 100ms, we get this same result as this pull request's VM:
```
Main
RUNNING
Prioritized
Prioritized
RUNNING SYNCHRONIZED
```

In conclusion, this new thread scheduler is much more close to the Asha's :-)

marco-c:
I think the bug you're trying to fix isn't directly due to the scheduling order, but to some other bug that is triggered by it.
The scheduling order shouldn't matter for correctness, as long as we're not violating the Java memory model.

The results of these tests are likely going to be different, since there's no synchronization when accessing the *result* variable. JVM implementations might even reorder instructions in these cases, for example:
```
RUNNING
Main
Prioritized
Prioritized
RUNNING SYNCHRONIZED
```

Shows that one of the threads is running and setting *result* to *Prioritized* (because it prints "RUNNING" only after setting result to *Prioritized*), but the other thread isn't seeing this change and still has *result = Main*.
From http://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.4.5, "Without correct synchronization, very strange, confusing and counterintuitive behaviors are possible."

Since removing preemption is fixing a bug, I'd take the patch and reintroduce preemption when we'll be able to fix the underlying bug, but I wouldn't add the test enforcing a particular scheduling order.

yxl:
@marco-c The test enforcing scheduling order is removed in c77a23a8ff904

PR #1198, opened 2015-03-04
Gfx fixes

TimAbraldes:
Revert to using the PhoneME implementation of the `Graphics` object.
Obsoletes #1104 because we now use the reference implementation so
it no longer makes sense to test our conformance to the behavior of
the reference implementation. I found that the PhoneME implementation
was visually faster on the MIDlets I tested than the native
implementation.

Maintain information about what display is currently in the foreground,
and only allow drawing operations that are intended to go to the current
foreground display. This is part of the fix for #823.

Create a `ContextInfo` object that represents the translation, clipping,
font, and color information of a canvas context. Whenever a canvas
context gets modified, update its `ContextInfo`. Before any
drawing operation, check the `ContextInfo` and apply any values
from the `Graphics` object that don't match. This is the other part of
the fix for #823.


TimAbraldes:
This is the result of running the startup benchmark on my favorite MIDlet (20 runs baseline, 20 runs benchmark) on my laptop:

|      Test     | Baseline Mean |    Mean   |    +/-   |   %   |       P       |    Min    |    Max    | 
|:--------------|--------------:|----------:|---------:|------:|--------------:|----------:|----------:|
| startupTime   |       2,742ms |   2,896ms |    154ms |  5.62 | INSIGNIFICANT |   2,316ms |   3,941ms | 
| totalSize     |     126,343kb | 125,924kb |   -419kb | -0.33 | INSIGNIFICANT | 125,174kb | 127,814kb | 
| domSize       |         168kb |     169kb |      1kb |  0.54 | INSIGNIFICANT |     167kb |     173kb | 
| styleSize     |         441kb |     458kb |     17kb |  3.86 |         WORSE |     451kb |     480kb | 
| jsObjectsSize |      63,079kb |  63,662kb |    583kb |  0.92 |         WORSE |  63,357kb |  64,081kb | 
| jsStringsSize |       2,928kb |   2,931kb |      3kb |  0.10 | INSIGNIFICANT |   2,923kb |   2,952kb | 
| jsOtherSize   |      59,305kb |  58,230kb | -1,075kb | -1.81 |        BETTER |  57,631kb |  60,469kb | 
| otherSize     |         422kb |     474kb |     52kb | 12.44 | INSIGNIFICANT |     419kb |     690kb |

TimAbraldes:
Withdrawing this PR in favor of #1277

PR #1199, opened 2015-03-05
WIP: lazily link methods

mykmelez:
Here's a work-in-progress of a branch that links methods lazily. There's still some cleanup and more testing to do, but it passes automated tests as well as manual tests with a midlet. Unsure about the perf/memory impact.


mykmelez:
This changeset does a few different things:

1. Consolidates method linking into a global singleton Methods object to which all methods get linked by their implKeys. The Methods prototype is a Proxy that links methods lazily when their implKeys are dereferenced. It uses implKey (instead of a mangled class/method name) to extract its className part and use it to retrieve a Klass from the global object (and get classInfo and methodInfo through that).

(implKey makes those keys long, though; we should make them shorter.)

(On Chrome, which doesn't support Proxy, nor polyfills thereof, we still link methods eagerly.)

2. "Links" non-static methods into Klass.prototype objects by creating an "implKeyFor_[mangledName]" property whose value is the implKey, which callers then use to retrieve the method lazily from Methods. This avoids having to eagerly link the function itself (or a memoizer) at load time. (We still eagerly link a memoizer for bound methods, which are currently rare; we should fix that eventually.)

(I tried adding a proxy to the Klass prototype chain to trap and link the method, but it didn't seem to be trapping all the necessary accesses; we should investigate further.)

3. On ClassInfo.scanMethods, creates a MethodInfo.methodIDs array that stores method name, signature, isStatic, and isAbstract values. That's the minimum information we need at load time to prepare for lazy linking.

There's more we can do here, like reducing the size of the method calls at the call sites, but the changeset is complete, and it passes automatic and manual tests on desktop and device. Unfortunately, I don't see any impact on startup perf nor memory usage in my tests of a midlet. It might be a little faster on startup, but the results are inconclusive (although the numbers are noisy). Memory usage seems about the same.

mbebenita:
PR #1225 changes the name mangling scheme so that all  virtual function calls use the vTable index. The code looks something like:

```
o.v0();
o.v1();
o.v2();
```

Instead of using a proxy, which is likely to be very bad for performance, we could install traps in a root prototype object that all Java objects share, for each of the `v0` ... `vn` properties. Each trap would know it's own index in the vtable and can patch up the direct prototype of the object with the lazily linked method.

mykmelez:
Bumping this back to WIP while I figure out why it isn't having the expected impact on startup. Some preliminary investigations show that it correctly links just a subset of methods, but the same number of MethodInfo objects are created regardless, so some other code is creating them anyway.

vm/parser.ts
```diff
@@ -759,8 +764,17 @@ module J2ME {
       var s = this;
       var methods_count = s.readU2();
       var m = this.methods = new Array(methods_count);
+      var ids = this.methodIDs = new Array(methods_count);
       for (var i = 0; i < methods_count; i++) {
         m[i] = s.offset;
+
+        ids[i] = [
```
marco-c: We should try to avoid this, because it reduces the effectiveness of lazily loading classes and also uses more memory (because we're storing the method name and signature both here and in MethodInfo)

mykmelez: > We should try to avoid this, because it reduces the effectiveness of lazily loading classes

As long as we need that info to minimally link the methods, it doesn't actually reduce the effectiveness of lazily loading the classes, because we'll load the info at link time anyway.

> and also uses more memory (because we're storing the method name and signature both here and in MethodInfo)

We can fix that by deleting the structure after minimally linking the methods.

marco-c: > As long as we need that info to minimally link the methods, it doesn't actually reduce the effectiveness of lazily loading the classes, because we'll load the info at link time anyway.

Yes, but with this change we''re going to load eagerly all the methods of a class, even if we need only a few.

mykmelez: > Yes, but with this change we''re going to load eagerly all the methods of a class, even if we need only a few.
> (There are classes that contain a lot of methods that are never used)

Erm, no, with this change we're going to load eagerly only a subset of data about all the methods of a class: name, signature, and flags. And we have to load this information at startup in order to minimally link virtual methods on Klass prototypes at startup.

We don't have to do it this way, of course. We could instead make MethodInfo lighter weight and create lightweight MethodInfo objects for every method in every class we load at startup, as #1225 does. Or this branch could add a *MethodInfo.scanMethods2* method that *linkKlassMethods* calls when it's time to minimally link the methods.

But then we'd be scanning methods twice on startup: once to retrieve just their indexes, and once to retrieve their name/signature/flags.

(Ideally, we wouldn't have to minimally link methods on startup at all. And we may be able to avoid it. But neither this branch nor that one solves that problem yet. And in any case it's a minor optimization compared to the significant wins we can gain by loading only the minimum information about each method on startup.)

marco-c: > Erm, no, with this change we're going to load eagerly only a subset of data about all the methods of a class: name, signature, and flags. And we have to load this information at startup in order to minimally link virtual methods on Klass prototypes at startup.

We're not loading everything about a method, but IIRC decoding the strings was one of the slowest things in method loading (I'm not sure I recall correctly, @mbebenita did some measurements).
We do need to load this data anyway, you're right (I forgot we need to do it anyway even if we don't have any compiled methods), but I think it would be better to do it only once.

> We don't have to do it this way, of course. We could instead make MethodInfo lighter weight and create lightweight MethodInfo objects for every method in every class we load at startup, as #1225 does. Or this branch could add a MethodInfo.scanMethods2 method that linkKlassMethods calls when it's time to minimally link the methods.
> But then we'd be scanning methods twice on startup: once to retrieve just their indexes, and once to retrieve their name/signature/flags.

I suspect this would be better, as we would avoid loading this data about methods twice. We could scan methods to get their indexes, and then another method could load the name/signature/fields when needed. Or, we could directly load them in MethodInfo in scanMethods, but I'd avoid loading the data twice.

mykmelez:
This was superseded by other work to make linking lazier. There may be more we can do, and some value to these changes, but that'll happen in separate branches, so I'm closing this request.

PR #1202, opened 2015-03-06
When a thread dies, wake up other waiting threads.

yxl:
`Thead.join()` is used to wait for  another thread dies. It should return immediately after the target thread dies, but I found 1 second extra delay in some circumstance.

The cause of the delay is that the reference implementation uses `Object.wait()` to wait and check if the thread is alive:
```Java
    public final synchronized void join() throws InterruptedException {
        while (isAlive()) {
            wait(1000);
        }
    }
```

Before terminating the thread, the [reference implementation](https://java.net/projects/phoneme/sources/svn/content/components/cldc/branches/cldc-feature-mr4-active/src/vm/share/runtime/Thread.cpp?rev=20547) will call `Object.notifyAll()` to notify all waiting thread:
```C++
void Thread::finish() {
  ...
  // Signal waiters waiting on thread
  Scheduler::notify(&receiver, true, false JVM_NO_CHECK_AT_BOTTOM);
}
```

But our implementation doesn't do so. So if the thread is dead right after we call `Thread.join()`, we need to wait 1 second until timeout. That's why we has a delay sometimes.

I fix this issue by calling `notifyAll` when the thread gets killed.

### DIFF
--- a/native.js
+++ b/native.js
@@ -550,13 +550,28 @@ Native["java/lang/Thread.start0.()V"] = function() {
           { name_index: 9, signature_index: 10 },
           { bytes: "internalExit" },
           { bytes: "()V" },
+          { tag: TAGS.CONSTANT_Methodref, class_index: 12, name_and_type_index: 14 },
+          { tag: TAGS.CONSTANT_Class, name_index: 13 },
+          { bytes: "java/lang/Object" },
+          { name_index: 15, signature_index: 16 },
+          { bytes: "notifyAll" },
+          { bytes: "()V" },
         ]},
       }),
       code: new Uint8Array([
         0x2a,             // aload_0
         0x59,             // dup
-        0xb6, 0x00, 0x01, // invokespecial <idx=1>
-        0xb7, 0x00, 0x07, // invokespecial <idx=7>
+        0xb6, 0x00, 0x01, // invokevitual <idx=1> Thread.run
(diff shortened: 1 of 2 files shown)


native.js
```diff
@@ -550,13 +550,28 @@ Native["java/lang/Thread.start0.()V"] = function() {
...
-        0xb7, 0x00, 0x07, // invokespecial <idx=7>
+        0xb6, 0x00, 0x01, // invokevitual <idx=1> Thread.run
+        0xb7, 0x00, 0x07, // invokespecial <idx=7> Thread.internalExit
+        0x2a,             // aload_0
+        0x59,             // dup
+        0x4c,             // astore_1
+        0xc2,             // monitorenter
+        0x2a,             // aload_0
+        // Signal waiters waiting on thread
+        0xb7, 0x00, 0x0b, // invokespecial <idx=11> Object.notifyAll
```
marco-c: Instead of adding code to the synthetic method, we could call *$.ctx.notify(this, true);* in the *java/lang/Thread.internalExit.()V* native.

yxl: `notifyAll()` needs to be called within `synchronized` block or function, but the `synchronized` keyword doesn't take into effect for the native method of `internalExit()`.
I get `IllegalMonitorStateException` when calling `$.ctx.notify(this, true);` in the `internalExit()`.

marco-c: You could add monitorenter and monitorexit around the *Thread::internalExit* call. Anyway, it's not really important :smile:

PR #1203, opened 2015-03-06
A more compact encoding of unwinding code.

mbebenita:


mykmelez:
Looks good, and may improve perf, although it's unclear from my data thus far.

PR #1209, opened 2015-03-09
Print running test name

TimAbraldes:
I find myself adding this locally to be able to quickly determine which
test is timing out.

### DIFF
--- a/tests/RunTests.java
+++ b/tests/RunTests.java
@@ -82,6 +82,8 @@ public class RunTests extends MIDlet {
     void runTest(String name) {
         name = name.replace('/', '.');
 
+        System.out.println("Running " + name);
+
         Form form = new Form(name);
         Display display = Display.getDisplay(this);
         Harness harness = new Harness(name, display);
@@ -132,6 +134,8 @@ public class RunTests extends MIDlet {
         fail += harness.failed();
         knownFail += harness.knownFailed();
         unknownPass += harness.unknownPassed();
+
+        System.out.println("Completed " + name);
     }
 
     public void startApp() {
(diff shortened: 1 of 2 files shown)


tests/RunTests.java
```diff
@@ -132,6 +134,8 @@ void runTest(String name) {
         fail += harness.failed();
         knownFail += harness.knownFailed();
         unknownPass += harness.unknownPassed();
+
+        System.out.println("Completed " + name);
```
marco-c: Nit: this is redundant with harness.report()

tests/automation.js
```diff
@@ -265,6 +265,7 @@ casper.test.begin("unit tests", 20 + gfxTests.length, function(test) {
     // Graphics tests
 
     gfxTests.forEach(function(testCase) {
+        console.log("Running " + testCase.name);
```
marco-c: Nit: Logging for gfx tests shouldn't be needed:
```
[info] [phantom] Step anonymous 91/318 http://localhost:8000/main.html?fontSize=12&midletClassName=gfx/CanvasTest&jars=tests/tests.jar&logConsole=web,page&logLevel=log (HTTP 200)
...
PASS gfx/CanvasTest - Pass
```

mykmelez:
I've wanted something like this too. Just note that it pushes us very close to the 10,000 line limit on Travis, as the latest build on this branch generated 9963 lines of output. And once we breach the limit, Travis will start reporting that tests passed even when they didn't. We should keep an eye on that after landing this, and maybe follow up with a review of the output to see if there's something unuseful that we can pare.

brendandahl:
Getting rid of the verbose output for jar creation would get rid of a good number of lines.

PR #1211, opened 2015-03-09
Remove obsolete make clean step.

brendandahl:
We no longer have a j2me.js file in the root directory and as far as I can tell nothing generates files with a `~`.

make clean keeps failing for me since I use `~` to keep folders in git repos that are globally ignored.


yxl:
I guess files whose name ends with `~` are temporary file generated by some text editors, such as `gedit`.

PR #1213, opened 2015-03-10
Fix precedence when constants are negative.

mbebenita:


mbebenita:
This is intermittent, and I'm investigating.

PR #1214, opened 2015-03-10
Fix getting the static initializer.

brendandahl:
This was causing the canStaticInitiaizerYield function in analyze.ts to return incorrect results.

(we really need to run tests with everything compiled and compiler assertions on!)


mykmelez:
> (we really need to run tests with everything compiled and compiler assertions on!)

Can you file an issue for this?

PR #1215, opened 2015-03-10
Playing audio in parent window to gain the privilege to play AMR audio on Firefox OS device

yxl:
AMR audio could only be played with privileged app on Firefox OS device. In order to support this kind of audio, I have to move the audio playing code from main.html to index.html.

### DIFF
--- a/midp/media.js
+++ b/midp/media.js
@@ -204,8 +204,35 @@ Media.PlayerCache = {
 function AudioPlayer(playerContainer) {
     this.playerContainer = playerContainer;
 
+    this.messageHandlers = {
+        mediaTime: [],
+        duration: []
+    };
+
+    this.sender = DumbPipe.open("audioplayer", {}, function(message) {
+        switch (message.type) {
+            case "end":
+                MIDP.sendEndOfMediaEvent(this.playerContainer.pId, message.duration);
+                break;
+            case "mediaTime": // fall through
+            case "duration":
+                var f = this.messageHandlers[message.type].shift();
+                if (f) {
+                    f(message.data);
+                }
+                break;
+            default:
+                console.error("Unknown audioplayer message type: " + message.type)
+                break;
+        }
+    }.bind(this));
+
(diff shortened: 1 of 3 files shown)


midp/media.js
```diff
@@ -215,70 +242,69 @@ AudioPlayer.prototype.realize = function() {
...
 
 AudioPlayer.prototype.getMediaTime = function() {
-    return Math.round(this.audio.currentTime * 1000);
+    var p =  new Promise(function(resolve, reject) {
+        this.sender({ type: "getMediaTime" });
+        this.messageHandlers.mediaTime.push(function(data) {
+            resolve(data);
+        });
+    }.bind(this));
+    asyncImpl("I", p);
```
mykmelez: This change means that *nGetMediaTime* will sometimes be asynchronous, which isn't clear from reading that function, as it appears to return a value synchronously.

To clarify the behavior of these two functions, I'd make *AudioPlayer.getMediaTime* return its promise, and *nGetMediaTime* call *asyncImpl* if the return value of its call to *player.getMediaTime* is a promise, i.e. something like:

```JavaScript
AudioPlayer.prototype.getMediaTime = function() {
    return new Promise(function(resolve, reject) {
        this.sender({ type: "getMediaTime" });
        this.messageHandlers.mediaTime.push(function(data) {
            resolve(data);
        });
    }.bind(this));
};
…
Native["com/sun/mmedia/DirectPlayer.nGetMediaTime.(I)I"] = function(handle) {
    var player = Media.PlayerCache[handle];
    var mediaTime = player.getMediaTime();
    if (mediaTime instanceof Promise) {
        asyncImpl("I", mediaTime);
    } else {
        return mediaTime;
    }
};
```

(I'd do the same for *AudioPlayer.getDuration*/*nGetDuration*.)

yxl: @mykmelez Thanks, that makes code much clear and I fixed in  fae615b

Issue #1019, opened 2015-02-02
Update README with information on profiling


marco-c:
https://github.com/andreasgal/j2me.js/pull/1002#issuecomment-72495796

Issue #989, opened 2015-01-30
remove old instrumentation and profiling code


mykmelez:
The old instrumentation and profiling code in instrument.js (and elsewhere) is no longer hooked up to the new VM, so we should remove it from the codebase.

yxl:
@mykmelez BTW, what is the replacement of the old instrumentation to let me view the Java methods' calling costs and counts?

mykmelez:
cc: @mbebenita and @brendandahl for an explanation of how to profile method call costs/counts in the new VM.

mbebenita:
The VM has several profiling tools. The simplest feature is to use counters. For instance, `runtime.ts` defines several: `runtimeCounter`, `nativeCounter`, etc ... these should only ever be used in non release builds.

To use them, just sprinkle calls to `runtimeCounter.count("string")`. To see the accumulated counts, press the `Dump Counters` button, or `Clear Counters` if you want to reset them.

The second, more heavy weight profiling tool is to build with `PROFILE=1 make`. This will include Shumway's timeline viewer. One way to use it is to manually insert calls to `timeline.enter` / `timeline.leave` around code blocks that you're interested in measuring. 

If you want to record every method call, change the line in `runtime.ts` from:

```
if (false && methodTimeline) { 
```
to
```
if (false && methodTimeline) { 
```

This will wrap all methods with calls to `methodTimeline.enter` /  `methodTimeline.leave`. The resulting timeline is a very detailed flame chart of the execution. Because of the fine grained nature of this instrumentation, timing for very short lived events may not be recorded accurately.

You can get creative with the timelines. The API looks something like this:

```
timeline.enter(name: string, details?: Object);
timeline.leave(name?: string, details?: Object);
```

You must pair the `enter` and `leave` events but you don't necessarily need to specify arguments for `name` and `details`.

The `name` argument can be any string and it specifies a event type. The timeline view will draw different types of events in different colors. It will also give you some statistics about the number of times a certain event type was seen, how long it took, etc.. 

The `details` is an object, whose properties are shown when you hover over a timeline segment in the profiler view. You can specify this object when you call `timeline.enter` or when you call `timeline.leave`. Usually, you have more information when you call `leave` so that's a more convenient place to put it.

The way in which you come up with event names can produce different results. In the `profilingWrapper` function, the `key` is used to specify the even type. At the moment, this is set to `MethodType[methodType]`, but you can easily set it to `implKey` which will give you a very detailed output.

[...]

PR #1216, opened 2015-03-10
Update readme

yxl:
This will fix https://github.com/mozilla/j2me.js/issues/1019.
It includes  @mbebenita's docs on the new profiling tools in #989 (comment) and other miscellaneous updates related to Makefile and native methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,11 +182,11 @@ Java compiler will do nothing to ensure that implementation actually exists. At 
 
 We use `Native` object in JS to handle creation and registration of `native` functions. See native.js
 
-    Native.create("name/of/function.(parameterTypes)returnType", jsFuncToCall, isAsync)
+    Native["name/of/function.(parameterTypes)returnType"] = jsFuncToCall;
 
 e.g.:
 
-    Native.create("java/lang/System.arraycopy.(Ljava/lang/Object;ILjava/lang/Object;II)V", function(src, srcOffset, dst, dstOffset, length) {...});
+    Native["java/lang/System.arraycopy.(Ljava/lang/Object;ILjava/lang/Object;II)V" = function(src, srcOffset, dst, dstOffset, length) {...};
 
 If you need to implement a method in JS but you can't declare it `native` in Java, use `Override`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,6 +142,59 @@ Modelines for JS files:
...
+timeline.leave(name?: string, details?: Object);
+```
+
+You must pair the `enter` and `leave` events but you don't necessarily need to specify arguments for `name` and `details`.
+
+The `name` argument can be any string and it specifies a event type. The timeline view will draw different types of events in different colors. It will also give you some statistics about the number of times a certain event type was seen, how long it took, etc.. 
+
+The `details` is an object, whose properties are shown when you hover over a timeline segment in the profiler view. You can specify this object when you call `timeline.enter` or when you call `timeline.leave`. Usually, you have more information when you call `leave` so that's a more convenient place to put it.
+
+The way in which you come up with event names can produce different results. In the `profilingWrapper` function, the `key` is used to specify the even type. At the moment, this is set to `MethodType[methodType]`, but you can easily set it to `implKey` which will give you a very detailed output.
```
mykmelez: Nit: even -> event.

README.md
```diff
@@ -142,6 +142,59 @@ Modelines for JS files:
...
+Additionally, you may create your own timelines. At the moment there are 3:
+- `timeline`: VM Events like loading class files, linking, etc.
+- `methodTimeline`: Method execution.
+- `threadTimeline`: Thread scheduling.
+
+You may have to change the CSS height style of the `profileContainer` if you don't see all timelines.
+
+![Shumway's timeline viewer](https://cloud.githubusercontent.com/assets/311082/5998278/644761ec-aa7a-11e4-8149-3556b08b8c54.png)
+
+Top band is an overview of all the timelines. Second band is the `timeline`, third is the `threadTimeline` and finally the fourth is the `methodTimeline`. Us your mouse wheel to zoom in and out, pan and hover.
```
mykmelez: Nit: Us -> Use

README.md
```diff
@@ -182,11 +235,11 @@ Java compiler will do nothing to ensure that implementation actually exists. At
 
 We use `Native` object in JS to handle creation and registration of `native` functions. See native.js
 
-    Native.create("name/of/function.(parameterTypes)returnType", jsFuncToCall, isAsync)
+    Native["name/of/function.(parameterTypes)returnType"] = jsFuncToCall;
 
 e.g.:
 
-    Native.create("java/lang/System.arraycopy.(Ljava/lang/Object;ILjava/lang/Object;II)V", function(src, srcOffset, dst, dstOffset, length) {...});
+    Native["java/lang/System.arraycopy.(Ljava/lang/Object;ILjava/lang/Object;II)V" = function(src, srcOffset, dst, dstOffset, length) {...};
```
mykmelez: In a followup, we should also add info about the new *Bindings* mechanism for defining natives.

README.md
```diff
@@ -200,17 +253,31 @@ If raising a Java `Exception`, throw new instance of Java `Exception` class as d
...
+
+The `asyncImpl` is optional if part of the code doesn't make async calls, the method could return value as regular native method and the VM will handle it properly.
+
+e.g:
+
+    Native["java/lang/Thread.sleep.(J)V"] = function(delay) {
+        asyncImpl("V", new Promise(function(resolve, reject) {
+            window.setTimeout(resolve, delay.toNumber());
+        }));
+    };
```
mykmelez: This example comes after the paragraph that describes how `asyncImpl` is optional, so it looks like it's supposed to be an example of that, but it's actually an example of an unconditional asynchronous native. So I would switch the order of that paragraph and the example, i.e.:

>e.g:
>```
>    Native["java/lang/Thread.sleep.(J)V"] = function(delay) {
>        asyncImpl("V", new Promise(function(resolve, reject) {
>            window.setTimeout(resolve, delay.toNumber());
>        }));
>    };
>```
>The `asyncImpl` is optional if part of the code doesn't make async calls, the method could return value as regular native method and the VM will handle it properly.

I would also rephrase and extend this description a bit, to clarify what it means for a native method to be partly asynchronous and emphasize that you always need to add such methods to `yieldMap`:

> The `asyncImpl` call is optional if part of the code doesn't make async calls. The method can sometimes return a value synchronously, and the VM will handle it properly. However, if a native ever calls `asyncImpl`, even if it doesn't always do so, then you need to add the method to `yieldMap`.

README.md
yxl: The followup bug is https://github.com/mozilla/j2me.js/issues/1223

README.md
yxl: @mykmelez Fixed in  45c0837. I add an example of partly asynchronous native methods as well.

mykmelez:
Nice example! I could use some newSleep. ;-)

PR #1221, opened 2015-03-10
Don't check isStatic when looking up methods and fields (since they're u...

marco-c:
...niquely identified by name and descriptor)


yxl:
I checked with Java compiler that you cannot declare a new static method with the same signature as that of a non-static method. So that means we only need to check the signature.

PR #1224, opened 2015-03-11
Add settings to change display size.

PinZhang:


mykmelez:
Some of this refactoring will be helpful in the future when hacking on the console.

main.js
```diff
@@ -177,6 +178,22 @@ document.getElementById("start").onclick = function() {
...
+    }
+  });
+
+  if (this.value) {
+    document.body.classList.add(this.value);
+  }
+
+  MIDP.updatePhysicalScreenSize();
+  MIDP.updateCanvas();
+  start();
```
mykmelez: This doesn't quite do what I would expect, as the display div does change size, but the canvas element within it doesn't do the same, when I pick a new size while running tests.

mykmelez: Here are some screenshots that show what happens. This is what I saw when switching from 240x320 to 480x320 while running unit tests:

screen shot 2015-03-12 at 17 54 09

And then after switching to 320x240:

screen shot 2015-03-12 at 17 54 20

PinZhang:
It seems that the canvas is not refreshed after `start()` was invoked, it might because the first few tests didn't use the canvas. Let's force to reset the canvas before calling `start()`.

PinZhang:
OK, the image floating on the top is not the canvas, it's the `image-launcher` div which should disappear when the app is restarting.

PinZhang:
I tried the AlertTest, we have the same issue with the alert UI, every time I clicked on the `start` button, the alert UI generated by the last test running wasn't removed, which means there will be lots of same UI elements left there if we clicked on the `start` button over and over again.

PinZhang:
> I tried the AlertTest, we have the same issue with the alert UI, every time I clicked on the start button, the alert UI generated by the last test running wasn't removed, which means there will be lots of same UI elements left there if we clicked on the start button over and over again.

The `start` button is only used for testing, I don't think this is really an issue, @mykmelez, what do you think?

PinZhang:
Fixed the styling issue of `image-launcher` div and re-based on top of master branch.

mykmelez:
> The start button is only used for testing, I don't think this is really an issue, @mykmelez, what do you think?

Agreed, it isn't an issue for this branch, since it's unrelated to your changes.

> Fixed the styling issue of image-launcher div and re-based on top of master branch.

Erm, I'm not sure exactly what changed here, since there isn't an individual commit with just this change. Can you provide more details?

mykmelez:
On Travis, LauncherTest is now failing with 720 different pixels.

PinZhang:
>>Fixed the styling issue of image-launcher div and re-based on top of master branch.

> Erm, I'm not sure exactly what changed here, since there isn't an individual commit with just this change. Can you provide more details?

I have put the change into an individual commit.

> On Travis, LauncherTest is now failing with 720 different pixels.

The image position is slightly different after I changed the styling rules, the image of the expected screenshot has been updated.

mykmelez:
> I have put the change into an individual commit.

Thanks! It makes good sense to put those rules into the stylesheet rather than applying them in the script.

marco-c:
This caused a regression of the "Downloading MIDlet" screen (when autosize=1).
I suspect it also caused a regression of the SMS verification screen.

PR #1225, opened 2015-03-11
UTF8ify Parser API & Lazify Linking

mbebenita:
This PR tries to remove most of the utf8 string parsing and simplifies name mangling. Virtual methods are now made available on the klass prototype as `v0`, `v1`, .... `vn`. (This name is referred to as the `virtualName` in the sources.) The numeric part of the name is the vTable index. Calls to virtual methods can be emitted as:
```
o.v1();
```
This is nice because we don't have to convert utf8 strings to JS strings whenever we want to mangle method names.

Class methods live on the klass objects, as `m0`, `m1`, ... , `mn`. The numeric part of the name is the method index in the class file. To call a class method directly, you can now write:

```
$java_lang_Object.m4(); // Static method.
$java_lang_Object.m4.call(x); // Invoke special. 
```

This also means that methods are no longer added to the JS global during linking. Instead, they hang off of the klass object.

This PR also includes a trampoline based lazy linking scheme. When a class is loaded, we inject trampoline methods for each new or overriden method in a class into the klass prototype. The first time the trampoline is called, we patch the prototype with the linked method. We only ever need as many trampolines as the maximum vTable size, which is usually small < 256.

Static methods are more problematic, because the trampoline cannot be reused, we would need a fresh closure each time.

In the case of virtual calls:
```
o.v1();
```
trampolines have access to the receiver via the `this` reference. For special calls:

```
$java_lang_Object.m4.call(x);
```

The trampoline would have access to the receiver `x`, but not to the class `java_lang_Object` which is what it needs to link up the method. Because of this, we now need a linking function `m` on each klass object:

```
$java_lang_Object.m(4).call(x);
```


mbebenita:
@mykmelez this should be ready for review. There are some low hanging fruit related to how we handle implKeys, but that belongs to a different PR.

mykmelez:
Hmm, I'm not seeing any startup perf improvement here. What are you seeing in your tests?

mbebenita:
> Hmm, I'm not seeing any startup perf improvement here. What are you seeing in your tests?

I have not yet profiled the code, will do so this afternoon.

interpreter.ts
```diff
@@ -106,8 +106,9 @@ module J2ME {
   var argArray = [];
 
   function buildExceptionLog(ex, stackTrace) {
-    var className = ex.klass.classInfo.className;
-    var detailMessage = util.fromJavaString(ex.klass.classInfo.getFieldByName("detailMessage", "Ljava/lang/String;", false).get(ex));
+    var className = ex.klass.classInfo.getClassNameSlow();
+    var classInfo: ClassInfo = ex.klass.classInfo;
+    var detailMessage = util.fromJavaString(classInfo.getFieldByName(toUTF8("detailMessage"), toUTF8("Ljava/lang/String;"), false).get(ex));
```
mykmelez: Nit: this function defines a *classInfo* local and then uses it only once, even though there are two references to *ex.klass.classInfo* in this function (the second one being the *getClassNameSlow* call above this line). The function should use the *classInfo* local for both references.

mbebenita:
If I isolate class parsing (deflate + parse class info), I'm seeing about 60% of the time spent in deflate. 

screen shot 2015-03-13 at 11 06 23 am

@marco-c didn't we have a patch to disable compression at some point?

vm/parser.ts
```diff
@@ -711,6 +861,17 @@ module J2ME {
       }
     }
 
+    getClassNameSlow(): string {
+      return this._name || (this._name = ByteStream.readString(this.utf8Name));
+    }
```
mykmelez: This (and all the other getters that are called often) should be memoizing, so they're only ever called once per instance:

```JS
    get className(): string {
      Object.defineProperty(this, "className", { value: ByteStream.readString(this.utf8Name) });
      return this.className;
    }
```

vm/runtime.ts
```diff
@@ -1349,130 +1352,153 @@ module J2ME {
...
+    }
+    linkKlassMethod(methodInfo.classInfo.klass, methodInfo);
+    assert (methodInfo.fn);
+    return methodInfo.fn;
+  }
 
-      // Link even non-static methods globally so they can be invoked statically via invokespecial.
-      if (updateGlobalObject) {
-        jsGlobal[methodInfo.mangledClassAndMethodName] = fn;
+  function linkKlassMethod(klass: Klass, methodInfo: MethodInfo) {
```
mykmelez: Nit: *klass* is always available from *methodInfo.classInfo.klass*, so it isn't necessary to pass it into *linkKlassMethod* as a separate parameter.

jit/baseline.ts
```diff
@@ -861,12 +861,16 @@ module J2ME {
...
           args.unshift(object);
-          call = methodInfo.mangledClassAndMethodName + ".call(" + args.join(",") + ")";
-        } else {
+          call = classConstant(methodInfo.classInfo) + ".m(" + methodInfo.index + ").call(" + args.join(",") + ")";
+        } else if (opcode === Bytecodes.INVOKEVIRTUAL) {
+          call = object + "." + methodInfo.virtualName + "(" + args.join(",") + ")";
+        } else if (opcode === Bytecodes.INVOKEINTERFACE) {
           call = object + "." + methodInfo.mangledName + "(" + args.join(",") + ")";
+        } else {
+          assert(false);
```
mykmelez: Nit: this wants an informative assertion message.

jit/compiler.ts
```diff
@@ -401,7 +350,7 @@ module J2ME {
           }
           classInfoList.push(classInfo);
         } catch (e) {
-          stderrWriter.writeLn(e);
+          stderrWriter.writeLn(e.stack);
```
mykmelez: This should also write the error message itself.

vm/runtime.ts
```diff
@@ -1349,130 +1352,153 @@ module J2ME {
...
+  }
 
-      methodInfo.fn = fn;
+  export function tryLinkMethod(methodInfo: MethodInfo) {
+    if (methodInfo.fn) {
+      return methodInfo.fn;
+    }
+    linkKlassMethod(methodInfo.classInfo.klass, methodInfo);
+    assert (methodInfo.fn);
+    return methodInfo.fn;
```
mykmelez: Nit: *try* in the name suggests this could fail. But no code accounts for that failure. Do we really expect this function to fail? If not, we should call it simply *linkMethod* (and make the assertion conditional on *!release*). Otherwise we need a better story for handling runtime failures of this function in release builds.

types.ts
```diff
@@ -116,32 +116,180 @@ module J2ME {
...
+   *  ComponentType:
+   *    FieldType
+   */
+
+  // Global state for signature parsing, kind of hackish but fast.
+  var globalNextIndex = 0;
+  var descriptorKinds = [];
+
+  /**
+   * Don't alias the returned array.
```
mykmelez: Nit: some additional explanation would be helpful here.

types.ts
```diff
@@ -116,32 +116,180 @@ module J2ME {
...
+      case UTF8Chars.V:
+        return Kind.Void;
+      case UTF8Chars.L: {
+        // parse a slashified Java class name
+        var endIndex = parseClassNameKind(value, startIndex + 1, UTF8Chars.Slash);
+        if (endIndex > startIndex + 1 && endIndex < value.length && value[endIndex] === UTF8Chars.Semicolon) {
+          globalNextIndex = endIndex + 1;
+          return Kind.Reference;
+        }
+        Debug.unexpected();
```
mykmelez: Nit: this should include a message explaining what's unexpected.

types.ts
```diff
@@ -116,32 +116,180 @@ module J2ME {
...
+        }
+        var dimensions = index - startIndex;
+        if (dimensions > 255) {
+          throw "array with more than 255 dimensions";;
+        }
+        var component = parseTypeDescriptorKind(value, index);
+        return Kind.Reference;
+      }
+      default:
+        Debug.unexpected(value[startIndex]);
```
mykmelez: Nit: I'd include a descriptive message here, like `"Unexpected type descriptor: " + value[startIndex]`.

vm/parser.ts
```diff
@@ -6,13 +6,54 @@
 module J2ME {
   declare var util;
   import assert = J2ME.Debug.assert;
+  import concat5 = StringUtilities.concat5;
+  import pushUnique = ArrayUtilities.pushUnique;
```
mykmelez: Nit: this code doesn't actually use *pushUnique*, so it doesn't need to import it.

vm/parser.ts
```diff
@@ -933,16 +1153,22 @@ module J2ME {
     }
     
     getAllInterfaces(): ClassInfo [] {
+      if (this.allInterfaces) {
+        return this.allInterfaces;
+      }
```
mykmelez: It isn't clear how often this is called, and it does cache the result, but *ArrayUtilities.pushMany* and *unique* are both pretty inefficient, so we should take a closer look at the cost of these changes.

vm/runtime.ts
```diff
@@ -1504,23 +1530,125 @@ module J2ME {
     }
   }
 
-  export function extendKlass(klass: Klass, superKlass: Klass) {
+  // Links the virtual method at a given index.
+  function linkVirtualMethodByIndex(self: java.lang.Object, index: number) {
+    // Self is the object on which the trampoline is called. We want to figure
+    // out the appropraite prototype object where we need to link the method. To
```
mykmelez: Nit: typo.

vm/runtime.ts
```diff
@@ -1504,23 +1530,125 @@ module J2ME {
...
+    // do this we look at self's class vTable, then find out the class of the
+    // bound method and then call linkKlassMethod to patch it on the appropraite
```
mykmelez: Nit: ditto.

vm/runtime.ts
mykmelez: > Nit: klass is always available from methodInfo.classInfo.klass, so it isn't necessary to pass it into linkKlassMethod as a separate parameter.

Erm, except that I guess the *Klass* we want isn't necessarily *methodInfo.classInfo.klass*, in *linkVirtualMethodByIndex* anyway.

vm/runtime.ts
```diff
@@ -1504,23 +1530,125 @@ module J2ME {
...
+    var klass = self.klass;
+    var classInfo = klass.classInfo;
+    var methodInfo = classInfo.vTable[index];
+    var methodKlass = methodInfo.classInfo.klass;
+    linkKlassMethod(methodKlass, methodInfo);
+    release || assert(methodInfo.fn);
+    return methodInfo.fn;
+  }
+
+  // Cache virtual trampolines.
```
mykmelez: Nit: virtual trampolines -> interface forwarders.

vm/parser.ts
mbebenita: My goal is to convert all accesses to classInfo.className, methodInfo.name and methodInfo.signature to getXXX methods. To keep the size of the patch down, I used getters for name and signature so I wouldn't have to update all the call sites.

I tried a quick benchmark, and the getXXX methods are the fastest, probably because they get inlined and have monomorphic call sites:

Getter: 216.31
Memoizer: 3.94
Call: 0.92

Getters are obviously really bad.

```

function foo() {
	var s = 0;
	for (var i = 0; i < 1000; i++) {
		var o = {_x: 123};	
		Object.defineProperty(o, "x", {
			get: function () {
				return this._x;
			}
		})
		for (var j = 0; j < 1000; j++) {
			s += o.x;
		}
	}
}

function bar() {
	var s = 0;
	for (var i = 0; i < 1000; i++) {
		var o = {_x: 123};	
		Object.defineProperty(o, "x", {
			get: function () {
				Object.defineProperty(o, "x", {value: this._x});
				return this._x;
			}, configurable: true
		})
		for (var j = 0; j < 1000; j++) {
			s += o.x;
		}
	}
}

function car() {
	var s = 0;
	for (var i = 0; i < 1000; i++) {
		var o = {_x: 123, getX: function () {
			return this._x;
		}};	
		for (var j = 0; j < 1000; j++) {
			s += o.getX();
		}
	}
}

var s = dateNow();
foo();
print("Getter: " + (dateNow() - s).toFixed(2));

var s = dateNow();
bar();
print("Memoizer: " + (dateNow() - s).toFixed(2));

var s = dateNow();
car();
print("Call: " + (dateNow() - s).toFixed(2));
```

mbebenita: The situation improves slightly if you keep the memorizing getter on the prototype, but it's still slower than getXXX.

marco-c:
> @marco-c didn't we have a patch to disable compression at some point?

We are currently keeping the classes JAR uncompressed and the MIDlet JAR compressed. We could easily store the MIDlet JAR as uncompressed in the JARStore if it improves performance.
A MIDlet JAR compressed is 2 MB, uncompressed 3,6 MB.

mbebenita:
> We are currently keeping the classes JAR uncompressed and the MIDlet JAR compressed. We could easily store the MIDlet JAR as uncompressed in the JARStore if it improves performance.
> A MIDlet JAR compressed is 2 MB, uncompressed 3,6 MB.

We eventually decompress most of it anyway, so we're looking at 2+3.6MB instead of 3.6MB. The new parsing code overlays the original buffer, so there's no additional copying going on.

marco-c:
> We eventually decompress most of it anyway, so we're looking at 2+3.6MB instead of 3.6MB. The new parsing code overlays the original buffer, so there's no additional copying going on.

We're releasing the original ArrayBuffer when we decompress a file from the JAR, so it doesn't increase to 2+3,6 MB.
Part of the 2 MB will grow when decompressed (up to 3,6 MB), but some of the data will still be in the compressed form and never be decompressed (so it's never really going to be 3,6 MB). I'm not sure how much data is never loaded from the JAR in a typical run of a MIDlet, we should run some measurements.

mbebenita:
> we should run some measurements.

For some midlets we load roughly 1,463,260 bytes at startup.

marco-c:
> For some midlets we load roughly 1,463,260 bytes at startup.

From the MIDlet JAR or from both JARs?

mbebenita:
> From the MIDlet JAR or from both JARs?

From all JARs.

marco-c:
> From all JARs.

OK, I've run a quick measurement. We're loading a total of 617892 bytes (that grow to 1292806 bytes when decompressed).

vm/parser.ts
mbebenita: This is called ~ 630 times during the startup of some midlets. I agree that we should think of better ways to optimize interface collection, maybe using a custom built hash map.

mbebenita:
> OK, I've run a quick measurement. We're loading a total of 617892 bytes (that grow to 1292806 bytes when decompressed).

Question is, what's faster, loading or decompressing? If it's just a memory concern, I think we can live with the additional memory consumption if we get to skip out on decompression.

marco-c:
> Question is, what's faster, loading or decompressing? If it's just a memory concern, I think we can live with the additional memory consumption if we get to skip out on decompression.

Loading is faster, not sure by how much. Memory consumption could be increased by nearly 1 MB.

vm/parser.ts
andreasgal: Can you put a getter on the proto and a direct property on the object? After the first access it should be super fast.

mbebenita: That's what the previous comment was about. I think old school Java like getters are the way to go. They have the advantage that they don't mutate the shape of the object and avoid the need for calling Object.defineProperty which is pretty slow itself.

mykmelez: > I tried a quick benchmark, and the getXXX methods are the fastest, probably because they get inlined and have monomorphic call sites:

Erm, but your benchmark doesn't measure what your code actually does, nor what I suggest it should do instead. Here's a [more accurate one in JSFiddle](http://jsfiddle.net/dt4z21af/) based on the actual JavaScript code to which the TypeScript compiles. It shows that the memoizer is faster.

Part of the problem may be that your version mutates the shape of the object, since it doesn't define *this._name* in the *ClassInfo* constructor. And [fixing that in an updated fiddle](http://jsfiddle.net/dt4z21af/1/) brings it in line with the memoizer's perf (and works as long as you ensure that the get*() method correctly distinguishes between the initial value and the eventual one; which can probably be a simple check for truthiness), for 1000 objects with 1000 accesses each, anyway.

But bump those numbers, and your version starts slowing down again.

Here's [one thousand objects with ten thousand accesses](http://jsfiddle.net/dt4z21af/2/). And [ten thousand objects with one thousand accesses](http://jsfiddle.net/dt4z21af/3/). I'm not sure what these numbers will actually look like in the field, but I see ~4500 calls to *getClassNameSlow* on startup when loading the very simple [InitMidlet](https://github.com/mozilla/j2me.js/blob/master/tests/midlets/InitMidlet.java).

> To keep the size of the patch down, I used getters for name and signature so I wouldn't have to update all the call sites.

That's a worthy concern, but this changeset would be even smaller if you didn't replace all the *className* references with *getClassNameSlow* calls.

mbebenita: > But bump those numbers, and your version starts slowing down again.

Part of the issue here is that `_x` is of a different type than the final value, setting a sentinel value of the same type as the final value makes the getXXX pattern pull out ahead in all configurations: http://jsfiddle.net/dt4z21af/5/

The memoizer version falls behind here because it allocates a lot of temporary objects. Here is a workaround for that: http://jsfiddle.net/dt4z21af/6/. This closes the gap but falls behind when the number of accesses is low and the number of objects is high.

```
Objects: 128 x Accesses: 32768

Memoizer: 7.31
Call: 7.52

Objects: 256 x Accesses: 16384

Memoizer: 6.38
Call: 6.88

Objects: 512 x Accesses: 8192

Memoizer: 6.61
Call: 6.67

Objects: 1024 x Accesses: 4096

Memoizer: 7.38
Call: 5.64

Objects: 2048 x Accesses: 2048

Memoizer: 7.61
Call: 6.92

Objects: 4096 x Accesses: 1024

Memoizer: 8.41
Call: 5.70

Objects: 8192 x Accesses: 512

Memoizer: 11.04
Call: 6.58

Objects: 16384 x Accesses: 256

Memoizer: 16.19
Call: 5.89

Objects: 32768 x Accesses: 128

Memoizer: 27.84
Call: 7.18
```

mbebenita: > That's a worthy concern, but this changeset would be even smaller if you didn't replace all the className references with getClassNameSlow calls.

True. I was intending to change name, signature, implKey and all the other getters to getXXXSlow() to discourage use, but I postponed that for a future PR.

mykmelez: > The memoizer version falls behind here because it allocates a lot of temporary objects. Here is a workaround for that: http://jsfiddle.net/dt4z21af/6/. This closes the gap but falls behind when the number of accesses is low and the number of objects is high.

Indeed, it looks like *Object.defineProperty* is responsible for the cost. Replacing it with a private *this._x* member makes the prototype getter (no longer accurate to call it a memoizer) twice as fast again:

http://jsfiddle.net/dt4z21af/7/

Things change somewhat if the value is computed (as it will be in our use case). Then, the prototype getter is still faster, but the numbers are closer, and occasionally flip:

http://jsfiddle.net/dt4z21af/8/

Which still litigates for the prototype getter.

mbebenita: This would be a very unexpected result. At this point, both the getter and the function call are effectively the same thing, and are most likely optimized the same way. I was really surprised to see that the function call was slower, so I tried it in the jsshell. In the shell there is no difference between getters and function calls.

I also tried it in an html page and the results were the same as the shell. This suggests that jsfiddle is not to be trusted, ... now I'm curious why it's slower there, I suspect they may be using `eval`.

Which, btw, I just tried and the perf in `(1, eval)(...)` is 2x slower on both.

mykmelez: > This would be a very unexpected result. At this point, both the getter and the function call are effectively the same thing, and are most likely optimized the same way. I was really surprised to see that the function call was slower, so I tried it in the jsshell. In the shell there is no difference between getters and function calls.

Ah, indeed, I get the same results. I suspect our use case resembles the few objects/many accesses scenario, where the difference between the memoizer and (either) getter is insignificant. But I'd still rather not depend on that. So you've convinced me that the getter approach is better.

> I also tried it in an html page and the results were the same as the shell. This suggests that jsfiddle is not to be trusted, ... now I'm curious why it's slower there, I suspect they may be using eval.

Indeed, I'm also curious. cc: @zalun for his thoughts.

mbebenita:
The last bug fix has a fairly large impact on performance, since we used to interpret a lot of compiled code.

mykmelez:
> The last bug fix has a fairly large impact on performance, since we used to interpret a lot of compiled code.

How much of a win are we getting overall now on this branch?

mbebenita:
Too noisy to tell on my desktop. They are both around 1s for me, this branch being slightly faster. I'm actually measuring against #1236 since that has a lot of follow up improvements to this branch.

mykmelez:
> Too noisy to tell on my desktop. They are both around 1s for me, this branch being slightly faster. I'm actually measuring against #1236 since that has a lot of follow up improvements to this branch.

Erm, how were you able to determine that b222861 has a fairly large impact on performance?

In any case, improvements don't always show up on desktop (and vice-versa), so it'd be better to test on device. On a Flame with 2.1, I see some improvement on startupTime and a huge improvement on jsStringsSize with both this branch and #1236 merged along with @brendandahl's #1233, although jsObjectsSize is worse, and overall memory consumption is unchanged (but see the discussion in #1233 for reasons to suspect the memory numbers):

|      Test     |   Mean   | Baseline Mean |   +/-  |    %   |       P       |    Min   |    Max   | 
|:--------------|---------:|--------------:|-------:|-------:|--------------:|---------:|---------:|
| startupTime   | 13,238ms |      13,819ms | -581ms |  -4.21 |        BETTER | 12,204ms | 14,155ms | 
| totalSize     | 35,089kb |      34,872kb |  216kb |   0.62 | INSIGNIFICANT | 31,855kb | 38,829kb | 
| domSize       |     83kb |          83kb |    0kb |   0.00 | INSIGNIFICANT |     83kb |     83kb | 
| styleSize     |    282kb |         282kb |    0kb |   0.15 | INSIGNIFICANT |    281kb |    283kb | 
| jsObjectsSize | 19,040kb |      18,217kb |  824kb |   4.52 |         WORSE | 18,999kb | 19,083kb | 
| jsStringsSize |    682kb |       1,121kb | -439kb | -39.18 |        BETTER |    681kb |    691kb | 
| jsOtherSize   | 14,831kb |      15,005kb | -174kb |  -1.16 | INSIGNIFICANT | 11,638kb | 18,544kb | 
| otherSize     |    170kb |         164kb |    6kb |   3.38 | INSIGNIFICANT |    157kb |    185kb |

mbebenita:
> Erm, how were you able to determine that b222861 has a fairly large impact on performance?

I was measuring with caching disabled and no AOT. The bug was causing a large number of methods to be interpreted after they had been compiled.

mbebenita:
> although jsObjectsSize is worse

Probably because most strings are now represented as Uint8Array buffers, so the change in jsStringSize and jsObjectSize numbers make sense.

mykmelez:
> I was measuring with caching disabled and no AOT. The bug was causing a large number of methods to be interpreted after they had been compiled.

Erm, sorry, I guess I misunderstood. I understand the theoretical impact, it just sounded like you'd seen an actual impact in testing.

vm/parser.ts
zalun: The result is always prepared as a new page. Generated inside Django. Maybe there is a difference when a tested code is run inside iframe?

PR #1233, opened 2015-03-12
Add memory sampling to startup benchmark.

brendandahl:
With firefox and no security we can now capture memory use.  Seems relatively stable with a 5000ms timer.

Creates pretty tables now:

|      Test     |   Mean   | Baseline Mean |   +/-  |   %   |       P       |    Min   |    Max   | 
|:--------------|---------:|--------------:|-------:|------:|--------------:|---------:|---------:|
| startupTime   |  1,118ms |       1,104ms |   14ms |  1.23 | INSIGNIFICANT |  1,078ms |  1,194ms | 
| totalSize     | 51,937kb |      52,452kb | -515kb | -0.98 | INSIGNIFICANT | 50,518kb | 53,287kb | 
| domSize       |    182kb |         182kb |    0kb |  0.00 | INSIGNIFICANT |    182kb |    182kb | 
| styleSize     |    419kb |         419kb |    0kb | -0.05 | INSIGNIFICANT |    419kb |    420kb | 
| jsObjectsSize | 21,190kb |      21,189kb |    1kb |  0.00 | INSIGNIFICANT | 21,181kb | 21,196kb | 
| jsStringsSize |  1,455kb |       1,455kb |    0kb |  0.00 | INSIGNIFICANT |  1,455kb |  1,455kb | 
| jsOtherSize   | 28,277kb |      28,790kb | -513kb | -1.78 | INSIGNIFICANT | 26,857kb | 29,640kb | 
| otherSize     |    413kb |         416kb |   -3kb | -0.70 | INSIGNIFICANT |    409kb |    425kb |

### DIFF
--- a/benchmark.js
+++ b/benchmark.js
@@ -177,6 +177,9 @@ var Benchmark = (function() {
       storage.current.startupTime.push(took);
       storage.round++;
       saveStorage();
+      this.runNextRound();
+    },
+    sampleMemory: function() {
       if (NO_SECURITY) {
         forceCollectors();
         var memoryReporter = Components.classes["@mozilla.org/memory-reporter-manager;1"].getService(Components.interfaces.nsIMemoryReporterManager);
@@ -206,15 +209,21 @@ var Benchmark = (function() {
         storage.current.otherSize.push(otherSize.value);
         saveStorage();
       }
-      if (storage.round >= storage.numRounds) {
-        this.finish();
-        return;
-      }
-      this.runNextRound();
     },
     runNextRound: function() {
+      var self = this;
+      var done = storage.round >= storage.numRounds;
       function run() {
         forceCollectors();
+        if (storage.round !== 0) {
+          if (NO_SECURITY) {
+            self.sampleMemory();
+          }
+          if (done) {
+            self.finish();
+            return;
+          }
+        }
         DumbPipe.close(DumbPipe.open("reload", {}));
       }
       if (storage.deleteFs) {
@@ -225,8 +234,12 @@ var Benchmark = (function() {
         console.log("Deleting jit cache.");
         indexedDB.deleteDatabase("CompiledMethodCache");
       }
-      console.log("Scheduling round " + (storage.round + 1) + " of " + storage.numRounds + " to run in " + storage.roundDelay + "ms");
-      setTimeout(run, storage.roundDelay);
+      if (storage.round !== 0) {
+        console.log("Scheduling round " + (storage.round) + " of " + storage.numRounds + " finalization in " + storage.roundDelay + "ms");
+        setTimeout(run, storage.roundDelay);
+      } else {
+        run();
+      }
     },
     finish: function() {
       storage.running = false;
--- a/benchmark.js
+++ b/benchmark.js
@@ -10,44 +10,140 @@ var Benchmark = (function() {
   var defaultStorage = {
     numRounds: 10,
     roundDelay: 5000, // ms to delay starting next round of tests
-    baseline: [],
+    baseline: {},
+    current: {},
     running: false,
     round: 0,
-    times: [],
-    deleteFs: true,
-    deleteJitCache: true,
-    buildBaseline: false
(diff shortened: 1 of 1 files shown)


benchmark.js
```diff
@@ -10,44 +10,125 @@ var Benchmark = (function() {
...
+        var middle = ((n - str.length) / 2) | 0;
+        return str.padRight(repeat, middle + str.length).padLeft(repeat, n);
+      } else if (align === RIGHT) {
+        return str.padLeft(repeat, n);
+      }
+      throw new Error("Bad align value." + align);
     }
-  };
+    var maxColumnLengths = [];
+    var numColums = rows[0].length;
```
mykmelez: Nit: numColums -> numColumns

mykmelez:
Looks great.

I'd probably order the columns: `Baseline Mean, Mean, +/-, %, P, Min, Max`; reversing the order of *Baseline Mean* and *Mean*, so it reads as `from -> to`.

I still get significant noise in the memory numbers while benchmarking a midlet in a profile with *security.turn_off_all_security_so_that_viruses_can_take_over_this_computer* enabled. Here's a sample run where I built the benchmark and then did a run without making any changes:

|      Test     |   Mean   | Baseline Mean |  +/-  |   %   |       P       |    Min   |    Max   | 
|:--------------|---------:|--------------:|------:|------:|--------------:|---------:|---------:|
| startupTime   |  1,237ms |       1,239ms |  -2ms | -0.13 | INSIGNIFICANT |  1,182ms |  1,294ms | 
| totalSize     | 36,724kb |      35,887kb | 837kb |  2.33 |         WORSE | 36,028kb | 37,932kb | 
| domSize       |    167kb |         167kb |   0kb |  0.00 | INSIGNIFICANT |    167kb |    167kb | 
| styleSize     |    418kb |         418kb |   0kb |  0.00 | INSIGNIFICANT |    418kb |    418kb | 
| jsObjectsSize | 18,276kb |      18,259kb |  17kb |  0.09 | INSIGNIFICANT | 18,244kb | 18,402kb | 
| jsStringsSize |    995kb |         995kb |   0kb |  0.00 | INSIGNIFICANT |    995kb |    996kb | 
| jsOtherSize   | 16,476kb |      15,656kb | 820kb |  5.24 |         WORSE | 15,811kb | 17,715kb | 
| otherSize     |    392kb |         392kb |   0kb |  0.00 | INSIGNIFICANT |    391kb |    392kb | 

Also, I noticed that we don't force CC/GC before the first run. We should do that (and perhaps force it right before reloading the page, rather than before the timeout, which gives the midlet time to accumulate more garbage). But those are both existing issues, unrelated to this branch, I think.

The raw numbers probably aren't as interesting anymore, especially now that they're incomplete.

mykmelez:
(Although I have found it useful to look for outliers in the raw startup numbers.)

mykmelez:
> Also, I noticed that we don't force CC/GC before the first run. We should do that (and perhaps force it right before reloading the page, rather than before the timeout, which gives the midlet time to accumulate more garbage). But those are both existing issues, unrelated to this branch, I think.

Hmm, no, the first one is from moving the forceCC/forceGC calls from *runNextRound* to *stopTimer*. But I see why you did it: to collect the garbage before measuring memory. May still make sense to do it in *runNextRound*, to avoid any effect on startup time. Putting that call inside the *run* function would fix the second issue.

brendandahl:
I'm now forcing gc before we measure and right before we reload the page so we can hopefully get rid of more objects before we start the next round.

brendandahl:
I should mention I'm still seeing too much variance on the memory. Re-running doesn't seem to provide consistent results.

brendandahl:
I moved the memory sampling to happen before the page is reloaded to run the next round. This seem to give more stable results. I built three baselines and ran five benchmarks on each, all but one showed insignificant results.

mykmelez:
The build was only busted by #1247, which I've merged, so I'm merging this.

PR #1235, opened 2015-03-13
Show SMS prompt when the SubscribeMessages method of the nokia.messaging...

marco-c:
... localmsg server is used


marco-c:
We've seen MIDlets that use both the APIs to receive SMS messages and MIDlets that only use the Nokia API, so creating the prompt when the Nokia API is used should cover both cases.

mykmelez:
Makes sense, but let's keep an eye out for those that only use the MIDP API (if there are any), in case we need to do something more sophisticated here.

PR #1236, opened 2015-03-14
Class loading optimizations. 

mbebenita:
This PR includes optimizations on top of #1225


vm/parser.ts
```diff
@@ -6,20 +6,62 @@
...
+  export function strcmp(a: Uint8Array, b: Uint8Array): boolean {
     if (a === b) {
       return true;
     }
     if (a.length !== b.length) {
       return false;
     }
-    for (var i = 0; i < a.length; i++) {
+    var l = a.length;
+    for (var i = 0; i < l; i++) {
```
mykmelez: Doesn't SM optimize `i < a.length`, so there's no value in assigning *a.length* to a local (here and in *uint8ArrayEqual*)?

vm/parser.ts
```diff
@@ -645,50 +787,73 @@ module J2ME {
...
       if (!buffer) {
+        sealObjects && Object.seal(this);
         return;
       }
+      enterTimeline("ClassInfo");
       var s = this;
-      s.readU4(); // magic
+      s.skipU4(); // magic
       s.readU2(); // minor_version
       s.readU2(); // major_version
```
mykmelez: We could also add *skipU2* for the major and minor versions.

jit/analyze.ts
```diff
@@ -220,6 +221,10 @@ module J2ME {
   }
 
   export function canYield(methodInfo: MethodInfo): YieldReason {
+    if (phase === ExecutionPhase.Runtime && methodInfo.codeAttribute && methodInfo.codeAttribute.code.length > 100) {
+      // Large methods are unlikely to not yield, so don't even bother checking at runtime.
+      return YieldReason.Unlikely;
```
mykmelez: Nit: I'd reverse the logic here, calling this YieldReason.Likely, for large methods that are likely to yield, to avoid the double negative.

vm/parser.ts
mbebenita: I don't know if it does or not. I did see some comments on bugzilla about SM being able to optimize range checks if it sees the `i < a.length` pattern. I'll change it back to that.

jit/analyze.ts
mbebenita: True. Will fix.

mykmelez:
@mbebenita Are there any additional changes you plan to make here?

mbebenita:
I'm just doing little profile based optimizations. Should I just submit a new PR?

mykmelez:
> I'm just doing little profile based optimizations. Should I just submit a new PR?

Yes, that way we can review and land this one, which already has some demonstrated perf benefit.

PR #1240, opened 2015-03-15
Fix resolving methods for abstract classes

marco-c:


mykmelez:
Nice catch!

(Nit: I'd probably rename *a* to *intMethod* and *method* to *absMethod* in the testlet to make their provenance clearer.)

PR #1246, opened 2015-03-16
Refactor j2me localization support

marco-c:
This PR:
1) Fixes a bug (there was a mismatch between en-US.xml and ResourceConstants.java)
2) Stores the localized strings in a JSON file (instead of an XML one)
3) Adds support for languages other than en-US

In a follow-up, I'll add some documentation for localization in the README file.


marco-c:
(The it-IT.xml file isn't actually localized, I've just translated a couple of strings in order to test loading strings in a different language)

marco-c:
The failure isn't related to this branch.

l10n/Makefile
```diff
@@ -0,0 +1,10 @@
+LANG_FILES=$(shell find . -name "*.xml")
+LANG_DESTS=../java/l10n/$(LANG_FILES:.xml=.json)
+
+$(LANG_DESTS): $(LANG_FILES)
```
mykmelez: There's no need to put this rule into a separate Makefile, and [Recursive Make Considered Harmful](http://aegis.sourceforge.net/auug97.pdf), so I would put this into the top-level Makefile, even though it takes files from l10n/ as input and generates output in java/.

l10n/xml_to_json.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
```
mykmelez: Nit: I'd put this in tools/.

mykmelez:
Just needs a rebase.

PR #1251, opened 2015-03-17
Make strcat and strcat3 concatenate single characters instead of Uint8Arrays

marco-c:


vm/parser.ts
```diff
@@ -81,18 +78,23 @@ module J2ME {
...
-  function strcat(a: Uint8Array, b: Uint8Array): Uint8Array {
-    var r = new Uint8Array(a.length + b.length);
-    r.set(a, 0);
-    r.set(b, a.length);
+  function strcatSingle(a: number, b: Uint8Array): Uint8Array {
+    var r = new Uint8Array(1 + b.length);
+    r[0] = a;
+    for (var i = 1; i < b.length + 1; i++) {
+      r[i] = b[i - 1];
+    }
```
mykmelez: Nit: I would add a comment here (and in strcat4Single) about why we don't use Uint8Array::set.

mykmelez:
Just needs a rebase.

PR #1252, opened 2015-03-17
Remove unneeded Permissions native implementations

marco-c:
They're unneeded since we're relying on the Firefox OS security model (or web security model).


mykmelez:
Nice. Is there Java code we can strip out (or avoid loading)?

marco-c:
> Nice. Is there Java code we can strip out (or avoid loading)?

Yes, there's a lot of stuff that we could remove, but nothing that seems particularly slow.

PR #1257, opened 2015-03-17
Only create the pathname when debugging fs.

brendandahl:


mykmelez:
Just needs a rebase because of #1256.

PR #1260, opened 2015-03-18
Reduce relooper's memory.

mbebenita:


marco-c:
What happens if it runs out of memory?

mbebenita:
Usually, it throws an assertion. I haven't seen it run out of memory yet. We have a maximum method size threshold in the JIT, we can lower it if we run out of memory.

PR #1263, opened 2015-03-18
Trim slack, saves ~ 0.5MB on some midlets.

mbebenita:


mbebenita:
I should mention that, the `0.5MB` number I got from counting the amount of slack we were producing, not from the memory benchmark profiler.

PR #1265, opened 2015-03-18
Emit constant pool references instead of escaping strings and some optimizations.

mbebenita:


jit/baseline.ts
```diff
@@ -380,8 +380,16 @@ module J2ME {
       this.emitPrologue();
       this.emitBody();
 
-      if (this.variables.length) {
-        this.bodyEmitter.prependLn("var " + this.variables.join(",") + ";");
+      if (Object.keys(this.variables).length) {
```
marco-c: You could do something like:

```JavaScript
var variables = [];
for (var k in this.variables) {
  if (this.variables[k] !== undefined) {
    variables.push(k + "=" + this.variables[k]);
  } else {
    variables.push(k);
  }
}
if (variables.length > 0) {
  this.bodyEmitter.prependLn("var " + variables.join(",") + ";");
}
```

mbebenita: Good point.

marco-c:
There was a conflict, so I've merged it from the command line with an additional commit to remove some more unused code (https://github.com/mozilla/j2me.js/commit/393713c09a4146ed97e3d526b71e0fe64e7b3029)

Issue #1266, opened 2015-03-18
Investigate FS memory usage


marco-c:
We're relying on the fact that IndexedDB doesn't store Blobs in memory right away.
We should:
- Find out if this is the case on older Firefox OS versions as well.
- Find out what happens when we close a file. We're using Blob::readAsArrayBuffer and we release the ArrayBuffer when a file is closed, is it actually going to be released from memory? Or is it still in memory as long as we hold the blob?

mykmelez:
> Find out if this is the case on older Firefox OS versions as well.

Even if Gecko lazy-loads blobs on all the versions we care about, there's still the problem that we keep the blob in memory in perpetuity once a file has been opened for the first time (at least, that's what I expect is happening after we read the blob into the ArrayBuffer). So we'll need to move file data to a separate object store anyway to release that memory after a file has been closed. See #1268 for the WIP that does that.

> Find out what happens when we close a file. We're using Blob::readAsArrayBuffer and we release the ArrayBuffer when a file is closed, is it actually going to be released from memory? Or is it still in memory as long as we hold the blob?

Good questions, but probably we don't need to answer them, if we land #1268, since then we'll be releasing the blob after reading from it.

marco-c:
I've run some measurements as part of https://github.com/mozilla/j2me.js/pull/1462 and I noticed the FS library is probably the cause of high USS peaks (it's also using more memory in general compared to https://github.com/mozilla/j2me.js/pull/1462).

PR #1268, opened 2015-03-18
WIP: store file data in separate object store

mykmelez:
This branch moves file data to a separate object store, to avoid keeping it in memory after a file is opened for the first time, and to avoid having to read it into memory during startup on older FxOS versions that don't lazy-read blobs from IndexedDB (although it's unclear exactly when that optimization landed; so this may or may not matter).

WIP as I'm still testing correctness and perf impact. I don't expect this to have much impact on startup perf and memory usage, although it might have a bit. I do expect it to matter for ongoing memory usage, as currently we keep file data in memory in perpetuity after the first time a file is opened.

This addresses part of #1266.


mykmelez:
I've done a variety of tests, and I haven't been able to detect a difference in memory usage, neither on startup nor at runtime, with these changes. Not even on FxOS 1.3. So I think Gecko must maintain only a reference to a blob, regardless of whether or not it's been read into an ArrayBuffer. (Of course the ArrayBuffer itself takes up memory.)

PR #1273, opened 2015-03-18
build without console library

mykmelez:
The console library stores each message sent to the consoles in perpetuity, so the in-page console can be filtered later, which means its memory usage grows unbounded. And that feature is useless on device, where we can't see the in-page console anyway. So we should disable it when building for distribution.

Here's a changeset that adds a *CONSOLE* flag to configure the build to exclude the library (and in-page console). It's set to `1` by default, which includes the library. `make CONSOLE=0` to exclude it.

There's more we should remove when we're building for distribution, like the rest of the developer panel. Maybe we need a different flag to disable all that stuff, like *DISTRIB*. But I'll start with this, which has memory implications.

I didn't simply overload *RELEASE* because it's useful to build a release build for testing on desktop.

We still want *console.print* to work, for *com/sun/cldchi/io/ConsoleOutputStream.write*, so I moved it to a new *NativeConsoleOutputStream* object.

jsshell also defines *console.print*, but it doesn't need to, so I removed it (and also redirected *console.info* and *console.error* to *print*, which seems like the correct way of handling calls to those functions). Running the Basic program, which calls `System.out.println("The end")`, results in:

>03-18 15:09 > js jsshell.js Basic
>pre-main prep time: 0 ms
>Loaded java/classes.jar in 0.48 ms.
>The end
>Time: 40.1230 ms


marco-c:
I've restarted the build a few times, but it's always failing for a timeout. Unsure if these changes are the cause of the timeout, since I've been seeing a lot of PRs failing for timeouts recently.

mykmelez:
> I've restarted the build a few times, but it's always failing for a timeout. Unsure if these changes are the cause of the timeout, since I've been seeing a lot of PRs failing for timeouts recently.

The changes are the cause of the timeout, although the reason is more interesting than I expected.

After landing more tests recently, the "basic tests" take almost 70 seconds to run (~66 on firstrun, ~63 on second run on my machine). So anything that slows them down a few seconds will trigger test timeout failure. Which is probably why you've been seeing PRs failing for timeouts. (I requested #1291 to increase the timeout.)

And these changes were slowing down tests pretty significantly, causing them to take ~90 seconds to run on my machine, which is a significant regression.

After some investigation, I figured out that the regression is caused by the *flush* function being in a separate script from its caller, since the caller is in console.js, and I moved *flush* into native.js.

I don't actually understand how crossing a script boundary can cause this kind of regression. I thought perhaps it was a closure boundary, since the console code uses the module pattern to encapsulate its implementation in a function that closes over the global scope to which I'd moved *flush*.

But putting all of console.js into the global scope, removing the module encapsulation, didn't make a difference; and moving the function back into console.js did fix the problem, even if I put the function into the global scope.

This makes me wonder how many other instances there are in our codebase of a caller in one script calling a function in another script, and whether they're all similarly slower. In any case, the solution for this particular PR is to define flush inside console.js, if building with CONSOLE=1; and otherwise define it in native.js.

Nevertheless, this doesn't buy us much, since release builds for device are already built with `logConsole=web`, which disables most of console.js and pipes messages directly to the web console. So this won't gain us much in distribution builds (besides whatever small benefit we get from removing console.js itself).

Here are some memory usage numbers (in MB) after running SystemOutBench, which writes a bunch of messages to the console:

With Console: 14.38, 14.43, 13.84, 14.20, 14.32
Without Console: 11.08, 11.50, 11.52, 11.15, 11.33
With Console, but logConsole=web: 11.25, 11.32, 11.36, 11.52, 11.33

The savings without the console are significant, but `logConsole=web` is equivalent. So I'm not sure we should take this.

marco-c:
> This makes me wonder how many other instances there are in our codebase of a caller in one script calling a function in another script, and whether they're all similarly slower. In any case, the solution for this particular PR is to define flush inside console.js, if building with CONSOLE=1; and otherwise define it in native.js.

We should introduce a build step that joins the source files in a single file (and uses the closure compiler to optimize it). This way we wouldn't have to define functions in different files only for this purpose.

marco-c:
I guess the changes should be simpler now. I think we can take the patch, even if it doesn't improve memory usage much.

mykmelez:
> I guess the changes should be simpler now. I think we can take the patch, even if it doesn't improve memory usage much.

Ok, I've updated the branch to resolve conflicts and remove some additional HTML/CSS if the console is disabled.

mykmelez:
I guess I don't really understand the *console* implementation in jsshell.js. I've reverted my changes, which were unnecessary, and now the jsshell test succeeds.

(FWIW, my changes work fine with the 37.0a1 build of jsshell that I was using to test, but they fail with the 37.0 version that Travis uses. I'll switch to the version that Travis uses, so I don't experience this problem in the future.)

libs/console.js
```diff
@@ -351,9 +333,26 @@
     info: logAtLevel.bind(null, "info"),
     warn: logAtLevel.bind(null, "warn"),
     error: logAtLevel.bind(null, "error"),
-    print: print,
     profile: typeof console !== "undefined" && console.profile ? console.profile.bind(console) : null,
     profileEnd: typeof console !== "undefined" && console.profileEnd ? console.profileEnd.bind(console) : null,
   };
 
+  window.console.buffer = "";
```
marco-c: Nit: Could we define these only in native.js?

marco-c:
Looks good to me, just needs to be updated.

libs/console.js
mykmelez: > Nit: Could we define these only in native.js?

I tried to do that originally, but it resulted in a performance regression, because of the cross-script calls. That was before we started collating main.html scripts, however, and it no longer causes that regression. So I've done this.

mykmelez:
The CasperJS-based tests are all passing, but `make test` is returning a non-zero error code:

> The command "make test" exited with 2.

Also, Travis is hitting its 10000 line log limit, which may or may not be related. Looking into it…

mykmelez:
jsshell.js defines *console.print* and then loads native.js, which was redefining it. f99c0fa fixes that by defining it conditionally.

marco-c:
I think you could simply redefine console.print in jsshell.js after loading native.js. I'd prefer a change to jsshell.js instead of native.js, but it's a nit. So feel free to merge either with or without applying this last comment.

mykmelez:
> I think you could simply redefine console.print in jsshell.js after loading native.js. I'd prefer a change to jsshell.js instead of native.js, but it's a nit.

I think you're right, so I did that in e5c64af.

> So feel free to merge either with or without applying this last comment.

Ok, merging…

PR #1274, opened 2015-03-18
Create ZIP buffers with the correct size right from the beginning

marco-c:
The upgrade step is just deleting the old object store and creating a new one (main.js will then re-download the MIDlet JAR).
It'd be complex to migrate the old data to the new format with the *uncompressed_len* property (well, it wouldn't be difficult, but it would require either writing a new function similar to *inflate* or modifying *inflate* to work for buffers with an unknown length, so I think handling it this way is much cleaner).

### DIFF
--- a/libs/jarstore.js
+++ b/libs/jarstore.js
@@ -5,8 +5,9 @@
 
 var JARStore = (function() {
   var DATABASE = "JARStore";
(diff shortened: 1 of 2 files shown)


mykmelez:
Redownloading the midlet is a fine upgrade step.

PR #1275, opened 2015-03-18
Change executeFrames to only execute one frame.

brendandahl:
We don't currently execute more than one frame at a time in executeFrames. This gets rid of the array allocation.

### DIFF
--- a/vm/context.ts
+++ b/vm/context.ts
@@ -414,13 +414,9 @@ module J2ME {
       release || assert (Frame.isMarker(marker));
     }
 
-    executeFrames(group: Frame []) {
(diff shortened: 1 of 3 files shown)


vm/context.ts
```diff
@@ -414,13 +414,9 @@ module J2ME {
       release || assert (Frame.isMarker(marker));
     }
 
-    executeFrames(group: Frame []) {
+    executeFrame(frame: Frame) {
```
marco-c: Nit: update the comment in vm/context.ts with the new function name (otherwise someone reading the comment could end up not finding the function)

PR #1276, opened 2015-03-19
Add a new profiling mode that can be used on the phone.

mbebenita:
Use: `RELEASE=1 PROFILE=2 make app`. This will save a file called `profile.txt` to the `/sdcard/downloads/` directory which you can pull and view with adb using: `adb pull /sdcard/downloads/profile.txt`

`PROFILE=1` has the same effect it did before, showing a flame chart UI but this was not very convenient on the phone.


mykmelez:
Handy, thanks!

marco-c:
As a followup, we should add some info to the profiling section of the README file (basically what @mbebenita said in the first comment)

PR #1278, opened 2015-03-19
Collect profiling stats.

mbebenita:
This PR prints aggregated profiling stats followed by detailed events. It also filters out events whose elapsed time is under a certain threshold.


mykmelez:
Nice. I'll find this useful on desktop too.

PR #1281, opened 2015-03-20
Don't override ResourceInputStream::read and ResourceInputStream::availa...

marco-c:
...ble

This will avoid the slow *getField* calls.

@mbebenita saw these being called > 80000 times on startup.


mykmelez:
What's the perf win here?

marco-c:
*getField* is slower than using the new binding system, because it needs to split the key string into name and signature (and 'I' or 'S' for instance or static) and then convert them to UTF8 and iterate over the fields until it finds the right one.

mykmelez:
> getField is slower than using the new binding system, because it needs to split the key string into name and signature (and 'I' or 'S' for instance or static) and then convert them to UTF8 and iterate over the fields until it finds the right one.

Yes, I understand the theoretical win, but I'd like to know if this has a practical impact on an actual measurement.

marco-c:
I've measured on desktop:
startupTime |       1,993ms | 1,902ms | -91ms | -4.58% | INSIGNIFICANT | 1,748ms | 2,270ms

marco-c:
I'm writing a benchmark for ResourceInputStream.

marco-c:
~200 ms for both *available* and *read single* before, ~40 ms for *available* and ~15 ms for *read single* after

mykmelez:
Thanks. It's what we expected, but it's good to have the data to back up our assumptions.

PR #1285, opened 2015-03-20
Remove unused localization strings

marco-c:


mykmelez:
Is this just a package size win, or does it improve startup perf too?

I also wonder why there are so many strings we don't use. Are we missing parts of the platform? Do midlets themselves ever access these strings?

marco-c:
> Is this just a package size win, or does it improve startup perf too?

This should reduce both memory consumption (because we have less strings to store) and startup time (because the JSON file we need to parse is smaller).

> I also wonder why there are so many strings we don't use. Are we missing parts of the platform?

Yeah, some strings were used in Chameleon, others are never used in general in the reference implementation. There are probably a few that we may need if we ever import other classes from the reference implementation, but then we can just re-add the strings we need.

> Do midlets themselves ever access these strings?

No, these strings are only used in the platform classes.

PR #1288, opened 2015-03-20
Remove some unused classes

marco-c:


mykmelez:
Theoretically a good idea. Less to ship and load. I only wonder about the public classes. Could midlets be using them? *LayerManager* at least is documented by both [Oracle](http://docs.oracle.com/javame/config/cldc/ref-impl/midp2.0/jsr118/javax/microedition/lcdui/game/LayerManager.html) and [RIM](http://www.blackberry.com/developers/docs/7.0.0api/javax/microedition/lcdui/game/LayerManager.html). And discussed in [Nokia forums](http://developer.nokia.com/community/discussion/showthread.php/24616-Layer-manager-method-in-MIDP-2-0).

marco-c:
I've inspected a few MIDlets, and none were using these classes.

PR #1289, opened 2015-03-20
Use self hosted java strings.

brendandahl:
I tried not to remove too many of our custom implementations, but the UTF_8_Reader has some nasty bug that I couldn't work around.

There were a few tests testing "UTF8" encoding. In our implementation there is not a UTF8_Writer only a UTF_8_WRITER so these tests were failing. There's some commented out code in `Helper.internalNameForEncoding` that may deal with this, but it's also commented out in the reference implementation.

On flame:

| startupTime   |      13,598ms | 12,176ms | -1,422ms | -10.46 |        BETTER | 11,001ms | 13,324ms | 
| totalSize     |      27,456kb | 28,577kb |  1,121kb |   4.08 |         WORSE | 26,804kb | 29,545kb | 
| domSize       |          92kb |     92kb |      0kb |   0.42 |         WORSE |     92kb |     92kb | 
| styleSize     |         281kb |    281kb |      0kb |   0.06 | INSIGNIFICANT |    280kb |    282kb | 
| jsObjectsSize |      16,003kb | 16,371kb |    368kb |   2.30 |         WORSE | 16,365kb | 16,378kb | 
| jsStringsSize |         621kb |    592kb |    -28kb |  -4.57 |        BETTER |    592kb |    592kb | 
| jsOtherSize   |      10,267kb | 11,046kb |    779kb |   7.59 | INSIGNIFICANT |  9,268kb | 12,015kb | 
| otherSize     |         193kb |    194kb |      1kb |   0.64 | INSIGNIFICANT |    191kb |    206kb |


mykmelez:
The startup time win looks great. It's unfortunate that we trade off some memory usage to get it. Is that expected and necessary?

> I tried not to remove too many of our custom implementations, but the UTF_8_Reader has some nasty bug that I couldn't work around.

We can remove custom implementations, we should just confirm that doing so doesn't regress perf, since we added each one to address a particular perf issue when running the original implementation in the old VM (which issue may not exist with the new VM, but we should confirm that).

marco-c:
I was expecting the jsStringSize to reduce.
What MIDlet have you executed in your measurements? A real world one or a test? 28 MB seems very low.

Have you tried removing the cache for decoded strings?

brendandahl:
> The startup time win looks great. It's unfortunate that we trade off some memory usage to get it. Is that expected and necessary?

I'm not exactly sure why it's worse, it could be that before duplicated strings were just stored as one inside the js engine but are now duplicated in the typed arrays. Or it could be a large amount of the strings are for supporting the VM so we still convert them at some point. @mbebenita Thoughts?
I also still don't trust the memory benchmark.

> I was expecting the jsStringSize to reduce.

It did somewhat. It could be the majority of the strings are more for the VM so this didn't really reduce.

> Have you tried removing the cache for decoded strings?

I'm playing with this now, and on my initial testing, it is faster not to have decoded strings, but jsstrings only goes down by 32K(as opposed to the 28K above).

PR #1290, opened 2015-03-21
Inline fast paths for checkCast and instanceOf.

mbebenita:


jit/baseline.ts
```diff
@@ -1030,13 +1030,22 @@ module J2ME {
     }
 
     emitCheckCast(cpi: number) {
-      var object = this.peek(Kind.Reference);
+      var object = this.peek(Kind.Reference)
```
mykmelez: Nit: trailing semicolon.

mykmelez:
What kind of an impact do you expect this to have?

mbebenita:
The SM baseline compiler doesn't inline function calls. I suspect most of our code stays in the baseline JIT and thus never benefits from inlining. This patch speeds up type checks quite a bit by inlining th efast path which from experiments is triggered 70%-80% of the time. However I haven't noticed much of an improvement in perf on startup, but in theory it is a win.

mykmelez:
Hmm, indeed, I don't see an impact on startup, neither on desktop nor device. I also don't see a runtime boost, but we don't have good tests for that, so it's hard to say for sure. Nevertheless, there's no negative impact either, and this is a small, well-contained patch, so I suppose it's worth taking for the theoretical win.

PR #1291, opened 2015-03-21
extend wait timeout for tests

mykmelez:
The "basic tests" take 60+ seconds to run now that we've added more tests, and we previously set CasperJS's *waitTimeout* to 70 seconds, so tests time out sometimes. We should increase the timeout to provide more cushion for test anomalies. This PR sets it to 90 seconds.

On the other hand: the less cushion, the more likely we are to notice perf regressions that make the tests run longer, like the one I found in #1273. So I'm a bit ambivalent about this change. If we take it, we should make sure to monitor test run times closely. Ideally, we'd graph them over time.

### DIFF
--- a/tests/automation.js
+++ b/tests/automation.js
@@ -16,7 +16,7 @@ casper.on('remote.message', function(message) {
     this.echo(message);
 });
 
-casper.options.waitTimeout = 70000;
+casper.options.waitTimeout = 90000;
 casper.options.verbose = true;
 casper.options.logLevel = "debug";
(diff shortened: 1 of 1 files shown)


marco-c:
I think we should take this, we shouldn't rely on the time it takes to run the tests to measure perf (even if in some cases it's useful, as in #1273).
Maybe we could set it to a smaller value (e.g. 80 seconds) and then raise it when necessary.

mykmelez:
Sure, timeout reduced to 80 seconds in latest commit on this branch.

PR #1292, opened 2015-03-21
add benchmark for System.out.print(ln)

mykmelez:
This adds a benchmark for *System.out.println* and *System.out.print*, which print characters to the console one at a time.

### DIFF
--- a/bench/SystemOutBench.java
+++ b/bench/SystemOutBench.java
@@ -10,16 +10,18 @@ public class SystemOutBench {
     for (int i = 0; i < 1000; i++) {
       System.out.println("I am the very model of a modern major general.");
     }
+    time = JVM.monotonicTimeMillis() - start;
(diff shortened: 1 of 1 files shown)


bench/SystemOutBench.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class SystemOutBench {
+  void runBenchmark() {
+    long start, time;
+
+    start = JVM.monotonicTimeMillis();
+    for (int i = 0; i < 1000; i++) {
+      System.out.println("I am the very model of a modern major general.");
+    }
+
```
marco-c: You could print a result for println here and a different one for print later, so the output would be:
println: X ms
print: Y ms

mykmelez: > You could print a result for println here and a different one for print later, so the output would be:

Sure, the latest commit does that.

PR #1293, opened 2015-03-21
WIP: emit Klass constructor as closure

mykmelez:
*emitKlassConstructor* appears expensive in profiles, and it looks like that's because it dynamically creates and evals the Klass constructor. There's a note in the function about using a closure to improve startup perf, potentially at the expense of runtime perf.

And here's one implementation of that approach. I haven't yet measured the impact on runtime perf, nor have I measured on device yet, but startup perf and memory consumption appear better on desktop with a midlet (although the numbers aren't consistent; some benchmark runs show no improvement):

| startupTime   |       1,286ms |  1,229ms |    -57ms |  -4.43 |        BETTER |  1,189ms |  1,281ms | 
| totalSize     |      34,209kb | 32,968kb | -1,241kb |  -3.63 |        BETTER | 31,141kb | 34,346kb | 
| domSize       |         193kb |    193kb |      0kb |   0.00 | INSIGNIFICANT |    193kb |    193kb | 
| styleSize     |         433kb |    432kb |      0kb |  -0.02 | INSIGNIFICANT |    432kb |    433kb | 
| jsObjectsSize |      17,468kb | 17,363kb |   -104kb |  -0.60 |        BETTER | 17,362kb | 17,368kb | 
| jsStringsSize |         675kb |    580kb |    -95kb | -14.08 |        BETTER |    580kb |    580kb | 
| jsOtherSize   |      15,009kb | 13,969kb | -1,041kb |  -6.93 | INSIGNIFICANT | 12,136kb | 15,347kb | 
| otherSize     |         432kb |    430kb |     -1kb |  -0.34 | INSIGNIFICANT |    430kb |    431kb | 

To try to mitigate the impact on runtime perf, this implementation caches the default values of fields in their FieldInfo objects.

There may be a better way to emit these constructors. In particular, it might help to get them off the global object. *registerKlassSymbol* already attaches them to the RuntimeTemplate in addition to the global object, so perhaps we could always access them from there. Or maybe we could even create a meta-constructor function that creates and returns an instance of any type, given the mangledName as an argument (for which we'd need a mangledName -> ClassInfo lookup table).


mbebenita:
@mykmelez see https://bugzilla.mozilla.org/show_bug.cgi?id=1146022, in particular look at the 4th implementation approach.

mykmelez:

PR #1295, opened 2015-03-21
Fix UTF_8_Writer::sizeOf bug

marco-c:


marco-c:
Looks like this function is never used, this is why we haven't seen it failing.

tests/com/sun/cldc/i18n/UTF_8_Writer_sizeOf.java
```diff
@@ -0,0 +1,35 @@
...
+        str.getChars(0, str.length(), cbuf, 0);
+		    
+		    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        StreamWriter sw = new UTF_8_Writer();
+
+        sw.open(baos, "UTF_8");
+
+        th.check(sw.sizeOf(cbuf, 0, cbuf.length), 84);
+
+    		sw.close ();
```
mykmelez: Nit: there are some tabs on lines 23, 30, and 32 of this file; and this line has an extra space in between the method name and its argument list.

tests/com/sun/cldc/i18n/UTF_8_Writer_sizeOf.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class UTF_8_Writer_sizeOf implements Testlet {
+  public int getExpectedPass() { return 1; }
+  public int getExpectedFail() { return 0; }
+  public int getExpectedKnownFail() { return 0; }
+
+  public void test (TestHarness th)
+  {
+    try {
+        String str = "abcdefghilmnopqrstuvzABCDEFGHILMNOPQRSTUVZabcdefghilmnopqrstuvzABCDEFGHILMNOPQRSTUVZ";
```
mykmelez: Nit: inconsistent indentation: two characters on line 18, another four on line 19.

Does this test come from another project by any chance?

marco-c: No, I just copied the "skeleton" from another file, and this caused some problems with the indentation.

brendandahl:
> Looks like this function is never used, this is why we haven't seen it failing.

I'm curious why we have a custom native implementation and benchmarks for this if it doesn't seem to be used?

marco-c:
> I'm curious why we have a custom native implementation and benchmarks for this if it doesn't seem to be used?

I think we can remove the sizeOf benchmark.
I guess it was ported to a native when we ported the other UTF_8_Writer function, since it's basically the same code (so porting it didn't cost anything).

I'll remove the benchmark in this PR.

marco-c:
Actually, we don't have a benchmark for this function.

PR #1300, opened 2015-03-23
benchmark FileConnection memory usage

mykmelez:
This is a very basic benchmark for measuring FileConnection memory usage. It requires you to use about:config to set the boolean preference *security.turn_off_all_security_so_that_viruses_can_take_over_this_computer* to `true`.

The benchmark simply creates a file and writes 1MB of random data to it, reporting memory usage before and after. I see either ~2MB or ~4MB increase in memory usage across the operation, although usage drops back down if I manually check again a second later.


bench/FileConnectionBench.java
```diff
@@ -0,0 +1,48 @@
...
+public class FileConnectionBench {
+  void runBenchmark() {
+    try {
+      long start, time;
+
+      start = JVM.monotonicTimeMillis();
+
+      String privateDir = System.getProperty("fileconn.dir.private");
+      Random random = new Random();
+      String filename = new javax.bluetooth.UUID(java.lang.Math.abs(random.nextInt())).toString();
```
marco-c: javax.bluetooth.UUID only exists if you build with JSR_082=1, could you use another name?
Even a static name would work, with a:
```Java
        if (file.exists()) {
            file.delete();
        }
```

bench/FileConnectionBench.java
```diff
@@ -0,0 +1,48 @@
...
+      FileConnection file = (FileConnection)Connector.open(privateDir + filename);
+      System.out.println("Writing to file " + privateDir + filename);
+      file.create();
+
+      SecureRandom rnd = SecureRandom.getInstance(SecureRandom.ALG_SECURE_RANDOM);
+      OutputStream out = file.openOutputStream();
+      byte[] b = new byte[1024];
+      for (int i = 0; i < 1000; i++) {
+        rnd.nextBytes(b, 0, 1024);
```
marco-c: The secure random algorithm is going to use MD5 to generate random data, so this benchmark might be dominated by the time spent executing the hash.
You could create a static array before the benchmark and directly use it.

bench/FileConnectionBench.java
```diff
@@ -0,0 +1,48 @@
...
+        out.write(b);
+      }
+      out.close();
+
+      time = JVM.monotonicTimeMillis() - start;
+      System.out.println("FileConnection.write time: " + time);
```
marco-c: Could you split this time in *open* and *write*?

tests/native.js
```diff
@@ -163,3 +163,56 @@ Native["com/sun/midp/i18n/TestResourceConstants.setLanguage.(Ljava/lang/String;)
 // Many tests create FileConnection objects to files with the "/" root,
 // so add it to the list of valid roots.
 MIDP.fsRoots.push("/");
+
+var NO_SECURITY = typeof netscape !== "undefined" && netscape.security.PrivilegeManager;
```
marco-c: It could be worth moving this in a separate script to avoid duplicating code.

bench/FileConnectionBench.java
mykmelez: I originally did that, but I actually want to create unique files each time, so I can see if database growth over time affects memory usage. In 549e3e9 I switch to using *System.currentTimeMillis* to create unique-enough filenames.

bench/FileConnectionBench.java
mykmelez: This is more of a memory benchmark. Nevertheless, I do report elapsed time, so it makes sense to focus that measurement on the subject of the benchmark. (And in any case I want to exclude the memory used by the random data generator from the memory benchmark.) In 5979324 I make that change.

bench/FileConnectionBench.java
mykmelez: Yes, in 743084a I distinguish between the *open/create*, *openOutputStream*, and *write* times.

tests/native.js
mykmelez: Indeed. In b60f0b3, I move it back to benchmark.js, refactoring *sampleMemory* in that file so that *MemorySampler.sampleMemory* can use it from tests/native.js. I hear @brendandahl is changing the way we sample memory, so that function might change further, but in the meantime this means we aren't duplicating its functionality.

Note that you now have to `make BENCHMARK=1` to see memory results. Without *BENCHMARK*, you'll only see timings.

mykmelez:
Two more things we can do, either in this branch or a followup:

1. Make the memory output prettier.
2. Sample memory again after a timeout, to see if it drops back down.

(We could also time *OutputStream.close* and *FileConnection.close*, although I don't know of any potential issues with those methods.)

mykmelez:
> Make the memory output prettier.
> Sample memory again after a timeout, to see if it drops back down.

74056da and e2cc696 do this.

benchmark.js
```diff
@@ -179,36 +215,12 @@ var Benchmark = (function() {
...
-        storage.current.styleSize.push(styleSize.value);
-        storage.current.jsObjectsSize.push(jsObjectsSize.value);
-        storage.current.jsStringsSize.push(jsStringsSize.value);
-        storage.current.jsOtherSize.push(jsOtherSize.value);
-        storage.current.otherSize.push(otherSize.value);
-        saveStorage();
+    sampleMemoryToStorage: function() {
+      var mem = sampleMemory();
+      for (var p in mem) {
+        storage.current[p] = mem[p];
```
brendandahl: This value should be pushed to the array no replace it. I have fixed this in my upcoming changes.

mykmelez: D'oh, sorry about the bustage! I requested a pull for a fix in https://github.com/mozilla/j2me.js/pull/1308 while we await your upcoming changes.

Issue #1234, opened 2015-03-13
Overrides prevent optimizations in lazy linking.


mbebenita:
We can delay linking in the interpreter if we know the target of a virtual/static call. At the moment there can be 3 possible targets, interpreted methods, native methods, or overrides. Native targets we have to link right away, but this is easy to check for since we can just check the access flag, for all other methods we could just continue in the interpreter and avoid linking. Unfortunately, since any method can be overridden, we have to convert the method name to a JS string, generate an impl key, check against our overrides, etc.

I propose we mark all overridden methods as native in the Java sources, and get rid of the override mechanism, or am I missing something?

andreasgal:
We are trying to avoid changing the original Java sources. However, we could write a build tool to do this if we must do it.

mbebenita:
We have 98 overrides, the majority of which (74) appear in `string.js`. If we reland #1189, then changing the remaining sources wouldn't be a big effort.

brendandahl:
> We are trying to avoid changing the original Java sources. However, we could write a build tool to do this if we must do it.

We have a way to do this by copying the file into our custom folder. Our lives would be simplified further if we just changed the original sources though. We'll always have the originals in the repo history if we ever need to diff them.

@andreasgal Is there a specific reason we are avoiding changing the java sources?

marco-c:
> We have a way to do this by copying the file into our custom folder. Our lives would be simplified further if we just changed the original sources though. We'll always have the originals in the repo history if we ever need to diff them.

I think our current mechanism with the custom/ subdirectory is better. It's easier to compare the custom implementations with the reference implementation and it's also easy to have a list of all the classes we have changed.

> Is there a specific reason we are avoiding changing the java sources?

We're doing this to ensure correctness.

brendandahl:
It sure makes reviewing new customs much harder and it creates two files which is much more error prone if you're looking at the wrong one or if you misname/move it incorrectly.  We can easily look back in logs, blame, or diff to see the original state. We have a revision control system, why not use it?

> We're doing this to ensure correctness.

Sorry, I meant this as more why do we have a custom directory. I agree we should avoiding changing them unless necessary.

mykmelez:
> It sure makes reviewing new customs much harder and it creates two files which is much more error prone if you're looking at the wrong one or if you misname/move it incorrectly. We can easily look back in logs, blame, or diff to see the original state. We have a revision control system, why not use it?

[...]

PR #1301, opened 2015-03-23
Remove all overrides but the strings.js ones

marco-c:
The overridden version of Math.min is slower than the self-hosted version (~35 ms vs ~15 ms)
The overridden version of the DefaultCaseConverter functions are slower than the self-hosted versions (~60ms vs ~30ms)


README.md
```diff
@@ -241,14 +241,6 @@ e.g.:
 
     Native["java/lang/System.arraycopy.(Ljava/lang/Object;ILjava/lang/Object;II)V" = function(src, srcOffset, dst, dstOffset, length) {...};
 
-If you need to implement a method in JS but you can't declare it `native` in Java, use `Override`.
```
yxl: I'd rather keep the `override` functionality, but limit its usage.
It's  really helpful for hooking method when debugging a black-boxed midlet.

java/midp/com/sun/midp/security/Permissions.java
```diff
@@ -1,1230 +0,0 @@
-/*
```
yxl: I think we want to keep this file, though you add a customized one under the `java/custom` directory.

java/custom/com/sun/midp/security/Permissions.java
```diff
@@ -0,0 +1,294 @@
...
+     * Create a list of permission groups a domain is permitted to perform.
+     *
+     * @param name binding name of domain
+     *
+     * @return 2 arrays, the first containing the maximum level for each
+     *     permission, the second containing the default or starting level
+     *     for each permission supported
+     */
+    public static byte[][] forDomain(String name) {
+        NUMBER_OF_PERMISSIONS = PermissionsStrings.PERMISSION_STRINGS.length + 2;
```
yxl: It'd better to keep all the original comments about how to calculate the permission number:
// The 2 is the two hardcoded MIPS and AMS permissions.

midp/gfx.js
```diff
@@ -396,10 +396,6 @@ var currentlyFocusedTextEditor;
         return defaultFont;
     }
 
-    Override["javax/microedition/lcdui/Font.getDefaultFont.()Ljavax/microedition/lcdui/Font;"] = function() {
```
yxl: If we remove `getDefaultFont` override,  we end up with two `getDefaultFont()`. One in this file gfx.js and one in Font.java. They will create two default font instance, but actually we only need one of them.

java/custom/java/io/ByteArrayInputStream.java
```diff
@@ -83,6 +83,8 @@
      */
     protected int count;
 
+    native void init(byte buf[], int offset, int length);
```
yxl: Good! It is concise and make the code more readable.

java/midp/com/sun/midp/security/Permissions.java
marco-c: We've decided to *git mv* the files in java/custom/ instead of copying them in #1234 (there's a discussion in the issue to explain the reasons).

README.md
marco-c: Unfortunately in order to fix #1234 and enable optimizations in lazy linking, we need to get rid of overrides.
To hook methods while debugging, we could do:
1) In vm/parser.ts, in MethodInfo::isNative return true for the method we want to hook
2) Create a native in native.js
I think this should work, but supporting it in a release build is going to slow us down. We could support it only on debug builds.
I'm going to try and implement this approach in a follow-up (since in this PR I'm not yet removing overrides).

java/custom/com/sun/midp/security/Permissions.java
marco-c: Done in b27d8c8.

midp/gfx.js
marco-c: Fixed in d5a3236.

README.md
yxl: That will be nice:)

PR #1303, opened 2015-03-23
Remove some code from the startup path

marco-c:
This is a first step, there's still some code that we might be able to remove (there are a lot of event producers, event listeners, proxies, etc.)


mykmelez:
Hmm, getting a diff of the changes from an original file to its custom is a bit more of a pain than I'd realized, especially if the file was moved and customized in the same commit. Here's one way to do it:

```
FILE=java/custom/com/sun/midp/main/CldcMIDletSuiteLoader.java
TAIL=`git log --follow --format=%H ${FILE} | tail -n 1`
git diff --follow ${TAIL}...HEAD ${FILE}
```

PR #1307, opened 2015-03-24
WIP: Remove support for overrides

marco-c:
Depends on #1301 and #1314.


marco-c:
Need to wait for the other two PRs to land and then fix the test result (I suspect it isn't due to this PR, but to the combined effect of the other two PRs)

brendandahl:
We may not want to wait for the self hosted strings patch to do this. I played a bit with it to see if I could get performance back to what it was, but there was nothing immediately noticeable that was wrong.

marco-c:
The problem is that we can't completely remove support for overrides unless we remove the String overrides.

brendandahl:
Can't we just convert it to custom and make all the overrides native?

marco-c:
Ah yes, sure. I'll do that.

marco-c:
Obsoleted by https://github.com/mozilla/j2me.js/pull/1391.

PR #1312, opened 2015-03-25
Use Soot to optimize classes.jar bytecode

marco-c:
From https://github.com/Sable/soot/wiki/Using-Soot-as-a-Program-Optimizer:
*Soot is able to optimize individual classfiles. Some of the transformations which can be carried out on individual classfiles include: common subexpression elimination, partial redundancy elimination, copy propagation, constant propagation and folding, conditional branch folding, dead assignment elimination, unreachable code elimination, unconditional branch folding, and unused local elimination.*

Using Soot does remove 20 kB of code, but it doesn't seem to improve startup time significantly (on Desktop).


marco-c:
Looks like when we exit the com/sun/cldc/util/j2me/TimeZoneImpl.getInstance.(Ljava/lang/String;)Ljava/util/TimeZone; method, the lock on the object is undefined.

This is the code generated by the baseline compiler:
```JavaScript
var th=$.ctx.thread;
var re,lk,k0=$G_INu3hj;
var a,b;
a=this;
var A,B;
var pc=0;
if(O){
  var _=O.local;
  a=_[0],b=_[1];
  pc=O.pc;
  O=null;
}else{
  lk=a._lock;
  if(lk&&lk.level===0){lk.thread=th;lk.level=1;}else{ME(a);
    if(U){$.B(0,0,[a,b],[],a);return;}
  }
}
try{
  var label=0;
  // Entry Dispatch Block
  if (pc===0) {
  label = 1;
  } else if (!(pc===35)) {
  J2ME.Debug.assert(false, 'Invalid OSR PC: ' + pc)
  }
  if (label == 1) {
  if (!(b!==null)) {
  if (!(($.$G_INu3hj.s0)!==null)) {
  if ($.initialized["java/lang/System"] === undefined) { $.$GBEt6NdQ.classObject.initialize(); }
  A=k0.c(4);
  U&&B1(12,12);
  re=$GBEt6NdQ.m(4)(A);
  U&&B0(12);
  $.$G_INu3hj.s0=re;
  if (!(($.$G_INu3hj.s0)!==null)) {
  $.$G_INu3hj.s0=k0.c(136);
  }
  }
  b=$.$G_INu3hj.s0;
  }
  a=0;
  }
  while(1) {
  if (a>=(($.$G_INu3hj.s30).length)) {
  label = 11;
  break;
  }
  if ((a >>> 0) >= ($.$G_INu3hj.s30.length >>> 0)) TI(a);
  CCK($.$G_INu3hj.s30[a],k0);
  re=(($.$G_INu3hj.s30[a]).f0).v2(b);
  if (U && re !== undefined) { J2ME.Debug.assert(false, 'Unexpected return value during unwind.'); }
  if(U){J2ME.Debug.assert(false,'Unexpected unwind.');}
  if (!(re===0)) {
  label = 9;
  break;
  }
  a+=1;
  }
  if (label == 9) {
  if ((a >>> 0) >= ($.$G_INu3hj.s30.length >>> 0)) TI(a);
  if(a._lock.level===1&&a._lock.ready.length===0)a._lock.level=0;else MX(a);
  return ($.$G_INu3hj.s30[a]);
  }
  else if (label == 11) {
  re=k0.m(10)(b);
  U&&B0(74);
  if(a._lock.level===1&&a._lock.ready.length===0)a._lock.level=0;else MX(a);
  return re;
  }
  J2ME.Debug.assert(false, 'Invalid PC: ' + pc)
}catch(ex){
  if(U){$.T(ex,[a,b],[A,B],a);return;}
  A=TE(ex);
  if(a._lock.level===1&&a._lock.ready.length===0)a._lock.level=0;else MX(a);
  throw A;
}
```

marco-c:
Looks like the identifier *a* is used both to hold a reference to *this* and as a number.

marco-c:
I think it's a Soot issue, filed as https://github.com/Sable/soot/issues/376

marco-c:
Soot also reduces the number of local variables that functions use, so it should have a good impact on memory.

mbebenita:
Does soot do any inlining, that would be nice?

marco-c:
> Does soot do any inlining, that would be nice?

It does inlining, but only in whole-program mode (that is not enabled in this PR).
The problem with whole-program optimization is that it needs the program (in our case MIDlet) to see what functions can be inlined.

marco-c:
| startupTime   |      15,443ms | 15,486ms |     43ms |   0.28 | INSIGNIFICANT | 13,874ms | 17,823ms | 
| totalSize     |      31,387kb | 30,093kb | -1,294kb |  -4.12 |        BETTER | 25,285kb | 31,797kb | 
| domSize       |          93kb |     93kb |      0kb |  -0.20 |        BETTER |     92kb |     93kb | 
| styleSize     |         328kb |    328kb |      0kb |  -0.03 | INSIGNIFICANT |    327kb |    330kb | 
| jsObjectsSize |      16,317kb | 16,440kb |    123kb |   0.75 |         WORSE | 16,080kb | 16,942kb | 
| jsStringsSize |       1,632kb |  1,075kb |   -556kb | -34.10 |        BETTER |  1,055kb |  1,082kb | 
| jsOtherSize   |      12,620kb | 11,760kb |   -860kb |  -6.81 | INSIGNIFICANT |  6,796kb | 13,680kb | 
| otherSize     |         397kb |    396kb |      0kb |  -0.05 | INSIGNIFICANT |    380kb |    422kb |

java/Makefile
```diff
@@ -26,6 +26,9 @@ classes.jar: $(SRCS) $(JPP_DESTS) $(EXTRA)
 	$(foreach dir,$(SRC_DIRS),cp -a $(dir)/. build-src/;)
 	find ./build-src -name *.java > build-srcs.txt
 	javac -cp build-src -g:none -source 1.3 -target 1.3 -bootclasspath "" -extdirs "" -d ./build @build-srcs.txt > /dev/null
+	java -jar ../build_tools/soot-trunk.jar -j2me -process-dir build -no-output-source-file-attribute -no-output-inner-classes-attribute -force-overwrite -include-all 2>&1 > /dev/null
+	cp -a sootOutput/. build/
+	rm -rf sootOutput
```
mykmelez: We remove the build and build-src directories too, as they're intermediate products, but I think we should actually leave these dirs around (with entries in .gitignore so they don't show up as untracked files), as it's occasionally useful to inspect them while troubleshooting.

marco-c: Done in 46f5e2d, the files in these directories are already ignored (because in .gitignore we have a rule for java/build-src/ and a rule for *.class)

mykmelez: > Done in 46f5e2d, the files in these directories are already ignored (because in .gitignore we have a rule for java/build-src/ and a rule for *.class)

Looks good, thanks!

PR #1313, opened 2015-03-25
Use minimize memory approach and gc in new page between rounds.

brendandahl:
I've copied what the minimize memory usage button in about:memory does. Gives much more consistent results on device.  Desktop is also more consistent but whipsaws more.

flame _totalsize
flame _startuptime

desktop _totalsize
desktop _startuptime
Above: There's a spike on the new way, but I seem to have got a bit unlucky on this round because I have only seen that once in many runs.

Below: Running the new code but with a 1mb allocation in main.js.
desktop 1mb _totalsize
desktop 1mb _startuptime

### DIFF
--- a/benchmark.js
+++ b/benchmark.js
@@ -29,12 +29,20 @@ var Benchmark = (function() {
   }
 
   function forceCollectors() {
-    enableSuperPowers();
-    console.log("Forcing CC/GC.");
-    for (var i = 0; i < 3; i++) {
-      Components.utils.forceCC();
-      Components.utils.forceGC();
+    if (!NO_SECURITY) {
+      return Promise.resolve();
     }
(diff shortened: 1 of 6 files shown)


brendandahl:
Just remembered I probably broke the new benchmark Myk added. Let me fix that too.

mykmelez:
Looks good. Interesting that it reports significantly less of an increase in usage in the FileConnection benchmark. It would also be useful to measure memory usage with GC/CC, to look for spikes in usage from implementations that may not leak but are inefficient (which I suspect in the case of FileConnection).

mykmelez:
Erm, in that previous comment, I meant to say that it would be useful to measure memory usage *without* GC/CC to look for usage spikes.

PR #1315, opened 2015-03-26
Limit refresh rate to that of screen. Update tests

TimAbraldes:
This prevents us from attempting to paint multiple times to the screen
when the screen can't actually keep up due to its refresh rate.

Update the asteroids test to use the accepted mechanism for doing 1
calculation per frame.

Add a benchmark test for `Graphics.drawRegion` performance.

### DIFF
--- a/bench/GraphicsBench.java
+++ b/bench/GraphicsBench.java
@@ -1,0 +1,239 @@
+package benchmark;
+
+import com.sun.cldchi.jvm.JVM;
+
+import javax.microedition.lcdui.*;
+import javax.microedition.lcdui.game.Sprite;
+import javax.microedition.midlet.*;
+
+public class GraphicsBench extends MIDlet {
+  interface TestInfo {
+    String getDescription();
+    void executeFrame(Graphics g, long numOps);
+  }
+
+  class TestCanvas extends Canvas implements Runnable {
+    TestCanvas(Display d) {
+      display = d;
+
+      Image img = Image.createImage(getWidth(), getHeight());
+      offscreenBuffer = img.getGraphics();
+
+      // NOTE: When adding new tests, append them to the end
+      //       of this array
+      tests = new TestInfo[2];
+      tests[0] = new DrawRegionTransNoneCenter(getWidth(), getHeight());
+      tests[1] = new DrawStringBaselineCenter(getWidth(), getHeight());
+
+      currentTest = tests[0];
+    }
+
+    protected void paint(Graphics screenG) {
+      if (currentTest != null) {
+        if (isOffscreen) {
+          long frameStartTime = JVM.monotonicTimeMillis();
+          currentTest.executeFrame(offscreenBuffer, numOpsPerFrame);
+          lastFrameTime = JVM.monotonicTimeMillis() - frameStartTime;
+        } else {
+          long frameStartTime = JVM.monotonicTimeMillis();
+          currentTest.executeFrame(screenG, numOpsPerFrame);
+          lastFrameTime = JVM.monotonicTimeMillis() - frameStartTime;
+        }
+      }
+    }
+
+    void startNextTest() {
+      if (isOffscreen) {
+        isOffscreen = false;
+        if (nextTestIndex < tests.length) {
+          currentTest = tests[nextTestIndex];
+          System.out.println(currentTest.getDescription());
+          nextTestIndex++;
+        } else {
+          System.out.println("DONE");
+          return;
+        }
+      } else {
+        isOffscreen = true;
+        System.out.println("Offscreen " + currentTest.getDescription());
+      }
+
+      repaint();
+      display.callSerially(this);
+    }
+
+    protected void showNotify() {
+      startNextTest();
+    }
+
+    public void run() {
+      totalFrameTime += lastFrameTime;
+      numSamples++;
+      if (numSamples < targetSamples) {
+        repaint();
+        display.callSerially(this);
+        return;
+      }
+
+      avgFrameTime = (totalFrameTime * 1000) / numSamples;
+      System.out.println("\tOps per second: " + ((numOpsPerFrame * 1000000) / avgFrameTime));
+      numSamples = 0;
+      totalFrameTime = 0;
+
+      startNextTest();
+    }
+
+    private final long targetSamples = 240;
+    private final long numOpsPerFrame = 500;
+    private long lastFrameTime = 0;
+    private long totalFrameTime = 0;
+    private long avgFrameTime = 0;
+    private long numSamples = 0;
+    private TestInfo[] tests;
+    private TestInfo currentTest;
+    private int nextTestIndex = 0;
+    private Display display;
+    private Graphics offscreenBuffer;
+    private boolean isOffscreen = true;
+  }
+
+  public void startApp() {
+    Display d = Display.getDisplay(this);
+    TestCanvas test = new TestCanvas(d);
+    test.setFullScreenMode(true);
+    d.setCurrent(test);
+  }
+
+
+
+  //
+  // Implement actual tests below
+  //
+
+  class DrawRegionTransNoneCenter implements TestInfo {
+    DrawRegionTransNoneCenter(int w, int h) {
+      width = w;
+      height = h;
+
+      int rectWidth = width / 2;
+      int rectHeight = height / 2;
+
+      srcImages = new Image[2];
+
+      srcImages[0] = Image.createImage(width, height);
+      Graphics srcG = srcImages[0].getGraphics();
+      srcG.setColor(255, 0, 0);
+      srcG.fillRect(0, 0, rectWidth, rectHeight);
+      srcG.setColor(0, 255, 0);
+      srcG.fillRect(rectWidth, 0, rectWidth, rectHeight);
+      srcG.setColor(0, 0, 255);
+      srcG.fillRect(rectWidth, rectHeight, rectWidth, rectHeight);
+      srcG.setColor(255, 255, 0);
+      srcG.fillRect(0, rectHeight, rectWidth, rectHeight);
+
+
+      srcImages[1] = Image.createImage(width, height);
+      Graphics src2G = srcImages[1].getGraphics();
+      src2G.setColor(0, 255, 255);
+      src2G.fillRect(0, 0, rectWidth, rectHeight);
+      src2G.setColor(255, 255, 255);
+      src2G.fillRect(rectWidth, 0, rectWidth, rectHeight);
+      src2G.setColor(0, 0, 0);
+      src2G.fillRect(rectWidth, rectHeight, rectWidth, rectHeight);
+      src2G.setColor(255, 0, 255);
+      src2G.fillRect(0, rectHeight, rectWidth, rectHeight);
+    }
+
+    public void executeFrame(Graphics g, long numOps) {
+      for (long j = 0; j < numOps; j++) {
+        g.drawRegion(srcImages[curImageIndex], 0, 0, width, height, Sprite.TRANS_NONE, width / 2, height / 2, Graphics.VCENTER|Graphics.HCENTER);
+        curImageIndex = (curImageIndex + 1) % srcImages.length;
+      }
+      curImageIndex = (curImageIndex + 1) % srcImages.length;
+    }
+
+    public String getDescription() {
+      return description;
+    }
+
+    private int i = 0;
+    private int width;
+    private int height;
+    private Image[] srcImages;
+    private int curImageIndex = 0;
+    private final String description = "Graphics.drawRegion TRANS_NONE centered";
+  }
+
+  class DrawStringBaselineCenter implements TestInfo {
+    DrawStringBaselineCenter(int w, int h) {
+      width = w;
+      height = h;
+    }
+
+    public void executeFrame(Graphics g, long numOps) {
+      for (int j = 0; j < numOps; j++) {
+        g.drawString(string, j%width, j%height, Graphics.BASELINE|Graphics.HCENTER);
+      }
+    }
+
+    public String getDescription() {
+      return description;
+    }
+
+    private int width;
+    private int height;
+    private final String description = "Graphics.drawString baseline centered";
+    private final String string = "String string = new String(\"string\");";
+  }
+
+  //
+  // End impl of actual tests
+  //
+
+  public void pauseApp() {
+  }
+
+  public void destroyApp(boolean unconditional) {
+  }
+
+  // Future tests to consider:
+  //
+  // ImageDataFactory.createImmutableImageDecodeImage
+  // ImageDataFactory.createImmutableImageDataRegion
+  // ImageDataFactory.createImmutableImageDataCopy
+  // ImageDataFactory.createMutableImageData
+  // ImageDataFactory.createImmutableImageDecodeRGBImage
+  // ImageData.getRGB
+  // DirectUtils.setPixels
+  // DirectGraphicsImp.getPixels
+  // DirectGraphicsImp.drawPixels
+  // Graphics.render
+  // Graphics.renderRegion
+  // Graphics.drawSubstring
+  // Graphics.drawChars
+  // Graphics.drawChar
+  // Graphics.fillTriangle
+  // Graphics.drawRect
+  // Graphics.fillRect
+  // Graphics.drawRoundRect
+  // Graphics.fillRoundRect
+  // Graphics.drawArc
+  // Graphics.fillArc
+  // Graphics.drawLine
+  // Graphics.drawRGB
+  //
+  // Font.stringWidth
+  // Font.charWidth
+  // Font.charsWidth
+  // Font.substringWidth
+  //
+  // Graphics.resetGC
+  // Graphics.reset
+  // Graphics.copyArea
+  // Graphics.setDimensions
+  // Graphics.setClip
+  // Graphics.clipRect
+  // Graphics.drawRegion
+  // Graphics.drawImage
+}
+
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -112,8 +112,32 @@ var currentlyFocusedTextEditor;
         return 1;
     };
 
+    var hasNewFrame = true;
+    var ctxs = [];
+    function gotNewFrame(timestamp) {
+        if (ctxs.length > 0) {
+            var ctx = ctxs.pop();
+            window.requestAnimationFrame(gotNewFrame);
+            ctx.execute();
+        } else {
+            hasNewFrame = true;
+        }
+    }
 
-    addUnimplementedNative("com/sun/midp/lcdui/DisplayDevice.refresh0.(IIIIII)V");
+    Native["com/sun/midp/lcdui/RepaintEventProducer.waitForAnimationFrame.()V"] = function() {
+        if (hasNewFrame) {
+            hasNewFrame = false;
+            window.requestAnimationFrame(gotNewFrame);
+        } else {
+            ctxs.unshift($.ctx);
+            $.pause(asyncImplStringAsync);
+        }
+    }
+
+    Native["com/sun/midp/lcdui/DisplayDevice.refresh0.(IIIIII)V"] = function() {
+        // This is a no-op: The foreground display gets drawn directly
+        // to the screen.
+    };
 
     function swapRB(pixel) {
         return (pixel & 0xff00ff00) | ((pixel >> 16) & 0xff) | ((pixel & 0xff) << 16);
--- a/java/custom/com/sun/midp/lcdui/RepaintEventProducer.java
+++ b/java/custom/com/sun/midp/lcdui/RepaintEventProducer.java
@@ -174,6 +174,10 @@ public class RepaintEventProducer implements EventListener {
         return true;
     }
 
+    // Called from `process` to ensure that we don't try to repaint
+    // more than once per animation frame
+    private native void waitForAnimationFrame();
+
     /**
      * Process an event.
      *
@@ -181,6 +185,8 @@ public class RepaintEventProducer implements EventListener {
      */
     public void process(Event genericEvent) {
         RepaintEvent event = (RepaintEvent)genericEvent;
+
+        waitForAnimationFrame();
 
         synchronized (this) {
             queuedEvent = null;
--- a/jit/analyze.ts
+++ b/jit/analyze.ts
@@ -25,6 +25,7 @@ module J2ME {
    * at the right spots.
    */
   export var yieldMap = {
+    "com/sun/midp/lcdui/RepaintEventProducer.waitForAnimationFrame.()V": YieldReason.Root,
     "com/sun/midp/main/MIDletSuiteUtils.vmBeginStartUp.(I)V": YieldReason.Root,
     "com/sun/midp/lcdui/DisplayDevice.gainedForeground0.(II)V": YieldReason.Root,
     "com/sun/cdc/io/j2me/file/DefaultFileHandler.openForRead.()V": YieldReason.Root,
--- a/native.js
+++ b/native.js
@@ -3,6 +3,7 @@
 
 'use strict';
 
+var asyncImplStringAsync = "Async";
 function asyncImpl(returnKind, promise) {
   var ctx = $.ctx;
 
@@ -21,7 +22,7 @@ function asyncImpl(returnKind, promise) {
     ctx.frames.push(Frame.create(methodInfo, [exception]));
     ctx.execute();
   });
-  $.pause("Async");
+  $.pause(asyncImplStringAsync);
 }
 
 var Native = {};
--- a/tests/asteroids/Field.java
+++ b/tests/asteroids/Field.java
@@ -17,6 +17,8 @@
 
 package asteroids;
 
+import com.sun.cldchi.jvm.JVM;
+
 import javax.microedition.midlet.*;
 import javax.microedition.lcdui.*;
 import java.util.*;
@@ -222,6 +224,8 @@ public class Field extends Canvas implements Runnable {
      * Overriden from Canvas.
      */
     protected void paint(Graphics g) {
+        long paintStartTime = JVM.monotonicTimeMillis();
+
         Graphics gr = (_buffer != null) ? _buffer.getGraphics() : g;
         gr.setColor(0x00FFFFFF);;
         gr.fillRect(0, 0, getWidth() - 1, getHeight() - 1);
@@ -281,6 +285,11 @@ public class Field extends Canvas implements Runnable {
         }
         if (_buffer != null) {
             g.drawImage(_buffer, 0, 0, Graphics.TOP|Graphics.LEFT);
+        }
+
+        paintavg += JVM.monotonicTimeMillis() - paintStartTime;
+        if (frames % 60 == 0) {
+            System.out.println("paintavg(x1000) = " + (1000 * Field.paintavg / Field.frames));
         }
     }
 
@@ -375,11 +384,10 @@ public class Field extends Canvas implements Runnable {
      */
     public void run() {
         try {
-            while (!_paused) {
-                long time1 = System.currentTimeMillis();
+            if (!_paused) {
+                long time1 = JVM.monotonicTimeMillis();
                 long time2;
                 long time3;
-                long ellapsed = 0;
                 if (!_frozen) {
 
                // If the field has been cleared, change the level.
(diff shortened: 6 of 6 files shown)


midp/gfx.js
```diff
@@ -112,8 +112,20 @@ var currentlyFocusedTextEditor;
...
-    addUnimplementedNative("com/sun/midp/lcdui/DisplayDevice.refresh0.(IIIIII)V");
+    Native["com/sun/midp/lcdui/DisplayDevice.refresh0.(IIIIII)V"] = function() {
+        // This is a no-op: The foreground display gets drawn directly
+        // to the screen.
+        //
+        // However, we sometimes "draw" more frequently than the
+        // screen can update. By calling `requestAnimationFrame` here,
+        // we ensure that whatever we just drew will stick around for
+        // one frame
+        asyncImpl("V", new Promise(function(resolve, reject) {
```
marco-c: Maybe in this case we shouldn't use asyncImpl (otherwise the thread would be resumed after a setZeroTimeout and after a requestAnimationFrame, but we want it to be resumed only after requestAnimationFrame).

marco-c:
It would be interesting to test whether this changeset improves the scrolling animation and whether it makes perf any worse.

marco-c:
@TimAbraldes could you merge from master? The Travis build is failing because it can't download the SpiderMonkey shell (fixed in #1318)

TimAbraldes:
> It would be interesting to test whether this changeset improves the scrolling animation and whether it makes perf any worse.

I haven't seen any difference in apps that use scrolling. I'm running startup benchmarks now and will post results when I have them

TimAbraldes:
These are the results for my favorite app. 100 rounds, finalization after 1.5s. Baseline is current master, benchmarked current state of this branch.

|      Test     | Baseline Mean |   Mean   |   +/-   |   %   |       P       |    Min   |    Max   | 
|:--------------|--------------:|---------:|--------:|------:|--------------:|---------:|---------:|
| startupTime   |       1,900ms |  1,906ms |     6ms |  0.32 | INSIGNIFICANT |  1,509ms |  2,450ms | 
| totalSize     |      31,576kb | 38,826kb | 7,249kb | 22.96 |         WORSE | 34,559kb | 42,692kb | 
| domSize       |         119kb |    119kb |     0kb |  0.04 |         WORSE |    119kb |    119kb | 
| styleSize     |         425kb |    425kb |     0kb | -0.02 | INSIGNIFICANT |    425kb |    437kb | 
| jsObjectsSize |      17,998kb | 20,521kb | 2,523kb | 14.02 |         WORSE | 19,286kb | 21,747kb | 
| jsStringsSize |         618kb |    619kb |     1kb |  0.17 |         WORSE |    615kb |    623kb | 
| jsOtherSize   |      12,061kb | 16,789kb | 4,728kb | 39.20 |         WORSE | 13,625kb | 19,439kb | 
| otherSize     |         355kb |    353kb |    -2kb | -0.67 | INSIGNIFICANT |    351kb |    434kb |

I'm not sure how this branch could have such worse memory usage, but that seems worth investigating

marco-c:
> I'm not sure how this branch could have such worse memory usage, but that seems worth investigating

Could you run the benchmark again? I suspect there's something wrong in the results.
I don't know how many rounds you've executed, but it'd be better to run at least 30.

mykmelez:
> I'm not sure how this branch could have such worse memory usage, but that seems worth investigating

Did you kill the app between the baseline and benchmark runs? I had a weird memory result on #1309, and I think it might have been related to using the *Install and Run* button in WebIDE to reinstall the app without restarting it (since that button appears to merely update the package and then reload the app without actually killing/restarting the app's process).

mykmelez:
I get better results on desktop (Nightly, 30 rounds, 1.5s delay, restart between runs) for a complex midlet:

| startupTime   |       1,201ms |  1,149ms |  -52ms |  -4.33 |        BETTER |  1,112ms |  1,216ms | 
| totalSize     |      26,780kb | 26,127kb | -653kb |  -2.44 |        BETTER | 25,087kb | 26,829kb | 
| domSize       |         152kb |    151kb |    0kb |  -0.29 | INSIGNIFICANT |    151kb |    151kb | 
| styleSize     |         412kb |    412kb |    0kb |  -0.05 |        BETTER |    411kb |    412kb | 
| jsObjectsSize |      16,881kb | 16,788kb |  -93kb |  -0.55 |        BETTER | 16,778kb | 16,795kb | 
| jsStringsSize |         930kb |    624kb | -307kb | -32.97 |        BETTER |    623kb |    624kb | 
| jsOtherSize   |       8,011kb |  7,757kb | -253kb |  -3.16 | INSIGNIFICANT |  6,725kb |  8,452kb | 
| otherSize     |         395kb |    396kb |    1kb |   0.21 |         WORSE |    395kb |    397kb |

mykmelez:
On device, however, my results are inconsistent.

On my first run, I saw the same memory regression you saw:

|      Test     | Baseline Mean |   Mean   |    +/-   |   %   |       P       |    Min   |    Max   | 
|:--------------|--------------:|---------:|---------:|------:|--------------:|---------:|---------:|
| startupTime   |      11,389ms | 12,003ms |    614ms |  5.39 |         WORSE | 11,585ms | 13,190ms | 
| totalSize     |      22,563kb | 37,775kb | 15,212kb | 67.42 |         WORSE | 37,040kb | 38,331kb | 
| domSize       |          71kb |     71kb |      0kb |  0.22 |         WORSE |     70kb |     71kb | 
| styleSize     |         291kb |    294kb |      3kb |  0.91 | INSIGNIFICANT |    285kb |    299kb | 
| jsObjectsSize |      15,453kb | 27,964kb | 12,511kb | 80.96 |         WORSE | 27,941kb | 27,975kb | 
| jsStringsSize |         922kb |  1,528kb |    606kb | 65.79 |         WORSE |  1,528kb |  1,539kb | 
| jsOtherSize   |       5,612kb |  7,688kb |  2,076kb | 36.99 |         WORSE |  7,034kb |  8,210kb | 
| otherSize     |         214kb |    230kb |     16kb |  7.42 | INSIGNIFICANT |    181kb |    255kb | 

On my second run, however, memory usage was slightly better than baseline:

| startupTime   |      11,389ms | 11,449ms |   60ms |   0.53 | INSIGNIFICANT | 10,302ms | 12,695ms | 
| totalSize     |      22,563kb | 22,222kb | -341kb |  -1.51 |        BETTER | 21,216kb | 23,146kb | 
| domSize       |          71kb |     71kb |    0kb |   0.01 | INSIGNIFICANT |     70kb |     71kb | 
| styleSize     |         291kb |    292kb |    0kb |   0.14 | INSIGNIFICANT |    285kb |    300kb | 
| jsObjectsSize |      15,453kb | 15,396kb |  -57kb |  -0.37 |        BETTER | 15,379kb | 15,414kb | 
| jsStringsSize |         922kb |    635kb | -286kb | -31.06 |        BETTER |    635kb |    636kb | 
| jsOtherSize   |       5,612kb |  5,608kb |   -4kb |  -0.06 | INSIGNIFICANT |  4,664kb |  6,441kb | 
| otherSize     |         214kb |    220kb |    6kb |   2.65 | INSIGNIFICANT |    182kb |    287kb |

TimAbraldes:
Updated to a new approach where we don't create a new function (and closure) every time that `DisplayDevice.refresh0` is called. Running startup benchmarks now and will post when I have data

TimAbraldes:
Not nearly as bad as before, but still a regression: (75 rounds, 1.5s)

|      Test     | Baseline Mean |   Mean   |   +/-   |   %   |       P       |    Min   |    Max   | 
|:--------------|--------------:|---------:|--------:|------:|--------------:|---------:|---------:|
| startupTime   |       1,876ms |  2,023ms |   147ms |  7.83 |         WORSE |  1,487ms |  2,371ms | 
| totalSize     |      28,365kb | 31,111kb | 2,745kb |  9.68 |         WORSE | 27,371kb | 34,363kb | 
| domSize       |         123kb |    123kb |     0kb |  0.05 |         WORSE |    122kb |    123kb | 
| styleSize     |         435kb |    435kb |     0kb |  0.08 | INSIGNIFICANT |    435kb |    448kb | 
| jsObjectsSize |      16,910kb | 17,862kb |   952kb |  5.63 |         WORSE | 16,941kb | 18,790kb | 
| jsStringsSize |         699kb |    685kb |   -14kb | -1.97 |        BETTER |    682kb |    689kb | 
| jsOtherSize   |       9,832kb | 11,641kb | 1,809kb | 18.40 |         WORSE |  8,691kb | 13,964kb | 
| otherSize     |         367kb |    365kb |    -2kb | -0.46 | INSIGNIFICANT |    362kb |    447kb |

TimAbraldes:
Just to be sure that I'm not getting bogus results, I ran the tests in opposite order (I made my branch the baseline and master the benchmarked branch): 75 rounds, 1500ms

|      Test     | Baseline Mean |   Mean   |   +/-  |   %   |       P       |    Min   |    Max   | 
|:--------------|--------------:|---------:|-------:|------:|--------------:|---------:|---------:|
| startupTime   |       1,771ms |  2,063ms |  292ms | 16.48 |         WORSE |  1,480ms |  2,343ms | 
| totalSize     |      28,203kb | 27,950kb | -253kb | -0.90 |        BETTER | 26,608kb | 28,550kb | 
| domSize       |         123kb |    123kb |    0kb | -0.02 |        BETTER |    122kb |    123kb | 
| styleSize     |         435kb |    435kb |    0kb |  0.06 | INSIGNIFICANT |    435kb |    446kb | 
| jsObjectsSize |      16,882kb | 16,888kb |    7kb |  0.04 |         WORSE | 16,876kb | 16,897kb | 
| jsStringsSize |         687kb |    682kb |   -4kb | -0.61 |        BETTER |    682kb |    685kb | 
| jsOtherSize   |       9,711kb |  9,457kb | -254kb | -2.62 |        BETTER |  8,079kb | 10,052kb | 
| otherSize     |         366kb |    365kb |   -1kb | -0.41 | INSIGNIFICANT |    362kb |    451kb | 
"

TimAbraldes:
So the startup time hit is clearly bogus, but the memory hit might be a real issue

marco-c:
According to @EricRahm sizeOfTab is not completely reliable on b2g, so 250 kB is probably insignificant.

mykmelez:
What kind of results do you get testing a simple midlet like TestCanvas?

mykmelez:
The other interesting test is to measure memory usage for a complex midlet at various points during runtime via `$B2G_REPO/tools/get_about_memory.py --minimize` (where $B2G_REPO is the path to your local clone of https://github.com/mozilla-b2g/B2G).

midp/gfx.js
```diff
@@ -112,8 +112,34 @@ var currentlyFocusedTextEditor;
...
+        // However, we sometimes "draw" more frequently than the
+        // screen can update. By calling `requestAnimationFrame` here,
+        // we ensure that whatever we just drew will stick around for
+        // one frame
+        if (hasNewFrame) {
+            hasNewFrame = false;
+            window.requestAnimationFrame(gotNewFrame);
+        } else {
+            ctxToExecuteOnNewFrame = $.ctx;
+            $.pause("Async");
```
mykmelez: This string may be allocated anew with each call to *refresh0*, in which case it would be better to store it in a variable outside the function.

midp/gfx.js
```diff
@@ -112,8 +112,34 @@ var currentlyFocusedTextEditor;
         return 1;
     };
 
+    var hasNewFrame = true;
+    var ctxToExecuteOnNewFrame;
+    function gotNewFrame(timestamp) {
+        if (ctxToExecuteOnNewFrame) {
+            ctxToExecuteOnNewFrame.execute();
+            ctxToExecuteOnNewFrame = null;
```
mykmelez: Can *execute* eventually call *refresh0*? If so, *refresh0* will set *ctxToExecuteOnNewFrame*, after which this code will set that to `null`.

TimAbraldes:
Using "about:memory" in a complex app, I compared master to this branch and found a difference of 0.01MB in the "class(Object)" category. The other numbers that looked relevant were all exactly the same.

midp/gfx.js
TimAbraldes: I copied this from our `AsyncImpl` implementation, so it seems like `AsyncImpl` has the same issue? Filed #1371. I'll fix both issues in this PR

midp/gfx.js
TimAbraldes: I'm not sure why but I had imagined that `Context.execute` ran asynchronously (i.e. calling it would return immediately and then on the next step of the event loop the VM would actually start up again). Looking at its definition, that's definitely not the case, so the issue you point out is a real one. I'll fix this and update the PR

TimAbraldes:
Rebased with changes addressing review comments

bench/GraphicsBench.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    public String getDescription() {
+      return description;
+    }
+
+    private int width;
+    private int height;
+    private String description = "Graphics.drawString baseline centered";
+    private String string = "String string = new String(\"string\");";
```
mykmelez: Nit: here and above, *description* and *string* can be *final*.

mykmelez:
I no longer see any significant difference in the startup benchmark (30 rounds, 5s delay) for a complex midlet on device:

|      Test     | Baseline Mean |   Mean   |   +/-  |   %   |       P       |    Min   |    Max   | 
|:--------------|--------------:|---------:|-------:|------:|--------------:|---------:|---------:|
| startupTime   |      11,293ms | 11,187ms | -106ms | -0.94 | INSIGNIFICANT | 10,500ms | 12,255ms | 
| totalSize     |      22,362kb | 22,393kb |   30kb |  0.14 | INSIGNIFICANT | 21,603kb | 23,124kb | 
| domSize       |          73kb |     73kb |    0kb | -0.08 | INSIGNIFICANT |     72kb |     73kb | 
| styleSize     |         301kb |    299kb |   -2kb | -0.52 | INSIGNIFICANT |    290kb |    307kb | 
| jsObjectsSize |      15,409kb | 15,402kb |   -8kb | -0.05 | INSIGNIFICANT | 15,382kb | 15,435kb | 
| jsStringsSize |         635kb |    636kb |    1kb |  0.09 | INSIGNIFICANT |    635kb |    647kb | 
| jsOtherSize   |       5,710kb |  5,761kb |   52kb |  0.90 | INSIGNIFICANT |  5,042kb |  6,446kb | 
| otherSize     |         234kb |    222kb |  -12kb | -5.28 | INSIGNIFICANT |    165kb |    282kb |

midp/gfx.js
```diff
@@ -112,8 +112,35 @@ var currentlyFocusedTextEditor;
...
+        // However, we sometimes "draw" more frequently than the
+        // screen can update. By calling `requestAnimationFrame` here,
+        // we ensure that whatever we just drew will stick around for
+        // one frame
+        if (hasNewFrame) {
+            hasNewFrame = false;
+            window.requestAnimationFrame(gotNewFrame);
+        } else {
+            ctxToExecuteOnNewFrame = $.ctx;
+            $.pause(asyncImplStringAsync);
```
brendandahl: I'm still trying to figure out some of the event queue/repaint stuff, but I'm starting to think synchronization with request animation frame should happen further up repaint call. Refresh0 is called at the very end of repaint whereas you'd typically want to start your drawing on the beginning of the request animation frame. I think this should happen in RepaintEventProducer in either process() or scheduleRepaint(). As far as I can tell on desktop we repaint too often(we can repaint under 3ms) whereas on a flame repaint takes 80ms-300ms.

brendandahl: I should also note, that it'd be preferable to use asyncImpl. Using ctx.execute() is very tricky and I'd like to only reserve it for a few special places and eventually make it private.

bench/GraphicsBench.java
```diff
@@ -0,0 +1,240 @@
+package bench;
```
mykmelez: Nit: all the other benchmark classes in this directory are in the *benchmark* package (unless they're in a package from some part of the platform they need to access), so this one should be too.

TimAbraldes:
> I'm still trying to figure out some of the event queue/repaint stuff, but I'm starting to think synchronization with request animation frame should happen further up repaint call. Refresh0 is called at the very end of repaint whereas you'd typically want to start your drawing on the beginning of the request animation frame. I think this should happen in RepaintEventProducer in either process() or scheduleRepaint(). As far as I can tell on desktop we repaint too often(we can repaint under 3ms) whereas on a flame repaint takes 80ms-300ms.

That makes sense. Do you have a patch/branch doing that? Otherwise I'm happy to investigate on this branch

brendandahl:
My naive approach:

```diff
diff --git a/java/custom/org/mozilla/internal/Sys.java b/java/custom/org/mozilla/internal/Sys.java
index c110f41..eeb9d09 100644
--- a/java/custom/org/mozilla/internal/Sys.java
+++ b/java/custom/org/mozilla/internal/Sys.java
@@ -77,6 +77,8 @@ public final class Sys {
    */
   public native static void eval(String src);
 
+  public native static void requestAnimationFrame();
+
   public static void throwException(Exception e) throws Exception {
     throw e;
   }
diff --git a/java/midp/com/sun/midp/lcdui/RepaintEventProducer.java b/java/midp/com/sun/midp/lcdui/RepaintEventProducer.java
index 06b0982..e978a7a 100644
--- a/java/midp/com/sun/midp/lcdui/RepaintEventProducer.java
+++ b/java/midp/com/sun/midp/lcdui/RepaintEventProducer.java
@@ -94,6 +94,7 @@ public class RepaintEventProducer implements EventListener {
      */
     public void scheduleRepaint(DisplayEventConsumer d,
                                 int x, int y, int w, int h, Object target) {
+        org.mozilla.internal.Sys.requestAnimationFrame();
         RepaintEvent freeEvent;
 
         synchronized (this) {
diff --git a/jit/analyze.ts b/jit/analyze.ts
index ee149ea..3b91b22 100644
--- a/jit/analyze.ts
+++ b/jit/analyze.ts
@@ -70,6 +70,7 @@ module J2ME {
     "java/lang/Thread.start0.()V": YieldReason.Root,
     "java/lang/Class.forName0.(Ljava/lang/String;)V": YieldReason.Root,
     "java/lang/Class.newInstance1.(Ljava/lang/Object;)V": YieldReason.Root,
+    "org/mozilla/internal/Sys.requestAnimationFrame.()V": YieldReason.Root,
     // Test Files:
     "gnu/testlet/vm/NativeTest.throwExceptionAfterPause.()V": YieldReason.Root,
     "gnu/testlet/vm/NativeTest.returnAfterPause.()I": YieldReason.Root,
diff --git a/native.js b/native.js
index bdd7d73..4435673 100644
--- a/native.js
+++ b/native.js
@@ -425,6 +425,12 @@ Native["java/lang/Double.longBitsToDouble.(J)D"] = (function() {
     }
 })();
 
+Native["org/mozilla/internal/Sys.requestAnimationFrame.()V"] = function() {
+  asyncImpl("V", new Promise(function(resolve) {
+    window.requestAnimationFrame(resolve);
+  }));
+};
+
 Native["java/lang/Throwable.fillInStackTrace.()V"] = function() {
     this.stackTrace = [];
     $.ctx.frames.forEach(function(frame) {
diff --git a/java/midp/com/sun/midp/lcdui/RepaintEventProducer.java b/java/midp/com/sun/midp/lcdui/RepaintEventProducer.java
index e978a7a..11dedf2 100644
--- a/java/midp/com/sun/midp/lcdui/RepaintEventProducer.java
+++ b/java/midp/com/sun/midp/lcdui/RepaintEventProducer.java
@@ -181,6 +181,7 @@ public class RepaintEventProducer implements EventListener {
      * @param genericEvent event to process
      */
     public void process(Event genericEvent) {
+        // try { Thread.sleep(1); } catch (Exception e) {}
         RepaintEvent event = (RepaintEvent)genericEvent;
 
         synchronized (this) {
@@ -192,10 +193,12 @@ public class RepaintEventProducer implements EventListener {
          * Target DisplayEventConsumer is obtained directly from event field.
          * Assumed that target consumer is not null.
          */
+        long start = com.sun.cldchi.jvm.JVM.monotonicTimeMillis();
         event.display.handleRepaintEvent(
                 event.paintX1, event.paintY1, 
                 event.paintX2, event.paintY2, 
                 event.paintTarget);
+        System.out.println("____ paint took " + (com.sun.cldchi.jvm.JVM.monotonicTimeMillis() - start));
 
         synchronized (this) {
             /* Change the ID here to signal waitForRepaint. */
diff --git a/native.js b/native.js
index 4435673..4ed5b7a 100644
--- a/native.js
+++ b/native.js
@@ -541,7 +541,7 @@ Native["java/lang/Thread.start0.()V"] = function() {
     // Create a context for the thread and start it.
     var newCtx = new Context($.ctx.runtime);
     newCtx.thread = this;
-
+console.log(newCtx.id);
     var classInfo = CLASSES.getClass("org/mozilla/internal/Sys");
     var run = classInfo.getMethodByNameString("runThread", "(Ljava/lang/Thread;)V", true);
     newCtx.start([new Frame(run, [ this ], 0)]);
```

brendandahl:
I still seem to have better results just doing a sleep(1) in the process() function though.

TimAbraldes:
Rebased with review comments addressed. This also make us wait in `Display.callPaint` instead of in `DisplayDevice.refresh0`. `Display.callPaint` is the function that actually processes paint events by calling into application logic so it seems like the most logical place to sync up with `requestAnimationFrame`

brendandahl:
As I mentioned in IRC my best results so far are putting the sync into RepaintEventProducer.process() at the top since this causes multiple repaint events to be merged into one if we're still painting. Two commonly tested scrollable windows in a midlet work pretty well, still not buttery smooth but way better.  I also have an idea why we're losing some touch events, but I'll look into that more tomorrow.

TimAbraldes:
> As I mentioned in IRC my best results so far are putting the sync into RepaintEventProducer.process() at the top since this causes multiple repaint events to be merged into one if we're still painting. Two commonly tested scrollable windows in a midlet work pretty well, still not buttery smooth but way better.

From looking at how events get merged, that makes sense. I'll update this PR to put the sync in `RepaintEventProducer.process`

TimAbraldes:
Rebased; this is what the change will look like. I haven't tested yet.

marco-c:
Won't blocking in *process* block the EventQueue for other events as well?

TimAbraldes:
> Won't blocking in process block the EventQueue for other events as well?

No lock is held when we block in `RepaintEventProducer.process`. I believe that other threads are free to do whatever they need to do with the event queue. Maybe I'm missing the issue you're pointing out?

marco-c:
No, you're right, the eventQueueLock is released before calling *process*.

TimAbraldes:
Rebased onto current master. Updated asteroids to use `JVM.monotonicTimeMillis`. Updated graphics bench reported values to be in units with a different order of magnitude

TimAbraldes:
Startup bench with a complex MIDlet, 100 rounds, 10s round delay:

|      Test     | Baseline Mean |   Mean   |  +/-  |   %   |       P       |    Min   |    Max   | 
|:--------------|--------------:|---------:|------:|------:|--------------:|---------:|---------:|
| startupTime   |       1,641ms |  1,633ms |  -7ms | -0.45 | INSIGNIFICANT |  1,539ms |  1,844ms | 
| totalSize     |      28,894kb | 28,874kb | -19kb | -0.07 | INSIGNIFICANT | 27,070kb | 29,475kb | 
| domSize       |         122kb |    122kb |   0kb | -0.00 | INSIGNIFICANT |    122kb |    123kb | 
| styleSize     |         435kb |    435kb |   0kb |  0.01 | INSIGNIFICANT |    435kb |    439kb | 
| jsObjectsSize |      17,371kb | 17,379kb |   8kb |  0.05 |         WORSE | 17,363kb | 17,389kb | 
| jsStringsSize |         666kb |    659kb |  -6kb | -0.96 |        BETTER |    658kb |    662kb | 
| jsOtherSize   |       9,936kb |  9,915kb | -21kb | -0.21 | INSIGNIFICANT |  8,126kb | 10,508kb | 
| otherSize     |         363kb |    364kb |   0kb |  0.10 | INSIGNIFICANT |    362kb |    371kb |

midp/gfx.js
brendandahl: If hasNewFrame is false, and this were to be executed twice on different contexts the first context to call it would never be resumed.

TimAbraldes: I believe that `RepaintEventProducer.process` should only ever be called from `RepaintEventProdcuer.serviceRepaints` after this check:

    if (EventQueue.isDispatchThread()) {
    ....

which means that it should be impossible for `waitForAnimationFrame` to be called on different contexts

brendandahl: Sorry for the chain of comments, but I think using execute() here is actually okay because we really do want to begin execution right away on the callback for the request animation frame and not wait until the next tick that asyncImpl would cause: Perhaps:
```javascript
var ctx = $.ctx;
$.pause(asyncImplStringAsync);
window.requestAnimationFrame(function() {
  ctx.execute();
});
```

TimAbraldes: My understanding is that this could potentially cause us to wait when we don't need to.

Let's say that our screen can refresh once every 20ms, and that all of our application logic means that it takes 21ms from one call of `waitForAnimationFrame` to the next. In this scenario, by the time `waitForAnimationFrame` is called, we can already immediately draw to the screen but we will instead call `window.requestAnimationFrame` and wait for a future opportunity to draw.

This is what `hasNewFrame` and `gotNewFrame` are intended to avoid. The idea is that, if we call `waitForAnimationFrame` but we know that we don't need to wait for a new frame, we should just go ahead and draw without waiting.

Let me know if my analysis is flawed.

brendandahl: > I believe that RepaintEventProducer.process should only ever be called from RepaintEventProducer.serviceRepaints after this check:

It's unlikely, but there could be two isolates each of which would have there own event queues.

As for the other thing about waiting. This seems to be a wash. Seth, one of the guys familiar with graphics here said to try both, but we'll never seen anything updated out of band with a requestAnimationFrame. As for the code in this patch...you'll always be waiting for the requestAnimationFrame cb so I don't see how it is any different than my suggestion above. e.g. if hasNewFrame=true you still do window.requestAnimationFrame.

brendandahl: Arg, I see the different(lack of pause).

TimAbraldes: > It's unlikely, but there could be two isolates each of which would have there own event queues.

I hadn't considered that possibility. Hopefully only the foreground isolate can ever be processing repaint events? Hmm, need to think on this one.

brendandahl: We probably don't need to support that, but an assert would be good.

TimAbraldes:
@brendandahl I've rebased with a potential solution to the "multiple isolates might call `waitForAnimationFrame` before a frame actually becomes available" issue. I'll post back with updated startup benchmark results when I have them. Let me know what you think!

TimAbraldes:
100 rounds, 10s round delay with a complex MIDlet

| startupTime   |       1,519ms |  1,499ms |  -20ms |  -1.32 | INSIGNIFICANT |  1,408ms |  1,845ms | 
| totalSize     |      28,167kb | 27,650kb | -517kb |  -1.84 |        BETTER | 26,266kb | 28,195kb | 
| domSize       |         122kb |    122kb |    0kb |   0.00 | INSIGNIFICANT |    122kb |    123kb | 
| styleSize     |         435kb |    435kb |    0kb |  -0.06 | INSIGNIFICANT |    435kb |    448kb | 
| jsObjectsSize |      16,812kb | 16,747kb |  -65kb |  -0.39 |        BETTER | 16,731kb | 16,758kb | 
| jsStringsSize |         904kb |    613kb | -291kb | -32.24 |        BETTER |    611kb |    617kb | 
| jsOtherSize   |       9,528kb |  9,369kb | -159kb |  -1.67 | INSIGNIFICANT |  7,998kb |  9,908kb | 
| otherSize     |         366kb |    365kb |   -2kb |  -0.41 | INSIGNIFICANT |    362kb |    447kb |

TimAbraldes:
100 rounds, 10s round delay. This time opening the MIDlet to a different state (one that has an animation) and actually remembering to use "logLevel=log":

|      Test     | Baseline Mean |   Mean   |  +/-  |   %   |       P       |    Min   |    Max   | 
|:--------------|--------------:|---------:|------:|------:|--------------:|---------:|---------:|
| startupTime   |       1,696ms |  1,693ms |  -2ms | -0.13 | INSIGNIFICANT |  1,598ms |  1,875ms | 
| totalSize     |      34,813kb | 34,714kb | -99kb | -0.28 | INSIGNIFICANT | 32,533kb | 36,082kb | 
| domSize       |         122kb |    122kb |   0kb |  0.01 |         WORSE |    122kb |    123kb | 
| styleSize     |         435kb |    435kb |   0kb |  0.11 |         WORSE |    435kb |    448kb | 
| jsObjectsSize |      19,884kb | 19,862kb | -22kb | -0.11 |        BETTER | 19,809kb | 20,103kb | 
| jsStringsSize |         895kb |    891kb |  -4kb | -0.45 |        BETTER |    890kb |    895kb | 
| jsOtherSize   |      13,114kb | 13,038kb | -76kb | -0.58 | INSIGNIFICANT | 10,878kb | 14,363kb | 
| otherSize     |         362kb |    365kb |   3kb |  0.78 | INSIGNIFICANT |    362kb |    446kb |

marco-c:
It'd be more interesting to test runtime performance with a gfx benchmark instead of the startup time (since startup time is up to the first call to "gainedForeground0", which probably happens before MIDlets do any drawing).

TimAbraldes:
> It'd be more interesting to test runtime performance with a gfx benchmark instead of the startup time (since startup time is up to the first call to "gainedForeground0", which probably happens before MIDlets do any drawing).

True. Is this information we want before merging? All empirical testing seems to show that this PR improves performance of complex MIDlets on device.

marco-c:
> True. Is this information we want before merging? All empirical testing seems to show that this PR improves performance of complex MIDlets on device.

No, I just wanted to suggest that in general we shouldn't rely on the startup benchmark to evaluate performance benefits of patches to gfx code.

TimAbraldes:
> No, I just wanted to suggest that in general we shouldn't rely on the startup benchmark to evaluate performance benefits of patches to gfx code.

Ah OK. In the future we'll be able to use the graphics benchmarks that this PR adds :)

marco-c:
@brendandahl can you review this, since you've been looking into this with @TimAbraldes?

PR #1318, opened 2015-03-26
Update SpiderMonkey shell version

marco-c:


marco-c:
I'm going to land this because the old SpiderMonkey URL is no longer there, so we can't run Travis build.

PR #1333, opened 2015-03-27
make some file writes asynchronous, then optimize them

mykmelez:
https://github.com/mozilla/j2me.js/commit/61f41ee74dac99042aabae9ab74b9c563e6b20bf improves the responsiveness of a particular interaction in a complex midlet by making some file writes asynchronous, so repeated writes on a background thread don't starve the UI thread.

Making the writes asynchronous makes the write operations themselves many times slower (even if the app can do other work in the meantime, so overall app perf doesn't suffer). https://github.com/mozilla/j2me.js/commit/e222f5e2565556941abfb5156b0878811b70cae8 claws back much of that performance by exiting *file/Protocol.ensureConnect* early if *fileHandler* is defined instead of relying on *connect* to do so, which avoids the expensive `filePath + fileName` string concatenation.

(@brendandahl's self-hosting strings branch might also help here, but even if concatenation becomes faster, there's still no need to do it here at all.)

On master, FileConnectionBench averages ~50ms to do these writes. On this branch, it averages ~150ms, so it's still three times slower. But that interaction is much more responsive (although it's still somewhat janky, and this isn't a complete fix).

I expect general runtime speed and responsiveness to improve as well with this fix, although I don't have a good metric for it. Unsure about memory usage, which depends on the cost of StringBuffer.

I didn't expect startup time/memory usage to improve, since I wouldn't expect too much file I/O to happen in the startup path; but I do see some improvement in one test on desktop:

| startupTime   |       1,164ms |  1,115ms |  -48ms |  -4.17 |        BETTER |  1,092ms |  1,131ms | 
| totalSize     |      26,681kb | 26,199kb | -482kb |  -1.81 | INSIGNIFICANT | 25,132kb | 26,941kb | 
| domSize       |         152kb |    152kb |    0kb |   0.00 | INSIGNIFICANT |    152kb |    152kb | 
| styleSize     |         412kb |    412kb |    0kb |   0.03 | INSIGNIFICANT |    412kb |    412kb | 
| jsObjectsSize |      16,872kb | 16,799kb |  -72kb |  -0.43 |        BETTER | 16,791kb | 16,806kb | 
| jsStringsSize |         922kb |    629kb | -293kb | -31.80 |        BETTER |    629kb |    629kb | 
| jsOtherSize   |       7,928kb |  7,812kb | -116kb |  -1.47 | INSIGNIFICANT |  6,754kb |  8,548kb | 
| otherSize     |         395kb |    394kb |    0kb |  -0.10 | INSIGNIFICANT |    394kb |    395kb |

### DIFF
--- a/java/custom/com/sun/cdc/io/j2me/file/Protocol.java
+++ b/java/custom/com/sun/cdc/io/j2me/file/Protocol.java
@@ -876,7 +876,9 @@ public class Protocol extends ConnectionBaseAdapter implements FileConnection {
      * @throws IOException if the connection is closed
      */
     protected void ensureConnected() throws IOException {
-        connect(fileRoot, filePath + fileName);
+        if (fileHandler == null) {
+            connect(fileRoot, filePath + fileName);
+        }
     }
 
     /**
--- a/jit/analyze.ts
+++ b/jit/analyze.ts
@@ -29,6 +29,7 @@ module J2ME {
     "com/sun/midp/lcdui/DisplayDevice.gainedForeground0.(II)V": YieldReason.Root,
     "com/sun/cdc/io/j2me/file/DefaultFileHandler.openForRead.()V": YieldReason.Root,
     "com/sun/cdc/io/j2me/file/DefaultFileHandler.openForWrite.()V": YieldReason.Root,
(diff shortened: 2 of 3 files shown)


marco-c:
ensureConnected was used a lot in a MIDlet I'm testing, it made one particular interaction way faster.

PR #1337, opened 2015-03-28
benchmark calling sync vs. async natives

mykmelez:
I tried a few things to speed up the asynchronous writes in #1333 further, but I didn't find anything that had much of an impact. So I rolled this benchmark of sync vs. async natives, which finds that async natives are about 2-5 times slower than synchronous ones.

The newly-async native in #1333 is about 3 times slower than the old, synchronous one on FileConnectionBench, so I think it's pretty close to optimal, even though its *OutputStream.write* calls indirect through several classes and perform some checks that are probably unnecessary (or could be made a bit faster).


java/custom/org/mozilla/Test.java
```diff
@@ -0,0 +1,6 @@
+package org.mozilla;
```
marco-c: Can you move this class in the bench/ directory? This way we don't need to package it in classes.jar for release.

mykmelez: > Can you move this class in the bench/ directory? This way we don't need to package it in classes.jar for release.

How about I put it into the tests/ directory? Then it'll go into tests.jar, which gets added to the *config.jars* list when you run the benchmark (although the mechanism is indirect; I think it happens because the default *midletClassName* is `RunTests`, and tests.jar gets added to *config.jars* in that case).

This benchmark already relies on tests/native.js (which gets loaded if tests.jar is in *config.jars*), so there's precedence for a dependency on tests/. Maybe we should even move the whole bench/ directory into tests/ at some point.

In any case, I've made that change in e21ac5a15e515a43063139cd1e4574e1e0ac8edf. What do you think?

mykmelez: (FWIW, in the long run, I think we should make this more explicit, removing the test-centric defaults from main.js.)

marco-c: Works for me! I've done the same in the past with the Sha256 class (which I added to tests/ and not bench/).

PR #1341, opened 2015-03-28
Fix alert buttons style

marco-c:


tests/automation.js
```diff
@@ -31,6 +31,7 @@ casper.options.onWaitTimeout = function() {
 
 var gfxTests = [
   { name: "gfx/AlertTest", maxDifferent: 1942 },
+  { name: "gfx/AlertTwoCommandsTest", maxDifferent: 2186 },
```
mykmelez: This is exactly the number of differing pixels I see on my Mac. Does this mean your Mac works now? :-)

Issue #1287, opened 2015-03-20
Camera image position is wrong


marco-c:
The video element that holds the camera stream isn't in the correct position (it's in a completely different spot of the screen, not on top of the canvas).

Maybe a regression from https://github.com/mozilla/j2me.js/pull/1224 ?

marco-c:
To reproduce, open http://localhost:8000/index.html?main=com/sun/midp/main/MIDletSuiteLoader&midletClassName=gfx/CameraTest&jars=tests/tests.jar

marco-c:
Fixed by https://github.com/mozilla/j2me.js/pull/1327.

PR #1327, opened 2015-03-27
Revert "Merge pull request #1224 from PinZhang/pzhang/resize-canvas"

marco-c:
This PR reverts https://github.com/mozilla/j2me.js/pull/1224, that caused a regression with the "Downloading MIDlet" screen when autosize=1

PR #1346, opened 2015-03-30
Fix regressions with camera image position and downloading MIDlet screen.

PinZhang:


PinZhang:
Fixed https://github.com/mozilla/j2me.js/issues/1287 and https://github.com/mozilla/j2me.js/pull/1327.

mykmelez:
Thanks, @PinZhang, this fixes all the regressions we noticed!

PR #1347, opened 2015-03-30
[WIP] Share stack array with all frames.

yxl:
We allocate two arrays for each frame, one storing local variables and one as operand stack. I try to reduce the number of arrays by merging the two array into one and making all frames share the same stack.

Also, for  the stack containing different types of value, it is little faster to implement with object properties than an array:
http://jsperf.com/stack-implementation-with-array-object

The new implementation is not equivalent to the original one and I'm trying to figure out why...

### DIFF
--- a/interpreter.ts
+++ b/interpreter.ts
@@ -165,6 +165,7 @@ module J2ME {
 
         return;
       }
+      frame.free();
       ctx.popFrame();
       frame = ctx.current();
       if (Frame.isMarker(frame)) {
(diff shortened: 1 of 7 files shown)


marco-c:
Have you tried disabling the JIT? It'd allow you to understand whether the problem is in the changes to the JIT or the interpreter.

yxl:
> Have you tried disabling the JIT? It'd allow you to understand whether the problem is in the changes to the JIT or the interpreter.
Not yet. I should have try it:)

yxl:
I made the test passed, but the performance is a bit poorer than the original one. I tried to make `stackPush` method inline, but the result is worse.

Since it causes performance regression, I have to leave this open :-(

mykmelez:

PR #1350, opened 2015-03-30
WIP: move constructors/methods from global to Methods object

mykmelez:
While investigating another issue a few weeks ago, I noticed that setting/getting properties on the global object is slower than doing so on another object. So I've been investigating moving all the constructor and method functions into a *Methods* object. This is a work-in-progress that doesn't quite work yet. It does run some simple midlets (like InitMidlet), but it fails with more complex ones (including TestCanvas, which reports completion but never draws to screen).


mykmelez:

PR #1353, opened 2015-03-30
Don't read FS blob into an ArrayBuffer when not necessary

marco-c:
This should improve memory usage when we show media files and when MIDlets use the Nokia image processing API (I think it's used pretty often).


mykmelez:
The changes make sense. Any chance you can put together a micro-benchmark for this?

marco-c:
5 runs without the patch I get results between 3500 ms and 4200 ms, with the patch between 3000 ms and 3600 ms.
I suppose the win is going to be higher for bigger images and drastic with videos (where Firefox doesn't need to read the entire video in memory in order to play it).

mykmelez:
Hmm, I was hoping for a memory benchmark, since this is suppose to be a memory win. Adding *MemorySampler.sampleMemory* calls around the *bench.runBenchmark* call suggests that the changes improve memory usage:

```
      MemorySampler.sampleMemory("Memory before");
      bench.runBenchmark();
      MemorySampler.sampleMemory("Memory after");
```

Can you add that to the benchmark? It isn't great—we should format it better, and perhaps run it multiple times automatically—but it's still useful information.

marco-c:
I added it initially, but there wasn't any difference before/after the patch, so I assumed that all the memory was being released when we perform the memory minimization step.

marco-c:
I've added them anyway in https://github.com/mozilla/j2me.js/commit/ccf4c06fa19e70d2fbe86e0c99d349308f7a1d9a

PR #1361, opened 2015-03-31
Optimize skip method for filesystem input streams

marco-c:


marco-c:
This is basically ready, but I can't test it with the MIDlet I wanted to test it with at the moment because of a regression.

mykmelez:
Nice.

PR #1363, opened 2015-03-31
More gfx cleanup

TimAbraldes:
Remove more `Graphics` methods that can't be called.

Make the native implementations of `Graphics.getMaxWidth` and
`Graphics.getMaxHeight` actually match the signature of those methods.

### DIFF
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -570,10 +570,6 @@ var currentlyFocusedTextEditor;
         reset(this, 0, 0, this.maxWidth, this.maxHeight);
     };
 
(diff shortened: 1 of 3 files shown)


midp/gfx.js
```diff
@@ -597,11 +593,11 @@ var currentlyFocusedTextEditor;
         return this.transY;
     };
 
-    Native["javax/microedition/lcdui/Graphics.getMaxWidth.()I"] = function() {
+    Native["javax/microedition/lcdui/Graphics.getMaxWidth.()S"] = function() {
```
marco-c: Can you add a simple test to make sure we don't regress those? (not a gfx test, a testlet)

TimAbraldes: Added

PR #1366, opened 2015-03-31
revert Closure compilation of j2me.js to SIMPLE optimizations to fix PROFILE

mykmelez:
J2ME_OPTIMIZATIONS breaks PROFILE=1|2 for some reason, unsure why. This branch works around the problem by reverting Closure compilation of j2me.js to SIMPLE optimizations if PROFILE is set to a value that enables the profiler.


marco-c:
Can you file an issue to remove the workaround?

mykmelez:
> Can you file an issue to remove the workaround?

Yes, filed as https://github.com/mozilla/j2me.js/issues/1367!

PR #1368, opened 2015-04-01
Fix rotation calculations. Add tests.

TimAbraldes:
The j2me API has far fewer footguns than the HTML Canvas API.
Implementing it using the HTML Canvas API involves a lot of severed
feet.

WIP because tests are failing

PR #1375, opened 2015-04-01
shush more stuff by default

mykmelez:
This should help reduce the number of lines we log to Travis. We may still need reduce it more to get under 10K lines once @TimAbraldes's #1368 lands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ RELEASE ?= 0
 VERSION ?=$(shell date +%s)
 PROFILE ?= 0
 BENCHMARK ?= 0
+VERBOSE ?= 0
 
 # Sensor support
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -27,6 +28,14 @@ else
   J2ME_JS_OPTIMIZATION_LEVEL = SIMPLE
 endif
 
+# Closure is really chatty, so we shush it by default to reduce log lines
+# for Travis.
+ifeq ($(VERBOSE),1)
+  CLOSURE_WARNING_LEVEL = VERBOSE
+ else
```
marco-c: I think you need to remove the space here (before the else)

marco-c: No, it works anyway.

Issue #1000, opened 2015-01-31
Handle Media download with JS also for audio files


marco-c:
We're already doing this for images, we should do the same for audio files.

Right now, for audio files we're letting Java handle the download of the file (either from http:// or file://) with the MediaDownload class. We should handle the download ourselves (at least for file://, that is what most MIDlets use).

marco-c:
This is now harder because the Audio element is in the privileged context.

PR #1379, opened 2015-04-02
Don't create unneeded HTMLAudioElement

marco-c:
Noticed this is unused while looking into https://github.com/mozilla/j2me.js/issues/1000


TimAbraldes:
Verified with `grep -r "\.audio"` that this isn't used anywhere

PR #1382, opened 2015-04-02
Show profiles per thread and don't leave() on unwinds.

brendandahl:
Makes it easier to get the full picture of thread execution.  Some notes:

- The profiler intercepts scroll events so to hack this so we can scroll down I added a "bumper" bar on the side so you can put your mouse there and scroll to see more of the threads.
- We don't currently have a good way to track async natives. I currently just always leave() the native even during an unwind.  This will cause whatever called an async native to appear as if it has been executing for a long time. To make this better would require some further hacking on the timeline and our code.


main.js
```diff
@@ -3,8 +3,9 @@
 
 'use strict';
 
-// The real profile variable declaration in config.ts is folded away by closure. Until we
+// The real profile and release variable declaration in config.ts is folded away by closure. Until we
 // make closure process this file also, make sure that |profile| is defined in this file.
```
mykmelez: Nit: `|profile| is` -> `|profile| and |release| are`

main.js
```diff
@@ -3,8 +3,9 @@
...
+var release;
 var profile;
 var jvm = new JVM();
```
mykmelez: Nit: I would also put a space after the dummy declarations to make it clear that the *jvm* declaration is unrelated to the comment.

vm/context.ts
```diff
@@ -399,6 +407,21 @@ module J2ME {
...
+      }
+      return frame;
+    }
+
+    pushFrame(frame: Frame) {
+      if (profile) {
+        this.enterMethodTimeline(frame.methodInfo.implKey, MethodType.Interpreted);
+      }
+      this.frames.push(frame);
+    }
```
mykmelez: I like the refactoring, but is there any potential for a perf regression here? I imagine *pushFrame* and *popFrame* could be pretty hot.

vm/runtime.ts
```diff
@@ -1467,12 +1482,8 @@ module J2ME {
...
     }
 
-    if (false && methodTimeline) {
-      fn = profilingWrapper(fn, methodInfo, methodType);
-    }
-
-    if (traceWriter) {
-      fn = tracingWrapper(fn, methodInfo, methodType);
+    if (profile) {
+      fn = wrapMethod(fn, methodInfo, methodType);
```
mykmelez: It looks like this disables the trace writer unless profiling is enabled, but I find the trace writer to be handy even without the profiler, so I'd rather remain able to enable it discretely.

vm/context.ts
```diff
@@ -466,8 +490,10 @@ module J2ME {
     }
 
     start(frames: Frame[]) {
-      frames.unshift(Frame.Start);
-      this.frames = frames;
+      this.frames = [Frame.Start];
+      for (var i = 0; i < frames.length; i++) {
+        this.pushFrame(frames[i]);
+      }
```
mykmelez: I like this refactoring too, since it avoids side-effects in the caller. It does result in the allocation of an extra array, but this method seems unlikely to be hot, so I don't think it matters.

Nevertheless, note that *this.frames* is already initialized to a new empty array in the *Context* constructor, so you could initialize it to `[Frame.Start]` there and then merely push the *frames* onto it here.

vm/runtime.ts
brendandahl: I was contemplating ripping that out all together because the same thing can be achieved with PROFILE=2 and it is more accurate now since it shows all method enters/exits. I'll restore for now, but something to think about.

vm/context.ts
brendandahl: I didn't notice anything on a quick desktop tests. I'll apply your comments and run on device in a bit.

mykmelez: (Presumably the *profile* conditional blocks will be removed by Closure. I'm only wondering about the extra function call for each push/pop.)

vm/runtime.ts
mykmelez: I'm ok with that. The trace writer currently outputs more readable graphs, but we can fix that in the profiler.

I've also been wanting an easy way to enable/disable the trace writer at runtime, so I can trace only particular interactions. That should be doable either way, though.

brendandahl:
I inlined all the frame pops and pushes so it was back to the old way and there was no performance improvement.  I then started looking at the other changes and noticed I changed how the clinit function was executed. This seems to be the source of the regression.

On flame 30rounds 5s delay:
| startupTime   |      12,016ms | 12,008ms |   -7ms |  -0.06 | INSIGNIFICANT | 11,237ms | 12,476ms | 
| totalSize     |      21,043kb | 20,673kb | -370kb |  -1.76 |        BETTER | 20,551kb | 20,808kb | 
| domSize       |          72kb |     72kb |    0kb |   0.49 |         WORSE |     72kb |     72kb | 
| styleSize     |         286kb |    287kb |    2kb |   0.55 | INSIGNIFICANT |    281kb |    296kb | 
| jsObjectsSize |      15,444kb | 15,395kb |  -49kb |  -0.32 |        BETTER | 15,373kb | 15,417kb | 
| jsStringsSize |         875kb |    577kb | -297kb | -33.98 |        BETTER |    577kb |    578kb | 
| jsOtherSize   |       4,148kb |  4,118kb |  -30kb |  -0.72 |        BETTER |  4,061kb |  4,161kb | 
| otherSize     |         219kb |    223kb |    3kb |   1.59 | INSIGNIFICANT |    177kb |    296kb | 

Desktop:

|      Test     | Baseline Mean |    Mean   |  +/-  |   %   |       P       |    Min    |    Max    | 
|:--------------|--------------:|----------:|------:|------:|--------------:|----------:|----------:|
| startupTime   |         858ms |     866ms |   8ms |  0.94 | INSIGNIFICANT |     832ms |   1,007ms | 
| totalSize     |     108,413kb | 108,788kb | 375kb |  0.35 |         WORSE | 107,629kb | 110,809kb | 
| domSize       |         178kb |     178kb |   1kb |  0.34 |         WORSE |     178kb |     179kb | 
| styleSize     |         438kb |     439kb |   1kb |  0.30 |         WORSE |     438kb |     440kb | 
| jsObjectsSize |      76,516kb |  76,532kb |  16kb |  0.02 |         WORSE |  76,523kb |  76,739kb | 
| jsStringsSize |       2,894kb |   2,900kb |   6kb |  0.19 |         WORSE |   2,898kb |   2,938kb | 
| jsOtherSize   |      27,946kb |  28,301kb | 355kb |  1.27 |         WORSE |  27,159kb |  30,101kb | 
| otherSize     |         442kb |     438kb |  -3kb | -0.78 | INSIGNIFICANT |     414kb |     445kb |

mykmelez:
Nice catch. Makes me wonder where else we could avoid calling *getLinkedMethod*, or if there's a way to make that function faster.

PR #1381, opened 2015-04-02
Run basic unit tests with JIT compilation forcibly enabled for all methods

marco-c:
Let's see what the failures are.

PR #1387, opened 2015-04-03
WIP: Revert "Merge pull request #1301 from marco-c/dont_override_Permissions"

mykmelez:
This reverts commit 92d4e822c66352dbc694a5236f9a58b03bd7be80, i.e. #1301, reversing
changes made to bb451ba472a5f97f1ead82918cf404d89cf18869. It's a work-in-progress because I still see the regression that I thought was caused by those changes. So those changes may not be the culprit.


mykmelez:
Testing with #1381 suggests that Soot optimization may be to blame, especially since it causes *InvalidRecordIDException*, which also happens when I experience the regression. And Soot optimization landed in #1312 around the same time as #1301. So I've merged #1381 and will close this PR unless new evidence arises.

PR #1394, opened 2015-04-07
add make target to package app

mykmelez:
This adds a Make target, *package*, that creates a ZIP package for distribution as a packaged app. Once #1393 lands, we might consider naming the package after the name and version of the app being packaged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 *.jar
 *.jad
 *.class
-*.zip
 .DS_Store
 certs/j2se_main.ks
 certs/j2se_test.ks
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.jar
 *.jad
 *.class
+*.zip
 .DS_Store
 certs/j2se_main.ks
 certs/j2se_test.ks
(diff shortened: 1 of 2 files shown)


marco-c:
Can you add "package.zip" to .gitignore?

marco-c:
Also, maybe we should move it to the output/ directory instead of the top-level directory?

mykmelez:
> Can you add "package.zip" to .gitignore?

I did so originally, but then I reverted that change, since the best practice is to add a *.zip rule to your global .gitignore, per [this gist](https://gist.github.com/octocat/9257657).

mykmelez:
> Also, maybe we should move it to the output/ directory instead of the top-level directory?

I've been wondering that as well, but I can think of a good reason to do it (although I also don't have a good reason to prefer the top-level directory). Do you have any thoughts on why this might be preferable?

mykmelez:
> I've been wondering that as well, but I can think of a good reason to do it

Erm, I mean that I *can't* think of a good reason to do it.

marco-c:
> Do you have any thoughts on why this might be preferable?

Not really, for some reason I expected it to be in output/ when I built, but I can't tell why :smile:

mykmelez:
> Not really, for some reason I expected it to be in output/ when I built, but I can't tell why :smile:

I suppose it might make it easier to clean up packages if we start naming them after the app name and version. Perhaps we can switch at that point.

PR #1396, opened 2015-04-07
increase default CasperJS log level to reduce logging for Travis

mykmelez:
We've exceeded the 10K log line display limit on Travis again. This should cut out ~4K lines of CasperJS logging.

### DIFF
--- a/tests/automation.js
+++ b/tests/automation.js
@@ -18,7 +18,6 @@ casper.on('remote.message', function(message) {
 
 casper.options.waitTimeout = 80000;
 casper.options.verbose = true;
-casper.options.logLevel = "debug";
 casper.options.viewportSize = { width: 240, height: 320 };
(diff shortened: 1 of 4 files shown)


mykmelez:
With this we're back down to about 7K lines. You can re-enable the effluent via `make test VERBOSE=1`.

tests/automation.js
```diff
@@ -18,7 +18,6 @@ casper.on('remote.message', function(message) {
 
 casper.options.waitTimeout = 80000;
 casper.options.verbose = true;
-casper.options.logLevel = "debug";
```
marco-c: Can we do the same with tests/fs/automation.js?

mykmelez: > Can we do the same with tests/fs/automation.js?

Yes, b40fbe5 does that.

PR #1399, opened 2015-04-08
Fix RecordStore race conditions

marco-c:


libs/fs.js
```diff
@@ -566,6 +566,17 @@ var fs = (function() {
...
     openedFile.dirty = false;
+
+    // Update in-memory copies of the same file, only if they haven't been
+    // modified.
+    // If they've been modified, the behavior is undefined.
+    for (var entry of openedFiles) {
+      if (!entry[1].dirty && entry[1].path === openedFile.path) {
+        entry[1].mtime = openedFile.mtime;
+        entry[1].size = openedFile.size;
+        entry[1].buffer = new FileBuffer(new Int8Array(openedFile.buffer.getContent()));
```
mykmelez: I wonder if this could allocate a lot of memory, in the case that a midlet opens the same file twice and flushes it frequently. I guess that doesn't happen very often, though.

mykmelez:
@marco-c and I chatted about this on IRC. I think we need to take this, even if it causes a perf hit, because it fixes a significant correctness issue that can get triggered even by unrelated changes (f.e. to thread scheduling). Nevertheless, it'd be useful to know what we're sacrificing, and @marco-c will do some measurements.

marco-c:
I've added some logging to understand how many times RecordStore::commitWrite was called.
I've run a complex MIDlet for a while, and it was called only two times.

I suspect MIDlets in general use the RecordStore to store some configuration data, so it's mostly static data that isn't going to change often.

There are more flushes on first run, when the store is used more (and where the race condition was most likely to occur).

PR #1410, opened 2015-04-10
Allow MIDlets to customize the SMS dialog verification input type and max length

marco-c:


marco-c:
This has been reviewed by @TimAbraldes in #1411.

PR #1411, opened 2015-04-10
Honor max length property also when the input type is 'number'

marco-c:
Depends on https://github.com/mozilla/j2me.js/pull/1410.


TimAbraldes:
This seems fine but could you rebase this so it isn't 10 commits?

PR #1414, opened 2015-04-10
Run main class static initializer before calling main method.

brendandahl:


TimAbraldes:
Is this a perf improvement? Do we have numbers showing its effect?

brendandahl:
This is purely a correctness issue.

brendandahl:
Let's hold off for a sec, I think I can write a test for this.

brendandahl:
@TimAbraldes added the test

TimAbraldes:
Great! Tests pass on Travis. Also verified that I could run my favorite MIDlet with this patch applied

PR #1420, opened 2015-04-16
Long Optimizations

mbebenita:


marco-c:
I assume these long values were the most used ones in a complex MIDlet?

If you remove the first unrelated commit, I can merge this PR right away.

mbebenita:
All these PRs build on each other. Any reason we don't merge them in order?

marco-c:
@brendandahl PR is being reviewed by @TimAbraldes, I asked you to decouple this PR from the other so that I can merge this before the other is merged (I thought they weren't related).

marco-c:
The other PR has been merged, so this should be ready.

Have you found these long values (0, 1, 2, 3, 4) to be the most used ones in a complex MIDlet?

mbebenita:
The 0,1,2,3,4 were the most common Longs in the SHA256 benchmark from my favorite midlet. It's unclear whether the benchmark was operating over a representative dataset.

marco-c:
These *fromBits* stats are from a run of a complex MIDlet: https://pastebin.mozilla.org/8830461
The format is ["LOWBITS.HIGHBITS", FREQUENCY].

Seems like NEG_ONE is used a lot, so we should optimize it as well.

libs/long.js
```diff
@@ -96,7 +96,17 @@
     * @param {number} highBits The high 32-bits.
     * @return {!Long} The corresponding Long value.
     */
+    var count = 0;
```
marco-c: Nit: remove unused variable

libs/long.js
```diff
@@ -96,7 +96,17 @@
...
     function fromBits(lowBits, highBits) {
+        if (highBits === 0) {
```
marco-c: Can you add a comment here explaining why we're doing this?

libs/long.js
```diff
@@ -345,6 +355,14 @@
     * @return {!Long} The sum of this and the given Long.
     */
     Long.prototype.add = function (other) {
+        if (other === ONE && this.high_ === 0) {
```
marco-c: Can you add a comment here explaining why we're doing this?

marco-c:
These *add* stats (with other === ONE) are from a run of a complex MIDlet: https://pastebin.mozilla.org/8830461
The format is ["LOWBITS", FREQUENCY].

Seems like NEG_ONE is used a lot, so we should optimize it as well.

libs/long.js
mbebenita: Looks like this was already addressed in another PR.

mbebenita:
> Anyway, if the result of add is something already cached by fromBits, I think we can avoid to cache it in add as well.

There is some extra computation in `add` that we could avoid, however it's straight line code and probably really fast, while the fast path code is branchy.

Can you rerun your analysis on a midlet that spends a lot of time computing SHA256? My profile, looked different which is why I'm wondering. I was profiling a synthetic benchmark, so I may have been wrong.

mbebenita:
I will close this PR, and continue the work on #1422

PR #1421, opened 2015-04-16
Faster static calls.

mbebenita:
This inlines the static method lookup fast path, improves perf on some benchmarks by 30%.


marco-c:
Can you remove the first two unrelated commits?

PR #1422, opened 2015-04-16
Long Optimizations

mbebenita:
Most of the optimizations in this PR deal with reducing the number of Long box allocations through caching. Caching takes more space, but reduces drastically the number of allocations. In some midlets, scrolling through UI elements can allocate upwards of 10K temporary Long objects. This PR reduces that to a few hundred.

### DIFF
--- a/jit/baseline.ts
+++ b/jit/baseline.ts
@@ -900,11 +900,13 @@ module J2ME {
         args.unshift(this.pop(signatureKinds[i]));
       }
       var object = null, call;
+      var classConstant = this.localClassConstant(methodInfo.classInfo);
+      var methodConstant = "(" + classConstant + ".M[" + methodInfo.index + "]||" + classConstant + ".m(" + methodInfo.index + "))";
       if (opcode !== Bytecodes.INVOKESTATIC) {
         object = this.pop(Kind.Reference);
         if (opcode === Bytecodes.INVOKESPECIAL) {
           args.unshift(object);
-          call = this.localClassConstant(methodInfo.classInfo) + ".m(" + methodInfo.index + ").call(" + args.join(",") + ")";
+          call = methodConstant + ".call(" + args.join(",") + ")";
         } else if (opcode === Bytecodes.INVOKEVIRTUAL) {
           call = object + "." + methodInfo.virtualName + "(" + args.join(",") + ")";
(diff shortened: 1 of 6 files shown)


jit/baseline.ts
```diff
@@ -71,7 +71,7 @@ module J2ME {
    * Emits array bounds checks. Although this is necessary for correctness, most
    * applications work without them.
    */
-  export var emitCheckArrayBounds = true;
+  export var emitCheckArrayBounds = false;
```
marco-c: I think we shouldn't do this unless we prove MIDlets work flawlessly without array bound checks.

libs/long.js
```diff
@@ -309,6 +342,15 @@
     *     if the given one is greater.
     */
     Long.prototype.compare = function (other) {
```
marco-c: Can you add some simple tests for the fast paths in *compare*, *add* and *shiftRightUnsigned* to gnu.testlet.vm.LongTest?

marco-c:
Can you write a simple benchmark for Long to see how much this improves perf in the fast path (and what changes in the slow path)?

jit/baseline.ts
mbebenita: Type-o on my part.

mbebenita:
This is most likely reduces peek perf but reduces GC pressure.

marco-c:
> This is most likely reduces peek perf but reduces GC pressure.

I'd expect this to be faster for example when comparing with 0.
Anyway, a benchmark would be nice to know how much we're losing/gaining.

mbebenita:
Doesn't really depend on #1424 but should land after, since it includes those commits.

marco-c:
Most used ones in fromBits:
https://pastebin.mozilla.org/8830461 (normal run of a MIDlet)
https://pastebin.mozilla.org/8830852 (first run of a MIDlet that is using sha256)

Most used ones in add (with OTHER === ONE):
https://pastebin.mozilla.org/8830464 (normal run of a MIDlet)
https://pastebin.mozilla.org/8830853 (first run of a MIDlet that is using sha256)

mbebenita:
Thanks. Why is the number of longs in runs with sha256 less than in a
normal run?

marco-c:
The "normal" run was a longer session (a few minutes).

mykmelez:

PR #1428, opened 2015-04-18
Fix MozActivity usage

marco-c:
It needs to happen in the privileged context.


main.js
```diff
@@ -149,7 +149,7 @@ if (config.downloadJAD) {
 
 if (jars.indexOf("tests/tests.jar") !== -1) {
   loadingPromises.push(loadScript("tests/native.js"),
-                       loadScript("tests/mozactivitymock.js"),
+                       loadScript("tests/mozactivitymock.unprivileged.js"),
```
mykmelez: Nit: I would have probably called these *testactivitymock-(index|main).js*.

PR #1430, opened 2015-04-20
WIP: Support MIDlet pausing/restarting

marco-c:
Some MIDlets perform initialization steps in *startApp* that are supposed to happen on startup or when they return to the foreground.


tests/automation.js
```diff
@@ -546,6 +546,31 @@ casper.test.begin("unit tests", 25 + gfxTests.length, function(test) {
...
 
+    // The following test tests that a MIDlet can be started, paused and then started again.
+    casper
+    .thenOpen("http://localhost:8000/index.html?midletClassName=tests.background.BackgroundMIDlet1&jad=tests/midlets/background/pause.jad&jars=tests/tests.jar&logConsole=web,page&logLevel=log")
+    .withFrame(0, function() {
+        casper.waitForText("startApp", function() {
+            casper.waitForText("pauseApp", function() {
+                casper.waitForText("startApp", function() {
+                    var content = this.getPageContent();
+                    test.assertEquals(content.match(/startApp/g).length, 2, "MIDlet started twice");
```
mykmelez: This happens to work, but I think it's only because the second *startApp* races the *waitForText* timeout. According to [the waitForText docs](http://casperjs.readthedocs.org/en/latest/modules/casper.html#waitfortext), it simply "Waits until the passed text is present in the page contents," which means the second `casper.waitForText("startApp", …)` call will find the first instance of *startApp* if the second hasn't appeared yet.

Fortunately, the test will still fail if the app doesn't get started twice; but it could also fail when it does, leading to a false negative. To ensure that doesn't happen, remove the *startApp* content before restarting the app or make the app write a different message to the content node the second time around.

tests/midlets/background/PauseMIDlet.java
```diff
@@ -0,0 +1,32 @@
+package tests.background;
```
mykmelez: (NB: I wonder if we should distinguish between midlets that are intended to be run manually and those that support automated tests. Right now we mix both in tests/midlets/.)

tests/automation.js
marco-c: IIRC I expressly tried with a single *startApp* and the test didn't pass, but I probably remember wrong.

tests/midlets/background/PauseMIDlet.java
marco-c: I'm trying to make all the tests I write under tests/midlets/ to be usable both manually and in automated tests, but there are probably some that only work in automated tests.

mykmelez:
I restarted the Travis build, and it failed the second time too, although it succeeds on my Mac.

marco-c:
It succeeds locally as well.

mykmelez:
Hmm, I saw the failure on the second run on my Mac, so the problem exists there too intermittently.

marco-c:
This will be obsoleted by https://github.com/mozilla/j2me.js/pull/1556 and is not actually needed for what I wanted to support (because many MIDlets don't actually implement pauseApp).

PR #1431, opened 2015-04-20
Add URL parameter to defer VM startup

mbebenita:
If you now specify `deferStartup=1000` as an URL parameter, VM startup will be delayed by `1000` ms.


marco-c:
Instead of doing this, we could break up the startup time reported by the startup benchmark in several parts.
For example:
 - Time to start the VM
 - Time from the start of the VM to the start of the background MIDlet
 - Time from the start of the background MIDlet to the start of the foreground MIDlet

mbebenita:
We should do both. Having a pause helps with reducing noise.

PR #1425, opened 2015-04-17
Use short class names whenever possible.

mbebenita:
Short class names, like `Ab` are now used as is and not expanded to larger names as they used to because of our hashing / name mangling scheme.

PR #1434, opened 2015-04-21
In the startup benchmark, break up the startup time in several values

marco-c:
- vmStartupTime: time to load scripts and MIDlet files (right before starting to load the BG MIDlet)
- bgStartupTime: time from the start of the BG MIDlet to the start of the FG MIDlet
- fgStartupTime: time from the start of the FG MIDlet to the point where the FG MIDlet goes in the foreground

We could break up these in even smaller values in followups.

### DIFF
--- a/benchmark.js
+++ b/benchmark.js
@@ -131,7 +131,10 @@ var Benchmark = (function() {
   }
 
   var valueFormatters = {
-    startupTime: msFormatter,
+    totalStartupTime: msFormatter,
+    vmStartupTime: msFormatter,
+    backgroundStartupTime: msFormatter,
+    foregroundStartupTime: msFormatter,
     totalSize: byteFormatter,
     domSize: byteFormatter,
     styleSize: byteFormatter,
@@ -181,7 +184,10 @@ var Benchmark = (function() {
     run: function(settings) {
       storage.round = 0;
       var current = storage.current = {};
-      current.startupTime = [];
+      current.totalStartupTime = [];
+      current.vmStartupTime = [];
+      current.backgroundStartupTime = [];
+      current.foregroundStartupTime = [];
       if (settings.recordMemory) {
         storage.recordMemory = true;
         current.totalSize     = [];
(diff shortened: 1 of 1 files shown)


marco-c:
Example of a benchmark run from #1425:

|          Test         | Baseline Mean |   Mean   |   +/-   |   %   |       P       |    Min   |    Max   | 
|:----------------------|--------------:|---------:|--------:|------:|--------------:|---------:|---------:|
| totalStartupTime      |      19,804ms | 20,221ms |   417ms |  2.11 | INSIGNIFICANT | 17,414ms | 23,477ms | 
| vmStartupTime         |       3,182ms |  3,311ms |   129ms |  4.06 | INSIGNIFICANT |  2,629ms |  4,363ms | 
| backgroundStartupTime |       3,817ms |  3,753ms |   -64ms | -1.68 | INSIGNIFICANT |  1,748ms |  4,978ms | 
| foregroundStartupTime |      12,804ms | 13,157ms |   352ms |  2.75 | INSIGNIFICANT | 10,650ms | 16,927ms | 
| totalSize             |      38,267kb | 43,284kb | 5,016kb | 13.11 | INSIGNIFICANT | 29,586kb | 77,294kb | 
| domSize               |         104kb |    105kb |     1kb |  0.86 | INSIGNIFICANT |    103kb |    109kb | 
| styleSize             |         322kb |    322kb |     0kb |  0.10 | INSIGNIFICANT |    319kb |    325kb | 
| jsObjectsSize         |      23,739kb | 27,257kb | 3,518kb | 14.82 | INSIGNIFICANT | 19,709kb | 58,486kb | 
| jsStringsSize         |       2,043kb |  2,057kb |    14kb |  0.68 | INSIGNIFICANT |  1,470kb |  4,600kb | 
| jsOtherSize           |      11,578kb | 13,061kb | 1,483kb | 12.81 |         WORSE |  6,118kb | 20,240kb | 
| otherSize             |         481kb |    481kb |     0kb |  0.03 | INSIGNIFICANT |    450kb |    572kb |

benchmark.js
```diff
@@ -390,11 +401,33 @@ var Benchmark = (function() {
...
-          originalFn.apply(null, arguments);
+
+        var vmImplKey = "com/sun/midp/main/MIDletSuiteUtils.vmEndStartUp.(I)V";
+        var vmOriginalFn = Native[vmImplKey];
+        var vmCalled = false;
+        Native[vmImplKey] = function() {
+          if (!vmCalled) {
+            vmCalled = true;
+            startup.stopTimer("vmStartupTime");
+            startup.startTimer("backgroundStartupTime");
```
mykmelez: Nit: since *startTimer* and *stopTimer* both call *performance.now*, a tiny bit of *totalStartupTime* will be unaccounted for in its constituent parts. It's probably insignificant, but I would still endeavour to use the same value when stopping the previous partial timer as I do when starting the next one, f.e. by calling *performance.now* here and passing its value to both functions.

benchmark.js
```diff
@@ -131,7 +131,10 @@ var Benchmark = (function() {
   }
 
   var valueFormatters = {
-    startupTime: msFormatter,
+    totalStartupTime: msFormatter,
+    vmStartupTime: msFormatter,
+    backgroundStartupTime: msFormatter,
+    foregroundStartupTime: msFormatter,
```
mykmelez: Nit: I would actually call these *bgStartupTime* and *fgStartupTime*, as in your example, to conserve the width of the table.

PR #1436, opened 2015-04-21
don't minify scripts unless RELEASE

mykmelez:
@brendandahl suggested this, and it seems useful when debugging. We should do it for main-all.js too, but that's trickier, since a simple *cat* of all the scripts doesn't strip the *use strict* declarations, and some of the code is non-strict.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,8 +236,10 @@ relooper:
 bld/j2me.js: $(BASIC_SRCS) $(JIT_SRCS) build_tools/closure.jar .checksum
 	@echo "Building J2ME"
 	tsc --sourcemap --target ES5 references.ts -d --out bld/j2me.js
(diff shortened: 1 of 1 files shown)


mbebenita:
Is the concern here that minified scripts are hard to debug? If so, there
is a closure flag that minifies yet emits pretty formatting.

mykmelez:
> Is the concern here that minified scripts are hard to debug? If so, there is a closure flag that minifies yet emits pretty formatting.

Yes, that's the concern. Are you referring to `--formatting PRETTY_PRINT`? I noticed that yesterday but didn't follow up with it. I'll give it a try today and request another pull if it seems to address the concern.

mbebenita:
Yes, that's what I've been using to make it easier to debug things locally.

PR #1451, opened 2015-04-24
Add regular expression benchmark.

brendandahl:


bench/Regex.java
```diff
@@ -0,0 +1,22 @@
...
+      RE r2 = new RE("(\\d{3})(\\d{4})");
+      String number = "5555231119";
+      boolean m1 = false;
+      boolean m2 = false;
+      for (int i = 0; i < 1000; i++) {
+        m1 = r1.match(number);
+        m2 = r2.match(number);
+      }
+      System.out.println("Matches: " + m1 + " " + m2);
+      System.out.println("REGEX: " + (JVM.monotonicTimeMillis() - start));
```
marco-c: Can you split the total time in the time taken to compile and the time taken to match?

Nit: move this line before the other System.out.println so we don't consider the time spent in System.out.println

mbebenita:
If the bytecode compilation threshold is increased to 3000 this beachmark becomes a lot faster.

mykmelez:
@marco-c Is this ready to go now?

marco-c:
Yes!

PR #1455, opened 2015-04-24
Optimize linking of classBinding fields

marco-c:


vm/runtime.ts
```diff
@@ -1321,47 +1321,47 @@ module J2ME {
...
-          delete symbols[key];
+      release || assert(!classBindings.fields.staticSymbols, "Static fields are not supported yet");
+
+      var instanceSymbols = classBindings.fields.instanceSymbols;
+
+      var fieldNames = Object.keys(instanceSymbols);
+
+      for (var i = 0; i < fieldNames.length; i++) {
+        var fieldName = fieldNames[i];
+        var fieldSignature = instanceSymbols[fieldName];
```
mykmelez: Nit: I'm unsure if it'd make a difference for perf, but you should be able to iterate *instanceSymbols* property names directly:

```JS
for (var fieldName in instanceSymbols) {
  var fieldSignature = instanceSymbols[fieldName];
  …
}
```

marco-c: I did it with Object.keys because I recall something about for ... in being slow, although a simple test (http://jsperf.com/forinobjectkeys) shows that for .. in is faster.

mykmelez:
[This jsperf test](http://jsperf.com/object-keys-vs-for-in-with-closure/3) is also relevant (although it doesn't measure exactly the same thing).

I've seen admonitions against using for…in, for example in [this StackOverflow thread](http://stackoverflow.com/questions/13645890/javascript-for-in-vs-for-loop-performance) and [this Jon Raasch blog post](http://jonraasch.com/blog/10-javascript-performance-boosting-tips-from-nicholas-zakas). But they typically address the different use case of iterating arrays, which is not only less performant but also runs the risk of iterating non-indexed properties.

The blog post also suggests that the JS engine "has to set up a function for each" iteration of a for…in loop, although I don't think that's correct.

mbebenita:
The problem with `for each` is memory allocation. You need to snapshot all the key values before iterating over them.

mykmelez:
> The problem with for each is memory allocation. You need to snapshot all the key values before iterating over them.

We aren't using *for each…in*, we're using *for…in*. So iterating over keys, not values. Does that have the same problem?

mbebenita:
```
var y = {a: 0, b: 0};
for (var x in y) {
  print (x);
  y.c = 2;
}
```
prints:
```
a
b
```

The JS engine needs to snapshot the list of keys before iterating. In theory, this could be optimized, but @shu says that it's probably not.

mykmelez:
The original implementation called *Object.keys()* to get the keys to iterate, effectively taking a snapshot of the list of keys itself instead of relying on the JS engine to do it. Is that any better?

(Either way, I think this is academic, given that the objects we're iterating are small. But I'm still curious.)

mbebenita:
I was just commenting on it in general, not comparing the two approaches. It seems like they would behave similarly. These days I just try to stick to C like loops.

PR #1457, opened 2015-04-24
File system write improvements.

brendandahl:
Mbx found that in an install run we were spending a huge amount of time in asyncimpl, subarray, and the interpreter.

The large amount of time in asyncimpl was because of nearly 60000 unwinds while writing one byte at a time. The large amount of interpreter code was because of the unwinds and because some very hot large methods were not being compiled. All the subarray calls were because of creating a 1 byte subarray.

In #1333 asyncImpl was added to not lock the UI during writes, but the huge amount of unwinds I think is the source of a regression that was noticed on install run.

Fixes:
 - Don't unwind during write
 - Compile larger methods to support some fs operations
 - Avoid allocations, by copying buffers instead of using subarray.

Results:
With this patch on the install run I am able to get to a second screen of a midlet in ~49s vs ~68s and my first tap was noticed. Also, for the file connection benchmark it goes from:
1:1  | OutputStream.write time: 105 
1:1  | OutputStream.write single byte time: 578 
to
1:1  | OutputStream.write time: 22 
1:1  | OutputStream.write single byte time: 6 

Device Testing for normal startup:

| User Agent       | Mozilla/5.0 (Mobile; rv:38.0) Gecko/38.0 Firefox/38.0 | 

|      Test     | Baseline Mean |   Mean   |   +/-  |   %   |   P   |    Min   |    Max   | 
|:--------------|--------------:|---------:|-------:|------:|------:|---------:|---------:|
| startupTime   |      12,201ms | 11,896ms | -305ms | -2.50 |  SAME | 10,429ms | 13,008ms | 
| vmStartupTime |       2,284ms |  2,316ms |   32ms |  1.40 |  SAME |  2,153ms |  2,460ms | 
| bgStartupTime |       1,662ms |  1,642ms |  -19ms | -1.15 |  SAME |  1,554ms |  1,821ms | 
| fgStartupTime |       8,255ms |  7,937ms | -318ms | -3.85 |  SAME |  6,556ms |  9,047ms | 
| totalSize     |      21,481kb | 21,570kb |   89kb |  0.42 | WORSE | 21,448kb | 21,707kb | 
| domSize       |          71kb |     71kb |    0kb |  0.09 |  SAME |     71kb |     71kb | 
| styleSize     |         285kb |    287kb |    2kb |  0.76 |  SAME |    280kb |    293kb | 
| jsObjectsSize |      15,823kb | 15,820kb |   -3kb | -0.02 |  SAME | 15,794kb | 15,850kb | 
| jsStringsSize |         870kb |    954kb |   84kb |  9.71 | WORSE |    953kb |    955kb | 
| jsOtherSize   |       4,216kb |  4,207kb |   -8kb | -0.20 |  SAME |  4,151kb |  4,267kb | 
| otherSize     |         217kb |    231kb |   14kb |  6.47 |  SAME |    178kb |    271kb |


libs/fs.js
```diff
@@ -521,13 +521,15 @@ var fs = (function() {
       from = buffer.contentSize;
     }
 
-    var newLength = (from + data.byteLength > buffer.contentSize) ? (from + data.byteLength) : (buffer.contentSize);
+    var newLength = (from + length > buffer.contentSize) ? (from + length) : (buffer.contentSize);
 
     buffer.setSize(newLength);
 
-    buffer.array.set(data, from);
+    for (var i = 0; i < length; i++) {
```
marco-c: We should use a loop when the amount of data to write is small, *set* otherwise.

vm/runtime.ts
```diff
@@ -1700,7 +1700,7 @@ module J2ME {
     }
 
     // Don't compile methods that are too large.
-    if (methodInfo.codeAttribute.code.length > 2000 && !config.forceRuntimeCompilation) {
+    if (methodInfo.codeAttribute.code.length > 4000 && !config.forceRuntimeCompilation) {
```
marco-c: We should investigate how increasing this affects memory usage (with https://github.com/mozilla/j2me.js/wiki/Tracking-Memory-Usage).

mykmelez:
> The large amount of time in asyncimpl was because of nearly 60000 unwinds while writing one byte at a time. The large amount of interpreter code was because of the unwinds and because some very hot large methods were not being compiled.

Can unwinds be made more efficient?

> With this patch on the install run I am able to get to a second screen of a midlet in ~49s vs ~68s and my first tap was noticed.

My test of the same midlet with this branch show that it takes 20-30 seconds to get from the first to the second screen when tapping a button as soon as the first screen appears. The button changes to a "depressed" state after 10-15 seconds, then the second screen appears after another 10-15 seconds. That's a long time to wait for the UI to update.

> 1:1 | OutputStream.write single byte time: 6

On my Flame this takes ~200ms, and the benchmark does 10k writes, so presumably 60k writes takes over a second.

libs/fs.js
brendandahl: Do we have any numbers on this? Also, what is considered small? The time spent in actual copying seems to be very small compared to the other things going during install run. I'd prefer simpler code that doesn't allocate and performs generally well.

brendandahl:
> My test of the same midlet with this branch show that it takes 20-30 seconds to get from the first to the second screen when tapping a button as soon as the first screen appears.

What version of fxos is this? I haven't seen it take that long on master or this branch?

mykmelez:
> What version of fxos is this? I haven't seen it take that long on master or this branch?

It's FxOS 2.1 on a Flame.

native.js
```diff
@@ -25,6 +25,13 @@ function asyncImpl(returnKind, promise) {
   $.pause(asyncImplStringAsync);
 }
 
+function preemptingImpl(returnKind, returnValue) {
```
brendandahl: Note: I'm not currently using a counter as a cheaper preemption check because we don't have enough calls into the method to always preempt in time for some writes. In a future version we should investigate using a counter and not do the time check on every call (similar to how the disabled preemption code works in baseline.ts). Whether we inline the counter in each preemptingImpl call or put it in here is also debatable.

brendandahl:
New run:

| User Agent       | Mozilla/5.0 (Mobile; rv:38.0) Gecko/38.0 Firefox/38.0 | 

|      Test     | Baseline Mean |   Mean   |   +/-   |    %   |    P   |    Min   |    Max   | 
|:--------------|--------------:|---------:|--------:|-------:|-------:|---------:|---------:|
| startupTime   |      12,178ms | 12,262ms |    84ms |   0.69 |   SAME | 10,904ms | 12,971ms | 
| vmStartupTime |       2,308ms |  2,323ms |    15ms |   0.66 |   SAME |  2,234ms |  2,478ms | 
| bgStartupTime |       1,649ms |  1,671ms |    22ms |   1.31 |   SAME |  1,561ms |  1,767ms | 
| fgStartupTime |       8,221ms |  8,268ms |    47ms |   0.58 |   SAME |  6,929ms |  9,001ms | 
| totalSize     |      20,447kb | 21,559kb | 1,112kb |   5.44 |  WORSE | 21,477kb | 21,687kb | 
| domSize       |          71kb |     71kb |     0kb |  -0.14 | BETTER |     71kb |     71kb | 
| styleSize     |         288kb |    284kb |    -4kb |  -1.26 | BETTER |    280kb |    293kb | 
| jsObjectsSize |      15,460kb | 15,820kb |   360kb |   2.33 |  WORSE | 15,803kb | 15,838kb | 
| jsStringsSize |         847kb |    954kb |   107kb |  12.59 |  WORSE |    953kb |    955kb | 
| jsOtherSize   |       3,546kb |  4,221kb |   674kb |  19.01 |  WORSE |  4,180kb |  4,262kb | 
| otherSize     |         235kb |    209kb |   -25kb | -10.73 | BETTER |    178kb |    271kb |

libs/fs.js
marco-c: IIRC @mbebenita run some measurements when self-hosting some Java class. I don't recall what was the best threshold, it should be easy to build a simple JS benchmark to find it.
On install run we're always writing a single byte, so the for loop is definitely faster, but in the general case we don't know how much data we're writing.

mykmelez: I'm not sure it's quite right, but here's a first cut at a benchmark:

http://jsfiddle.net/d2e3ga5d/

mykmelez:
I've done some stopwatch timings of the time it takes for a button on the initial screen of a midlet to respond to my tap. There are two numbers: the time it takes for the button to appear "depressed" after I tap it, and the time it takes for the next screen of the midlet to appear (counting from the time I initially tapped the button, so the second time is cumulative), in seconds.

On master, I got 4-10, 2-8, 3-10, while on this branch the numbers are 4-11, 4-13, 5-25. (By contrast, after reverting #1333, I get 14-69, 12-66, 16-25.) So this branch seems to regress responsiveness a bit, although the numbers are noisy, and the measurements are obviously quite imprecise. A benchmark would be handy here.

mykmelez:
> Mbx found that in an install run we were spending a huge amount of time in asyncimpl, subarray, and the interpreter.

@mbebenita Any tips on reproducing these results? I.e. which profiler you're using, etc.

libs/fs.js
marco-c: These are the results of the measurements I was talking about: https://github.com/mozilla/j2me.js/pull/996#issuecomment-72332180

marco-c: https://github.com/mozilla/j2me.js/issues/1356 is an issue I filed a while ago to implement this with a threshold.

mykmelez:
These changes were landed (with some updates) in #1519.

PR #1467, opened 2015-04-27
Don't create temporary strings in bin2String

marco-c:
This makes the MD5 benchmark ~2x faster (~280 ms -> ~120 ms).


mykmelez:
@mbebenita If this looks good to you now, please merge it!

PR #1470, opened 2015-04-27
WIP: Prevent ionmonkey inlining by adding try-catch in every compiled function

marco-c:


mykmelez:
@EricRahm Does the try/catch block actually need to surround the function body, or is it enough for there to be a try/catch in the body?

mbebenita:
I think the `try/catch` can be anywhere in the method.

mykmelez:
Ok, then it should be possible to do simply: `this.bodyEmitter.writeLn("try{}catch{}");`.

mykmelez:

PR #1452, opened 2015-04-24
Compiler Optimizations for Regex

mbebenita:
1. This PR increases the compilation threshold from 2000 to 3000 bytecodes (bytes, not instructions). This is now large enough to compile hot methods in the regexp library.

2. This PR also optimizes interface calls, which were slo, allocated memory and prevented inlining.

`Original: 82660ms`
`After 1: 6365ms, 12X faster.`
`After 2: 200ms, 440X faster.`

PR #1478, opened 2015-04-28
Interface Optimization

mbebenita:
This PR cherry picks the interface dispatch changes from #1452


mbebenita:
These two patches were reviewed in #1452.

PR #1482, opened 2015-04-28
simplify startup screens

mykmelez:
This branch makes several changes to simplify the sequence of startup screens that the user sees while starting up the app:

* Display the app splash screen as soon as possible, to minimize the transition between the OS and app splash screens.
* Hide the background screen if the app is visible (f.e. because it was started by the user rather than the background alarm), so it doesn't show up between transitions (from the OS splash screen to the app splash screen, and from the app splash screen to the initial midlet screen).
* Add the midlet name to the splash screen at build time, so it's always visible on the splash screen (instead of sometimes appearing after the splash screen is already displayed).

Along the way, it makes a few other changes:

* Make the style of the download screen consistent with the style of the splash screen, so the transition between them is less significant.
* Add an explanatory message and call to action to the download failure dialog.
* Refactor the code that shows/hides the various screens.
* Remove the download screen from the DOM when hiding it, to avoid the frequent style recalculations that it triggers.


Makefile
```diff
@@ -33,6 +33,7 @@ ifeq ($(PACKAGE_TESTS),1)
 endif
 
 NAME ?= j2me.js
+MIDLET_NAME ?= midlet
```
marco-c: Nit: We could get this from the JAD manifest (the MIDlet-Name property).

main.js
```diff
@@ -3,6 +3,54 @@
...
+  document.getElementById("background-screen").style.display = "none";
+}
+
+// The exit screen is hidden by default, and we only ever show it,
+// so we don't need a hideExitScreen function.
+function showExitScreen() {
+  document.getElementById("exit-screen").style.display = "block";
+}
+
+// If the document is hidden, then we've been started by an alarm and are in
```
marco-c: Nit: Can you move this to background.js?

main.js
```diff
@@ -124,20 +159,21 @@ if (config.downloadJAD) {
   loadingMIDletPromises.push(new Promise(function(resolve, reject) {
     JARStore.loadJAR("midlet.jar").then(function(loaded) {
       if (loaded) {
+        if (!document.hidden) {
+          // Show the splash screen as soon as possible.
+          showSplashScreen();
```
marco-c: I guess there's no way to show this earlier, because we don't know if we need to download the MIDlet or not.

midp/background.js
```diff
@@ -66,7 +66,7 @@ Native["com/nokia/mid/s40/bg/BGUtils.waitUserInteraction.()V"] = function() {
       localmsgServerWait = resolve;
     });
   }).then(function() {
-    document.getElementById("splash-screen").style.display = "block";
-    document.getElementById("background-screen").style.display = "none";
+    showSplashScreen();
```
marco-c: I think we can move this in the previous then function (before returning the localmsg promise).

marco-c: Also, if *!document.hidden*, aren't we calling showSplashScreen twice? Once here and once in main.js?

main.js
mykmelez: I can, but it calls *showBackgroundScreen*, which isn't defined until main.js. Shall I move the whole block of show/hide functions as well?

main.js
mykmelez: > I guess there's no way to show this earlier, because we don't know if we need to download the MIDlet or not.

Right. Initially, I showed it as soon as possible, before knowing about the download. But that created a strange effect on install run, as the app first showed "Starting…", then "Downloading…", then "Starting…" again.

midp/background.js
mykmelez: > I think we can move these lines in the previous then function (before returning the localmsg promise).

Yup, good point, 40d5e6a.

mykmelez: > Also, if !document.hidden, aren't we calling showSplashScreen twice? Once here and once in main.js?

Yes. And here we're also hiding the background screen, even though it may not ever have been shown. But it's ok, because those should be effectively noops.

main.js
marco-c: As you prefer, I don't actually have a preference

midp/background.js
marco-c: *appendChild* should remove the child from the parent and then append it again, but I guess that's not going to be noticeable.

main.js
mykmelez: One improvement we might be able to make is to front-load all the code that blocks displaying the Downloading…/Starting… screens. Currently, the code to display one or the other lives in main.js, which is the last script we load. And we load a lot of scripts before it. If we front-loaded it, we might shorten the time period between the OS splash screen (black screen with app icon) and the first app splash screen (Downloading…/Starting…).

But that would require more significant changes, like moving up the JAR store library (and its dependencies) in the load order. And this branch still makes that time period better, since it replaces the background screen (white screen with app icon), which currently appears there, with a solid black screen that is a less jarring transition.

Another option would be to identify the install run via some other mechanism than the absence of the midlet JAR from the JAR store. For example, we could use the *lastVersion* property that the compiled method cache sets in local storage. Then we could pick between Downloading…/Starting… earlier on that basis. This would work most of the time, except when users fail to download the midlet and then kill/restart the app. And we could make it work all the time by setting another property in local storage that definitely identified the presence of the midlet JAR.

Anyway, these are both worth considering, but I'd do any additional work here in a followup.

main.js
mykmelez: I moved it all to background.js in bb2b29c.

mykmelez:
(The last change is only semi-related, since it's about shutdown rather than startup. Perhaps we should consider it in a separate branch.)

PR #1484, opened 2015-04-28
return from Protocol.read0 synchronously when possible

mykmelez:
I've tested this locally and on Travis, and it includes a benchmark, which shows up to a 5x improvement in performance when using the socket API to retrieve a file from localhost via HTTP.


mykmelez:
Note: to run the benchmark, you have to load it as a midlet so it has access to the J2ME socket API: index.html?jars=bench/benchmark.jar:tests/tests.jar&midletClassName=benchmark.SocketBench&logLevel=log

mykmelez:
@mbebenita I'd be interested to know how this change affects unwinds for you.

mbebenita:
With this patch consistently see 17 unwinds when executing the benchmark:

Bytecode:	7,411
Compiled:	0/15/105/0
Interpreter:	239
OSR:	8
Unwind:	17
Preemption:	0

mykmelez:
This is ready, but I think it should still wait until @brendandahl's thread preemption work lands, so it doesn't inadvertently regress UI thread responsiveness.

mykmelez:
> This is ready, but I think it should still wait until @brendandahl's thread preemption work lands, so it doesn't inadvertently regress UI thread responsiveness.

This should be ready to land now that #1519 is landed. But when I test it with the StressBench benchmark from #1529, it causes the UI to be less responsive. Perhaps that's ok, because we're really stressing the VM in that benchmark, and the UI is still somewhat responsive, and evenly so: the animation is relatively smooth, without fits and starts.

@brendandahl Should we tune the scheduler to keep higher-priority threads like the UI thread more responsive under this kind of load?

mykmelez:
This almost completely starves the UI thread in the StressBench benchmark. How do we ensure those threads get enough time after this change?

brendandahl:
I'll take a look after I figure out the regression. Are you seeing preemptions happen during the benchmark?

mykmelez:
> I'll take a look after I figure out the regression. Are you seeing preemptions happen during the benchmark?

Yes, I see a ton of them if I enable the thread writer.

mykmelez:
@TimAbraldes I'd be interested to know how the benchmark performs on this branch after any changes you make to the scheduler.

mykmelez:
This request is still relevant, although it's still blocked on fixing UI thread responsiveness. I'm going to test it with @TimAbraldes's recent scheduler changes to see if they make a difference.

mykmelez:
@TimAbraldes I tested with https://github.com/TimAbraldes/pluotsorbet/tree/ctxswitch, but it doesn't seem to make a difference in UI thread responsiveness on this branch.

TimAbraldes:
> @TimAbraldes I tested with https://github.com/TimAbraldes/pluotsorbet/tree/ctxswitch, but it doesn't seem to make a difference in UI thread responsiveness on this branch.

Hmm, bummer :(

I'll test this out as well; maybe this is an opportunity to tune the scheduler.

Also the situation might improve when the compiler is working

TimAbraldes:
I think the results of the benchmark that I created in #1729 are relevant here. Preemptions fail to occur under certain circumstances, and as a result our VM is unresponsive (read: fails to run the GUI thread updating the screen) if another thread is doing intensive computation.

We have an underlying bug - which I've filed as #1863 - that threads are able to monopolize computation and, in particular, starve the GUI thread.

I think we should determine whether the change in this PR affects real MIDlets that we care about. If it causes #1863 to be more salient in real MIDlets, then we should hold off on landing it. If it only makes #1863 more salient in a benchmark, then we should modify the benchmark to be aware of #1863.

Either way, some MIDlet out there is likely to exhibit a starved GUI thread as long as our underlying bug exists.

mykmelez:
> I think we should determine whether the change in this PR affects real MIDlets that we care about.

WhatsApp downloads some large files on startup, and I suspect that it slurps down their data in a loop that repeatedly checks if data is available on the input stream. So if your network is fast enough, and data is always (or usually) available, then it'll often block the UI thread while slurping down the data.

mykmelez:

PR #1490, opened 2015-04-29
Add bouncy castle SHA-256 benchmark.

brendandahl:
Uses the same library as a popular midlet.

### DIFF
--- a/bench/BouncyCastleSHA256.java
+++ b/bench/BouncyCastleSHA256.java
@@ -1,0 +1,31 @@
+package benchmark;
+
+import com.sun.cldchi.jvm.JVM;
+import org.bouncycastle.crypto.digests.SHA256Digest;
+
+public class BouncyCastleSHA256 {
+
+    private static final int UPDATES = 170;
+
+    public static void main(String args[]) {
+        // Try to mimic a popular midlet that commonly does around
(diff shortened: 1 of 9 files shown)


marco-c:
Could you move the bouncycastle code to the tests/ directory and file an issue to add a bouncycastle sha256 test? (It's unlikely but it'd be better to ensure it keeps working after our optimizations).

bench/lcrypto-j2me-152/LICENSE.html
```diff
@@ -0,0 +1,22 @@
+<html>
```
mykmelez: Nit: could you convert this to text and rename it to LICENSE? That'll keep it consistent with all the other LICENSE files in the repo.

bench/BouncyCastleSHA256.java
```diff
@@ -0,0 +1,31 @@
...
+import org.bouncycastle.crypto.digests.SHA256Digest;
+
+public class BouncyCastleSHA256 {
+
+    private static final int UPDATES = 170;
+
+    public static void main(String args[]) {
+        // Try to mimic a popular midlet that commonly does around
+        // 170 update calls on 4096 bytes at a time.
+        int updates = 170;
```
marco-c: Nit: unused variable

PR #1492, opened 2015-04-29
prioritize foreground runtime threads more highly

mykmelez:
This is a terrible, horrible, no good, very bad hack. It also gives me numbers like 2-7, 1-6, and 1-4 on the test I described at https://github.com/mozilla/j2me.js/pull/1457#issuecomment-97487066.

The hack is to add the runtime ID to each thread's priority when enqueuing it, so a thread in runtime 2 with the same priority as a thread in runtime 1 will run first, even if it's enqueued second. For midlet suites that start a background midlet and then a foreground midlet, this gives the foreground midlet's threads a bit more priority.

Of course this is an awful solution to the problem. It doesn't account for other midlets and has a hardcoded limit on runtimes, so it'll crash and burn on midlets with more than 5 isolates (presumably why it doesn't pass tests).

@brendandahl, @marco-c, @mbebenita: How do we do this in some reasonable fashion?


mykmelez:
987d46f prioritizes by isolate priority, then thread priority, and it assigns the foreground midlet Isolate.MAX_PRIORITY. It seems to work, although I haven't tested extensively. But times in the aforementioned manual test are 1-12, 1-11, and 1-5, which are not quite as good as I was seeing with the hack.

Next I'll try doing something similar to the hack, but using isolate priority to give threads a small bump.

mykmelez:
> Next I'll try doing something similar to the hack, but using isolate priority to give threads a small bump.

I tried that on a separate branch (https://github.com/mozilla/j2me.js/compare/master...mykmelez:prioritize-runtimes-slightly), and it gave me similar numbers, except that I didn't see the button animate, so I only measured the time to second screen: 10, 9, 9.

So the initial hack seems the most performant (and correct). I just did another round, and I'm still seeing good numbers for it: 1-6, 1-5, 1-5. The timings are strange, since the hack should be identical to using that branch (in terms of the relative prioritization of threads). But maybe calling *getIsolatePriority* every time we enqueue is expensive. The lack of button animation is doubly strange. I did see something similar on this branch, but only once.

marco-c:
These might be timing issues. Maybe if you tried the branches on a different device / different OS version you could get different results.

vm/runtime.ts
```diff
@@ -421,7 +421,7 @@ module J2ME {
     ctx: Context;
     allCtxs: Set<Context>;
 
-    isolate: com.sun.cldc.isolate.Isolate;
+    isolate: any;
```
marco-c: Why is this *any* now?

java/custom/com/nokia/mid/s40/bg/BGUtils.java
```diff
@@ -33,7 +33,7 @@ static void startMIDlet() {
       }
 
       AmsUtil.executeWithArgs(MIDletSuiteStorage.getMIDletSuiteStorage(), 0, BGUtils.getFGMIDletNumber(),
-                              BGUtils.getFGMIDletClass(), null, null, null, null, -1, -1, -1, null, false);
+                              BGUtils.getFGMIDletClass(), null, null, null, null, -1, -1, 3, null, false);
```
marco-c: Nit: Could you use Isolate.MAX_PRIORITY to make it clearer what the value is?

vm/runtime.ts
mykmelez: > Why is this any now?

I reverted this change in e854d96, since the code no longer references runtime.isolate._priority, but I did something similar in aa6cbc3, so you'll have the same question about that.

The reason is that the *com.sun.cldc.isolate.Isolate* type doesn't appear to be defined anywhere. Apparently, that's fine with TypeScript, as long as you don't attempt to access any of its properties. But once you do, the TypeScript compiler will complain the property:

> tsc --sourcemap --target ES5 references.ts -d --out bld/j2me.js
> vm/jvm.ts(55,34): error TS2339: Property '_priority' does not exist on type 'Isolate'.
> make: *** [bld/j2me.js] Error 3

I suppose there's a right thing to do here, like declaring that type. I just lack the TypeScript knowledge to immediately know what it is (and where to put it, etc.). And I figured that since we're using the type without actually defining it anywhere, it doesn't provide much value. But I may well be missing something.

java/custom/com/nokia/mid/s40/bg/BGUtils.java
mykmelez: > Nit: Could you use Isolate.MAX_PRIORITY to make it clearer what the value is?

Yes, good thinking, I changed that in b01d5c5.

vm/runtime.ts
marco-c: It is defined at https://github.com/mozilla/j2me.js/blob/master/bindings.ts.in#L428

mykmelez:
> I tried that on a separate branch (master...mykmelez:prioritize-runtimes-slightly), and it gave me similar numbers, except that I didn't see the button animate, so I only measured the time to second screen: 10, 9, 9.

The latest iteration merges prioritize-runtimes-slightly back into this branch, with the change that the isolate priority is cached in a property on the runtime object (*runtime.priority*) when the isolate is started. That way *PriorityQueue.enqueue* doesn't have to call a function to get the priority, it just needs to dereference that property.

Otherwise, the behavior is identical to the initial hack: the foreground midlet's threads get a +1 bump in priority, since the background midlet's isolate priority is *Isolate.NORM_PRIORITY* (2), while the foreground midlet's isolate priority is *Isolate.MAX_PRIORITY* (3).

In practice, the midlets I've tested seem to set their threads' priorities to either *Thread.NORMAL_PRIORITY* (5) or *Thread.MAX_PRIORITY* (10). So the +1 bump is not enough to make "normal" foreground threads more important than "max" background ones. It just makes foreground threads more important than background threads with the same priority.

With the latest changes, I see numbers similar to the ones I saw with the hack: 1-5, 4-4, 1-4. (The middle one showed the button animation a moment before the midlet changed screens.) I think this is ready.

vm/runtime.ts
mykmelez: > It is defined at https://github.com/mozilla/j2me.js/blob/master/bindings.ts.in#L428

Aha! I updated the binding in that file but missed the *com.sun.cldc.isolate.Isolate* declaration, since the file doesn't declare that type directly but rather declares a *com.sun.cldc.isolate* module with an *Isolate* interface. Fixed in d61cb1f.

mykmelez:
0cdf62a pads console output to accommodate the runtime/thread priorities I added to the runtime/thread ID prefix. Now the messages should line up properly in common cases:

```
1.2:7.5  | UNKNOWN PROPERTY (java/lang/System): com.nokia.multisim.imsi.sim2 
2.3:8.5  | UNKNOWN PROPERTY (java/lang/System): c 
2.3:9.10 | javax/microedition/lcdui/CanvasLFImpl.createNativeResource0.(Ljava/lang/String;Ljava/lang/String;)I not implemented 
2.3:34.5 | RecordStoreRegistry.stopRecordStoreListening.(IL...String;)V not implemented (1, Settings) 
2.3:9.10 | Display.drawTrustedIcon0.(IZ)V not implemented (184549377, 1) 
```

We could also remove the priorities from those prefixes. They were very useful while I was developing this changeset, but I'm not sure how useful they'll be in the future.

marco-c:
> So the +1 bump is not enough to make "normal" foreground threads more important than "max" background ones. It just makes foreground threads more important than background threads with the same priority.

Maybe we should consider making the FG threads always more important than BG threads.

mykmelez:
> Maybe we should consider making the FG threads always more important than BG threads.

I did that earlier in the branch, in 987d46f. Per https://github.com/mozilla/j2me.js/pull/1492#issuecomment-97603561, the numbers weren't as good. But that could have been from *Runtime.getIsolatePriority*.

The comment on *Isolate.setPriority* in Isolate.java says:

>Adjust the priority of this Isolate. The priority controls the
>amount of CPU time that VM allocates to execute threads in this
>Isolate.
>
>Note: thread scheduling and task scheduling use separate mechanisms.
>In the current imeplentation, each task is guaranteed execution time
>relative to its priority.

That's vague, but it suggests to me that referenced VM splits up processor time first by isolate, then by thread. And a high-priority isolate will get more time. But a lower-priority isolate will still always get some, regardless of how greedy the high-priority isolate's threads are.

In our case, we can only simulate that behavior. And I'm unsure which version is a closer simulation. But I'm inclined to do what we're doing now first, since it's a more conservative approach; and then try the more aggressive approach if we still see BG threads still starving FG threads.

(Having said that, I'll test out the more aggressive approach now that we have a faster way to determine runtime priority, to see if it makes a difference.)

mykmelez:
This shouldn't affect the startup benchmark, but just to be sure, I ran it with a complex midlet on a Flame with FxOS 2.1, and the numbers are insignificant:

| User Agent       | Mozilla/5.0 (Mobile; rv:34.0) Gecko/34.0 Firefox/34.0 | 

|       Test      | Baseline Mean |    Mean    |    +/-   |     %    |     P    |     Min    |     Max    | 
|:----------------|--------------:|-----------:|---------:|---------:|---------:|-----------:|-----------:|
| startupTime     |      12,137ms |   11,967ms |   -171ms |    -1.40 |     SAME |   10,958ms |   15,058ms | 
| vmStartupTime   |       2,266ms |    2,223ms |    -43ms |    -1.88 |   BETTER |    2,073ms |    2,410ms | 
| bgStartupTime   |       1,840ms |    1,843ms |      3ms |     0.17 |     SAME |    1,721ms |    2,618ms | 
| fgStartupTime   |       8,032ms |    7,901ms |   -131ms |    -1.63 |     SAME |    6,920ms |   10,234ms | 
| *totalSize*     |    *22,998kb* | *22,389kb* | *-609kb* |  *-2.65* | *BETTER* | *21,505kb* | *23,266kb* | 
| *domSize*       |        *71kb* |     *71kb* |    *0kb* |   *0.12* |  *WORSE* |     *71kb* |     *71kb* | 
| *styleSize*     |       *297kb* |    *300kb* |    *4kb* |   *1.26* |  *WORSE* |    *291kb* |    *305kb* | 
| *jsObjectsSize* |    *15,573kb* | *15,501kb* |  *-72kb* |  *-0.46* | *BETTER* | *15,477kb* | *15,525kb* | 
| *jsStringsSize* |       *946kb* |    *653kb* | *-293kb* | *-31.01* | *BETTER* |    *652kb* |    *655kb* | 
| *jsOtherSize*   |     *5,911kb* |  *5,639kb* | *-272kb* |  *-4.61* | *BETTER* |  *4,757kb* |  *6,455kb* | 
| *otherSize*     |       *200kb* |    *225kb* |   *25kb* |  *12.41* |  *WORSE* |    *166kb* |    *271kb* | 
| *USS*           |    *50,795kb* | *49,967kb* | *-828kb* |  *-1.63* | *BETTER* | *48,340kb* | *51,596kb* | 
| *peakRSS*       |         *0kb* |    *NaNkb* |  *NaNkb* |    *NaN* |   *SAME* |    *NaNkb* |    *NaNkb* |

vm/runtime.ts
```diff
@@ -653,21 +654,25 @@ module J2ME {
...
-      for (var i = MIN_PRIORITY; i <= MAX_PRIORITY; i++) {
+      for (var i = MIN_PRIORITY + ISOLATE_MIN_PRIORITY; i <= MAX_PRIORITY + ISOLATE_MAX_PRIORITY; i++) {
         this._queues[i] = [];
       }
     }
 
     enqueue(ctx: Context) {
-      var priority = ctx.getPriority();
-      release || assert(priority >= MIN_PRIORITY && priority <= MAX_PRIORITY,
+      var priority = ctx.getPriority() + ctx.runtime.priority;
```
marco-c: Can you try using ctx.runtime.isolate._priority instead? It shouldn't be much slower and should make it possible to change the Isolate priority at runtime (although we probably will never do it).

marco-c:
Yeah, when I said we should consider always prioritizing the FG MIDlet I meant we should file an issue to investigate it (also to keep in mind that changes to the execution order might make one approach better than the other), in the meantime I think we can merge this.

mykmelez:
Here's a branch that extends this branch to prioritize aggressively:

https://github.com/mykmelez/j2me.js/compare/prioritize-runtimes...mykmelez:prioritize-runtimes-aggressively

My numbers are 1-4, 1-4, 1-4 on this branch, so about the same, although maybe a little faster (felt that way, but hard to be sure given the imprecision of clock timings).

vm/runtime.ts
mykmelez: I tried it on this branch:

https://github.com/mykmelez/j2me.js/compare/mykmelez:prioritize-runtimes...mykmelez:prioritize-runtimes-through-isolate

But PriorityQueue.enqueue threw `TypeError: ctx.runtime.isolate is undefined` for some reason, so I had to check if it existed first.

Then it worked, but the numbers were 8, 9, 8, i.e. the button never animated, but the second screen appeared after that number of seconds, which is 2x slower. I'm not sure why the button didn't animate, but presumably the perf regression is because of the conditional.

If we wanted to ensure that it's possible to change the Isolate priority at dynamically, we might be able to update its priority in *Isolate.setPriority0*. We can't do that when the Isolate is initially created, because *Isolate.setPriority0* is called before *JVM.startIsolate*, which is what creates the runtime. But we could do it later. Nevertheless, I don't think there's any code that sets the Isolate's priority later, so I don't think we need to worry about this.

PR #1495, opened 2015-04-30
Fix building with TypeScript 1.5 beta

marco-c:


mykmelez:
I wonder if it's better to build with a specific version of TypeScript rather than simply the latest version. I guess this way we find out early when TypeScript breaks, so let's stick with it for now.

marco-c:
I was wondering that as well, but then decided it was better to detect breakage as early as possible (for example we wouldn't have detected this problem if we had used a fixed version), so that fixes are supposed to be simpler.
If breakages happen too often, then we can revise the decision and use a fixed version.

PR #1499, opened 2015-05-01
Optimize socket implementation by avoiding to needlessly copy data

marco-c:


mykmelez:
Changes look good, and socket API perf seems somewhat improved on the SocketBench that recently landed.

PR #1513, opened 2015-05-06
Bring back the gamepad

TimAbraldes:
Properly set the "gamepad" class on the document element when the
gamepad parameter is specified.

Fix broken CSS related to gamepad.

### DIFF
--- a/midp/midp.js
+++ b/midp/midp.js
@@ -203,10 +203,6 @@ var MIDP = (function() {
     return bytes;
   };
 
-  if (config.gamepad && !/no|0/.test(config.gamepad)) {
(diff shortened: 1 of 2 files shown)


TimAbraldes:
This PR might work without #1514 but it's probably better to land them together.

midp/midp.js
```diff
@@ -203,6 +203,10 @@ var MIDP = (function() {
     return bytes;
   };
 
+  if (config.gamepad && !/no|0/.test(config.gamepad)) {
```
marco-c: This snippet of code is already in main.js

TimAbraldes:
This is updated with review comments. I can rebase before landing

PR #1514, opened 2015-05-06
Clean up MIDP foreground info

TimAbraldes:
Create an `FG` module inside of `MIDP` that keeps track of the
foreground isolate ID and the foreground display ID. Send native events
through a `FG` function so that we always properly check that the
display ID and isolate ID are valid.

Make the gamepad buttons send both press and release events. Use
mousedown and mouseup events instead of touchstart events.

Remove the unused SCREEN_REPAINT_EVENT constant.

Rename `MIDP.keyPress` and `MIDP.keyRelease` to `MIDP.send*`. This is
more consistent with other similar functions.

Remove the `MIDP.displayId` property.

### DIFF
--- a/game-ui.js
+++ b/game-ui.js
@@ -1,11 +1,39 @@
-function makeSendKey(keyCode) {
-  return function() {
-    MIDP.keyPress(keyCode);
-  }
+document.getElementById("up").onmousedown = function() {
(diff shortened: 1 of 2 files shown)


TimAbraldes:
Rebased

PR #1516, opened 2015-05-07
Don't play tones when Alerts are shown

marco-c:


java/custom/javax/microedition/lcdui/Display.jpp
```diff
@@ -2096,46 +2096,6 @@ public class Display {
      * @return <code>true</code>, if sound was played.
      */
     boolean playAlertSound(AlertType t) {
-
```
mykmelez: Can you add a comment explaining that we've removed the implementation of this method because the sounds are an annoyance? It'll be helpful for someone looking at this code later to understand why the method isn't implemented.

Issue #1384, opened 2015-04-02
`Graphics` objects representing the screen should not interfere with each other


TimAbraldes:
The `Display` object is the one and only consumer of the `Graphics.getScreenGraphics` method. It is intended that there is only ever one `Graphics` object per MIDlet representing the screen and that the `Display` object holds onto that reference.

Our implementation has trouble when there is more than one MIDlet (e.g. a foreground MIDlet and a background MIDlet). Each MIDlet's `Display` object has its own `Graphics` object representing the screen. In our implementation, those `Graphics` objects share an HTML Canvas Context, so when one MIDlet (e.g. the background MIDlet) changes something (e.g. the clipping rectangle, the font, the fill color, etc), it affects the foreground MIDlet's drawing operations.

TimAbraldes:
I think this has actually been fixed by #1395.

PR #1395, opened 2015-04-07
Fix objects returned from `Image.getGraphics`

TimAbraldes:
Fixes mozilla/j2me.j2#1383.

For each canvas context (1 for every `ImageData` object and 1 for the
screen canvas), keep track of what `Graphics` object has been applied to
it. Whenever a property of the currently applied `Graphics` object
changes, remove the association of that object with the context since
they are no longer in sync.

Add a test that verifies that `Graphics` objects created from distinct
calls to `Image.getGraphics` on the same `Image` object do not interfere
with each other.

PR #1517, opened 2015-05-07
Move fullscreen management from gfx to midp

TimAbraldes:
This applies over #1513 and #1514

### DIFF
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -15,25 +15,6 @@ var currentlyFocusedTextEditor;
     tempContext.canvas.width = 0;
     tempContext.canvas.height = 0;
 
-    var NativeDisplay = function() {
-        this.fullScreen = 1;
-    };
-
-    var NativeDisplays = {
-        get: function(id) {
-            var d = this._map.get(id);
-            if (!d) {
-                d = new NativeDisplay();
-                this._map.set(id, d);
-            }
-            return d;
-        },
-
-        _map: new Map(),
-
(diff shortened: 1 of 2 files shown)


TimAbraldes:
Rebased onto just #1514

mykmelez:
Is this just cleanup, or is there another benefit we should expect to get from it?

TimAbraldes:
> Is this just cleanup, or is there another benefit we should expect to get from it?

This was preparatory for work I thought I was going to need to do to fix #1384. It seems like #1395 may actually have fixed #1384 so this is mostly cleanup now.

This change has some minor benefits like the fact that it moves our handling of switching to/from fullscreen from `com/sun/midp/lcdui/DisplayDevice.gainedForeground0.(II)V` to `com/sun/midp/main/MIDletProxyList.setForegroundInNativeState.(II)V` - which will probably make the switch happen in a slightly more timely fashion - and the fact that our fullscreen state will correctly reflect a "reset" of the foreground display since we now update fullscreen state in `com/sun/midp/main/MIDletProxyList.resetForegroundInNativeState.()V` - before we only handled `com/sun/midp/lcdui/DisplayDevice.gainedForeground0.(II)V` which probably doesn't get called during a reset.

I'm not aware of any user-visible effects of this patch, but if they exist they're likely to be in MIDlets that are not fullscreen.

mykmelez:
Got it, thanks @TimAbraldes!

PR #1512, opened 2015-05-06
Load foreground MIDlet up to startApp when app started in the background

marco-c:
According to my measurements, this saves ~1 second on warm startups.

PR #1519, opened 2015-05-08
Enable preemption and use a fair scheduler.

brendandahl:
Still needs a bit of work to change the failing tests.


tests/automation.js
```diff
@@ -237,33 +237,41 @@ casper.test.begin("unit tests", 25 + gfxTests.length, function(test) {
...
+            ""];
+          output = output.split("\n");
+          test.assert(expectedOutput.length === output.length, "Same number of lines output.");
+          var allMatch = true;
+          for (var i = 0; i < output.length; i++) {
+            if (expectedOutput.indexOf(output[i]) === -1) {
+              allMatch = false;
+              break;
+            }
+          }
```
mykmelez: This is unlikely to produce any false positives, although it's theoretically possible if the test repeats one line while skipping another. To eliminate that possibility, you could sort both arrays and test that they have the same strings at the same indexes.

mykmelez:
I'm not sure why Travis didn't build the latest commit. But I merged the branch into a branch in my fork, and Travis built it there:

https://travis-ci.org/mykmelez/j2me.js/builds/61800417

The two main test suites (automation.js and fs/automation.js) both passed, but this error shows up at the end of the log, where the Basic jsshell test should output "The end,":

>TypeError: window.setTimeout is not a function
>Scheduler</Scheduler.processRunningQueue@/home/travis/build/mykmelez/j2me.js/bld/j2me.js:5280:13
>Scheduler</Scheduler.enqueue@/home/travis/build/mykmelez/j2me.js/bld/j2me.js:5236:13
>Context</Context.prototype.resume@/home/travis/build/mykmelez/j2me.js/bld/j2me.js:8908:13
>Context</Context.prototype.start@/home/travis/build/mykmelez/j2me.js/bld/j2me.js:8876:13
>JVM</JVM.prototype.startIsolate0@/home/travis/build/mykmelez/j2me.js/bld/j2me.js:4813:1
>@jsshell.js:146:17

mykmelez:
I ran the startup benchmark with a midlet on desktop and device, and the desktop results were inconsistent (one run showed a perf regression, the other showed no change), although my machine was under load at the time, so the results may not be reliable. Results from a run on device were favorable (with interesting deltas in the FG vs. BG midlets):

| User Agent       | Mozilla/5.0 (Mobile; rv:40.0) Gecko/40.0 Firefox/40.0 | 

|       Test      | Baseline Mean |     Mean    |     +/-    |    %    |     P    |     Min     |     Max     | 
|:----------------|--------------:|------------:|-----------:|--------:|---------:|------------:|------------:|
| startupTime     |      12,966ms |    12,058ms |     -908ms |   -7.00 |   BETTER |    10,465ms |    13,615ms | 
| vmStartupTime   |       2,539ms |     3,005ms |      466ms |   18.35 |    WORSE |     2,795ms |     3,188ms | 
| bgStartupTime   |         893ms |     1,257ms |      364ms |   40.72 |    WORSE |     1,133ms |     1,503ms | 
| fgStartupTime   |       9,534ms |     7,796ms |   -1,738ms |  -18.23 |   BETTER |     6,094ms |     9,399ms | 
| *totalSize*     |    *20,114kb* |  *20,013kb* |   *-101kb* | *-0.50* |   *SAME* |  *19,272kb* |  *20,938kb* | 
| *domSize*       |        *73kb* |      *73kb* |      *0kb* |  *0.02* |   *SAME* |      *72kb* |      *73kb* | 
| *styleSize*     |       *275kb* |     *276kb* |      *1kb* |  *0.30* |   *SAME* |     *269kb* |     *282kb* | 
| *jsObjectsSize* |    *14,704kb* |  *14,575kb* |   *-129kb* | *-0.88* | *BETTER* |  *14,289kb* |  *14,720kb* | 
| *jsStringsSize* |       *572kb* |     *561kb* |    *-11kb* | *-1.94* | *BETTER* |     *543kb* |     *572kb* | 
| *jsOtherSize*   |     *4,255kb* |   *4,270kb* |     *14kb* |  *0.33* |   *SAME* |   *3,453kb* |   *5,019kb* | 
| *otherSize*     |       *235kb* |     *258kb* |     *24kb* | *10.15* |  *WORSE* |     *212kb* |     *319kb* | 
| *USS*           |    *51,368kb* |  *51,368kb* |      *0kb* |  *0.00* |   *SAME* |  *49,640kb* |  *52,488kb* | 
| *peakRSS*       |   *125,807kb* | *121,822kb* | *-3,985kb* | *-3.17* | *BETTER* | *114,724kb* | *123,608kb* | 

I bet there's room for even more improvement with tuning.

libs/fs.js
```diff
@@ -521,13 +521,15 @@ var fs = (function() {
       from = buffer.contentSize;
     }
 
-    var newLength = (from + data.byteLength > buffer.contentSize) ? (from + data.byteLength) : (buffer.contentSize);
+    var newLength = (from + length > buffer.contentSize) ? (from + length) : (buffer.contentSize);
 
     buffer.setSize(newLength);
 
-    buffer.array.set(data, from);
+    for (var i = 0; i < length; i++) {
```
marco-c: As in #1457, we should use *set* when the amount of data to write is big.

mykmelez:
This also fixes the problem I experienced with #1512, although @marco-c found a bug in that branch that may be the underlying cause of that problem.

On the benchmark in #1529, however, the Worker thread eventually dies with:

```
09:38:23.325 Error: java/lang/NullPointerException: current is null
 - java/lang/RuntimeException.<init>(Ljava/lang/String;)V, pc=undefined
 - java/lang/NullPointerException.<init>(Ljava/lang/String;)V, pc=undefined
 - benchmark/StressBench$Worker.run()V, pc=undefined
 - java/lang/Thread.run()V, pc=undefined
 - org/mozilla/internal/Sys.runThread(Ljava/lang/Thread;)V, pc=undefined
 - benchmark/StressBench$Worker.run()V, pc=33
 - java/lang/Thread.run()V, pc=16
 - org/mozilla/internal/Sys.runThread(Ljava/lang/Thread;)V, pc=4

 j2me.js:314:412
09:38:23.325 Error: java/lang/NullPointerException: current is null
 - java/lang/RuntimeException.<init>(Ljava/lang/String;)V, pc=undefined
 - java/lang/NullPointerException.<init>(Ljava/lang/String;)V, pc=undefined
 - java/lang/Thread.run()V, pc=undefined
 - org/mozilla/internal/Sys.runThread(Ljava/lang/Thread;)V, pc=undefined
 - java/lang/Thread.run()V, pc=16
 - org/mozilla/internal/Sys.runThread(Ljava/lang/Thread;)V, pc=4

 j2me.js:314:412
```

brendandahl:
> On the benchmark in #1529, however, the Worker thread eventually dies with:

I wonder if there's an issue with sockets, because I've noticed that sometimes the browser gets in a strange state with the socket benchmark where no new tabs work.

mykmelez:
> I wonder if there's an issue with sockets, because I've noticed that sometimes the browser gets in a strange state with the socket benchmark where no new tabs work.

I see that too. The browser UI itself continues to work, but content never finishes loading. And it isn't just new tabs: reloading an existing tab also hangs.

mykmelez:
There's no *current* reference in this code that could generate the error. But there are other uses of *current* around the codebase. It'd be useful to know which one is actually responsible here.

mykmelez:
With RELEASE=0, I get a different error message:

```
11:06:43.971 Error: java/lang/NullPointerException: ctx is null
 - java/lang/RuntimeException.<init>(Ljava/lang/String;)V, pc=undefined
 - java/lang/NullPointerException.<init>(Ljava/lang/String;)V, pc=undefined
 - benchmark/StressBench$Worker.run()V, pc=undefined
 - java/lang/Thread.run()V, pc=undefined
 - org/mozilla/internal/Sys.runThread(Ljava/lang/Thread;)V, pc=undefined
 - benchmark/StressBench$Worker.run()V, pc=27
 - java/lang/Thread.run()V, pc=16
 - org/mozilla/internal/Sys.runThread(Ljava/lang/Thread;)V, pc=4

 j2me.js:7303:18
11:06:43.971 Error: java/lang/NullPointerException: ctx is null
 - java/lang/RuntimeException.<init>(Ljava/lang/String;)V, pc=undefined
 - java/lang/NullPointerException.<init>(Ljava/lang/String;)V, pc=undefined
 - java/lang/Thread.run()V, pc=undefined
 - org/mozilla/internal/Sys.runThread(Ljava/lang/Thread;)V, pc=undefined
 - java/lang/Thread.run()V, pc=16
 - org/mozilla/internal/Sys.runThread(Ljava/lang/Thread;)V, pc=4

 j2me.js:7303:18
```

mykmelez:
With 6fc63bf, I no longer see those exceptions. I do still see the issue that the Worker thread in StressBench stops attempting to download the file, but that problem exists on the stress-bench branch without these changes, so it isn't caused by them.

mykmelez:
With these changes, the UI in StressBench remains responsive while data is read from a socket in the background via the Worker thread. If I merge #1484, it still stays responsive, although it slows down noticeably. Either way, results are better than before this branch, so this is a win, and we should be able to tune it in conjunction with #1484 and #1457.

libs/fs.js
mykmelez: > As in #1457, we should use set when the amount of data to write is big.

I think this is the only remaining issue here. @brendandahl can you modify the *write* function to use *TypedArray.set* when `length > 128`? That seems to be a reasonable threshold based on @mbebenita's earlier tests.

brendandahl:
Couple of other things that may need to be addressed or be follow up issues:

- It's possible for a long sleeping thread to wake up and starve other threads because it will have a low virtualRuntime. A simple solution would be to set a waking thread's virtualRuntime to minVirtualRuntime less some amount. I have not seen this as an issue in real world midlet's yet.
- A long list in a midlet will now sometimes flicker because preemption happens during rendering. For this we may want to implement off screen drawing and only update on refresh0.

mykmelez:
https://www.traviscistatus.com/ says they are "investigating reports of stalled Linux builds in both container based and fully virtualized Linux infrastructure." But that doesn't explain why Travis hasn't been building any commit on this branch since the first one. In the case of a stalled build, it still creates the build; but on this branch it doesn't even create them.

In any case, Travis does create builds on my copy of this branch in my fork. [Build 516](https://travis-ci.org/mykmelez/j2me.js/builds/62146223) tests commit c63ae76, "Use set() for large copies." So we can use that build to determine the status of this branch.

mykmelez:
While I troubleshoot Travis's departure from this repository (and Travis troubleshoots their problems with Linux builds), I'm going to test branches locally. So I ran tests for this one, which reminded me that the jsshell test is still failing:

```
TypeError: window.setTimeout is not a function
Scheduler</Scheduler.processRunningQueue@/Users/myk/Sites/j2me.js/bld/j2me.js:5225:13
Scheduler</Scheduler.enqueue@/Users/myk/Sites/j2me.js/bld/j2me.js:5181:13
Context</Context.prototype.resume@/Users/myk/Sites/j2me.js/bld/j2me.js:8829:13
Context</Context.prototype.start@/Users/myk/Sites/j2me.js/bld/j2me.js:8797:13
JVM</JVM.prototype.startIsolate0@/Users/myk/Sites/j2me.js/bld/j2me.js:4766:1
@jsshell.js:146:17
```

mykmelez:
Tests now pass locally!

PR #1529, opened 2015-05-11
add mbx's socket/gfx stress benchmark to test UI responsiveness under load

mykmelez:
This is [mbx's StressBench benchmark](https://github.com/mbebenita/j2me.js/blob/protocol-read0-sync/bench/StressBench.java) from his https://github.com/mbebenita/j2me.js/tree/protocol-read0-sync branch of my https://github.com/mykmelez/j2me.js/tree/protocol-read0-sync branch.


bench/StressBench.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  static class Worker implements Runnable {
+
+    void benchmarkLargeRead() throws IOException {
+      SocketConnection client = (SocketConnection)Connector.open("socket://localhost:8000");
+
+      OutputStream os = client.openOutputStream();
+      os.write("GET /bench/benchmark.jar HTTP/1.1\r\nHost: localhost\r\n\r\n".getBytes());
```
brendandahl: Adding "Connection: close\r\n" here makes this test work with real webservers that keep sockets open for http 1.1.

mykmelez: > Adding "Connection: close\r\n" here makes this test work with real webservers that keep sockets open for http 1.1.

Good point, added in 13d783a.

mykmelez:
#1539 breaks this, so setting it to WIP until I resolve that regression.

mykmelez:
Now that *Display* draws to an offscreen buffer (#1539), *StressBench* needs to return from *Canvas.paint* calls in order for the display to refresh. That seems to reduce the UI's responsiveness noticeably in this benchmark (by my subjective judgement). I tried setting the UI thread's priority to MAX_PRIORITY, but that doesn't seem to make a difference.

brendandahl:
Notes from conversation with Myk:
The lag of the UI updating seems to come from execution window limit. When we have all the threads going, we repaint quickly but then the other threads run up to the 80ms window which causes a max repaint rate of once every 80ms.  Decreasing the execution window 16ms does improve things but then we unwind a lot and can't always finish a repaint for every frame.

mykmelez:
@brendandahl Any reason to hold off landing this? It'll be useful to diagnose remaining scheduler issues once we get back to those (and doesn't do any harm in the meantime, since it isn't run by default).

PR #1539, opened 2015-05-11
Implement refresh0.

brendandahl:
Depends on #1519

### DIFF
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -5,10 +5,17 @@
 
 var currentlyFocusedTextEditor;
 (function(Native) {
-    var screenContextInfo = new ContextInfo(MIDP.context2D);
-    MIDP.context2D.canvas.addEventListener("canvasresize", function() {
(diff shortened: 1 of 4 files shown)


marco-c:
How does perf in general look like after #1519?

mykmelez:
> How does perf in general look like after #1519?

In my most recent tests, the UI is more responsive, but it's also more flickery.

mykmelez:
#1519 has landed, and this now has conflicts with master. @brendandahl can you merge from master and resolve those conflicts?

brendandahl:
> How does perf in general look like after #1519?

Ask for a link to some notes.

mykmelez:
> Ask for a link to some notes.

Erm, can you elaborate?

mykmelez:
This resolves the flicker regression that was introduced by #1519. The perf impact isn't entirely clear, but GraphicsBench outputs similar numbers before and after this change, and tests of a complex midlet on a device suggest that the UI appears to be just as responsive, if not more so, afterward.

PR #1544, opened 2015-05-12
profile warm startup

mykmelez:
This branch adds a third *PROFILE* option, `PROFILE=3`, that profiles warm startup and outputs a text file. It works on both desktop and device, and it may be able to profile warm startup even if you don't start the app in the background (although I'm not sure how accurate it is in that case).

I also updated the README to document this option and correct some obsolete information about the method timeline, which is now enabled by default.

### DIFF
--- a/midp/background.js
+++ b/midp/background.js
@@ -91,9 +91,24 @@ Native["com/nokia/mid/s40/bg/BGUtils.maybeWaitUserInteraction.(Ljava/lang/String
 
   // If the page is visible, just start the FG MIDlet
   if (!document.hidden) {
-    profile === 3 && startTimeline();
     showSplashScreen();
     hideBackgroundScreen();
+
+    if (profile === 3) {
+      // Start the "warm startup" profiler after a timeout to better imitate
+      // what happens in a warm startup.  Even with this timeout, the profile
+      // won't be very accurate, and you should profile warm startup by letting
+      // the app start in the background, waiting until the background midlet
+      // settles down, and then opening the app.
+      console.warn("imitating warm startup time for profiler");
+      asyncImpl("V", new Promise(function(resolve, reject) {
+        setTimeout(function() {
+          startTimeline();
+          resolve();
+        }, 0);
+      }));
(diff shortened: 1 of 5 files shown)


mykmelez:
Travis is still busted for the central repo, but here's the build for the latest change in this branch on my fork: https://travis-ci.org/mykmelez/j2me.js/builds/62188183.

midp/background.js
```diff
@@ -91,6 +91,7 @@ Native["com/nokia/mid/s40/bg/BGUtils.maybeWaitUserInteraction.(Ljava/lang/String
 
   // If the page is visible, just start the FG MIDlet
   if (!document.hidden) {
+    profile === 3 && startTimeline();
```
marco-c: To make it accurate, you could add an asyncImpl with a timeout if profile === 3.

mykmelez: > To make it accurate, you could add an asyncImpl with a timeout if profile === 3.

I did that in 16addc27f3338f784d26af4b1a76f8a1fcea791e, but the numbers aren't much closer. Here are some measurements of the time spent in *CldcMIDletSuiteLoader.startSuite* (in ms) under various scenarios:

real warm run: 15239, 13043, 15170
fake warm run: 20478, 19726, 20606
fake warm run with setTimeout: 17258, 17810, 19247

marco-c: Actually, I meant a longer timeout (maybe configurable?) to give the BG MIDlet time to settle.

marco-c: To automate the "wait until the BG MIDlet settles".

mykmelez: > Actually, I meant a longer timeout (maybe configurable?) to give the BG MIDlet time to settle.

Heh, that makes a lot more sense. :-)

Indeed, a short timeout gives us numbers that are similar to an actual warm startup. With a 10000ms timeout, I get: 14797, 14019, 14364. With a 5000ms timeout, I get: 14246, 13701, 13171. I've set a 5000ms timeout in cf72a04.

PR #1545, opened 2015-05-12
Misc localmsg optimizations

marco-c:


midp/localmsg.js
```diff
@@ -1071,7 +1081,12 @@ Native["org/mozilla/io/LocalMsgConnection.receiveData.([B)I"] = function(data) {
         console.warn("receiveData from an unimplemented localmsg server (" + this.protocolName + ")");
     }
 
-    asyncImpl("I", this.connection.clientReceiveMessage(data));
+    var ret = this.connection.getClientMessage(data);
+    if (ret > 0) {
+      return ret;
+    }
```
mykmelez: Could *msg.length* ever be `0`? The old implementation always copied the message and returned its length, while this one only does so if `msg.length > 0`. Meanwhile, the return value here will actually be `-1` if there isn't wasn't a message to retrieve. So perhaps this should be `if (ret > -1)`. But I would refactor to avoid the return value ambiguity and continue to return a message regardless of its length.

mykmelez:
The Travis build for the latest change in this PR is https://travis-ci.org/mozilla/j2me.js/builds/62231091.

PR #1551, opened 2015-05-13
Fix localmsg race condition

marco-c:
This race condition happens when the client tries to connect to a server that hasn't yet been created.
In this case, the server was never returning from *acceptAndOpen*.


midp/localmsg.js
```diff
@@ -1046,6 +1049,10 @@ Native["org/mozilla/io/LocalMsgConnection.init.(Ljava/lang/String;)V"] = functio
 };
 
 Native["org/mozilla/io/LocalMsgConnection.waitConnection.()V"] = function() {
+    if (this.connection.clientConnected) {
+      return;
+    }
```
mykmelez: Nit: inconsistent indentation.

PR #1555, opened 2015-05-14
Remove macro byte codes to fix interpreter only mode.

brendandahl:
Let's remove for now since a popular midlet is broken when running in interpreter only mode with optimizations enabled. Happy to add back if there is some perf increase.


marco-c:
Could we simply disable it with a *enableMethodBytecodeOptimizations* instead of removing it altogether?

marco-c:
@brendandahl ping

mykmelez:

PR #1559, opened 2015-05-14
add benchrunner script for running benchmake from commandline

wfwalker:


mykmelez:
Nit: instead of tabs we usually use spaces for indentation (often 2-space indentation, although some files use 4-space).

PR #1561, opened 2015-05-15
In the startup benchmark, measure up to the first call to refresh0; increase maximum startFG delay

marco-c:


mykmelez:
Nit: the long name pushes the table even wider. It's already too wide for a GitHub comment again, though, so we can tackle this separately. Filed as #1562.

Also, while looking at this, I noticed that the startup time parts don't add up to the whole, both with and without this change. Filed as #1563.

PR #1564, opened 2015-05-18
Add a new customizable profiling mode

marco-c:
This can be used to select a profiling window by adding calls to Sys.startProfile and Sys.stopProfile.
For example I've used it to get a profile of the first call to Canvas::paint in CanvasLFImpl.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ To use them, just add calls to `runtimeCounter.count(name, count = 1)`. To view 
   }
   ```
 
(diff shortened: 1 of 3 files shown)


mykmelez:
Can you add some documentation about this profiling mode into the *Profiling > VM profiling* section of the README? Right now there are no other references than in code (and only in code that is separate from the other code that checks the *profile* variable), so it's hard to know it exists (and will be easy to trip over when adding another profiling mode).

marco-c:
Done in b527ebe.

mykmelez:
> Done in b527ebe.

One more addition is to include "4" in the list of possible modes in this statement earlier in the section:

> To use it, build j2me.js with PROFILE=[1|2]…

mykmelez:
It's funny that the mode options are currently 1, 2, and 4. It reminds me of the sequence guessing game. Perhaps the next option will be 8! ;-)

PR #1565, opened 2015-05-18
pause method timeline while awaiting async resolution

mykmelez:
This WIP "pauses" the profiler's method timeline while a context is blocked on asynchronous resolution of a method, so it's easier to distinguish between async methods that block for a long time and methods that are themselves expensive.

WIP for cleanup, because I'd like to ponder the implementation approach a bit more, and possibly to implement aggregation of method counts/times across threads, which would make that information more useful. I may also spin off a change or two into a separate branch.

### DIFF
--- a/vm/context.ts
+++ b/vm/context.ts
@@ -317,6 +317,12 @@ module J2ME {
 
     id: number;
 
+    /**
+     * Whether or not the context is currently paused.  The profiler uses this
+     * to distinguish execution time from paused time in an async method.
+     */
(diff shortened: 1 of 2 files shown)


mykmelez:
I've squashed the changes and removed those orthogonal to the core one here, which is the addition of a `<pause>` event in thread timelines while contexts are blocked on async resolution (or have yielded to other threads).

So this is now ready for review, and I'll request separate pulls for the other changes.

PR #1567, opened 2015-05-18
Support destroying/relaunching MIDlets

marco-c:


marco-c:
The failure seems unrelated to me.

marco-c:
The CanvasWithHeaderTest test is failing:
canvas

The reference image is:
canvaswithheadertest

There's probably some bad interaction with the mouse click in the test I've added and the CanvasWithHeaderTest test.

tests/native.js
```diff
@@ -234,3 +234,21 @@ Native["tests/recordstore/ReaderMIDlet.waitWriterWrote.()V"] = function() {
     }
   }));
 };
+
+Native["tests/background/DestroyMIDlet.sendDestroyMIDletEvent.(Ljava/lang/String;)V"] = function(midletClassName) {
```
mykmelez: Nit: it seems unnecessary for this native to take a *midletClassName* parameter, as it's intended to be used only by *tests.background.DestroyMIDlet*.

midp/midp.js
```diff
@@ -1261,6 +1289,9 @@ var MIDP = (function() {
     sendMediaSnapshotFinishedEvent: sendMediaSnapshotFinishedEvent,
     sendKeyPress: sendKeyPress,
     sendKeyRelease: sendKeyRelease,
+    sendDestroyMIDletEvent: sendDestroyMIDletEvent,
+    setDestroyedForRestart: setDestroyedForRestart,
```
mykmelez: Nit: although this has only one caller, and that caller only calls it to set *destroyedForRestart* to `true`, that isn't obvious here in this interface definition, and a *set** function conventionally lets you set a variable to the range of possible values for its type.

So I would make this function take a boolean argument, which it assigns to *destroyedForRestart*. (Alternately, you could rename it to something like *enableDestroyedForRestart* to make it clearer that the function only sets the variable to `true`.)

marco-c: Done in d27990d.

tests/native.js
marco-c: Done in 2a21111.

PR #1571, opened 2015-05-19
sort by self time descending

mykmelez:
Reversing the sort order makes the most expensive methods the most prominent ones:

```
Timeline Statistics: 3
java/lang/Object.wait.(J)V: count: 1, self: 1844.965 ms, total: 1844.965 ms.
…
com/sun/midp/main/CldcMIDletSuiteLoader.startSuite.()V: count: 1, self: 0.016 ms, total: 1845.145 ms.
```

### DIFF
--- a/shumway/profiler/timelineFrame.ts
+++ b/shumway/profiler/timelineFrame.ts
@@ -243,9 +243,9 @@ module Shumway.Tools.Profiler {
           s.push(this.statistics[i]);
         }
       }
-      // Sort by self time.
+      // Sort by self time descending, so expensive events are more prominent.
       s.sort(function (a, b) {
-        return a.selfTime - b.selfTime;
+        return b.selfTime - a.selfTime;
       })
       for (var i = 0; i < s.length; i++) {
         var statistic = s[i];


marco-c:
It could also be useful to sort by total time (for example if we wanted to find platform methods that would be worth removing/simplifying).

marco-c:
It'd be really nice to save the profile in JSON and build a profile viewer tool that allows us to dynamically inspect the profile.

mykmelez:
> It could also be useful to sort by total time (for example if we wanted to find platform methods that would be worth removing/simplifying).

Indeed. I had a change in my experimental branch that output in CSV format, which I then imported into a spreadsheet so I could sort by various columns. It's just this trivial change to *TimelineFrame.traceStatistics*:

```diff
+        if (csv) {
+          writer.writeLn([
+            (statistic.kind ? statistic.kind.name : "?"),
+            statistic.count,
+            statistic.selfTime.toFixed(3),
+            statistic.totalTime.toFixed(3)
+          ].join(","));
+        } else {
+          writer.writeLn(
+            (statistic.kind ? statistic.kind.name + ": " : "?: ") +
+            "count: " + statistic.count + ", " +
+            "self: " + statistic.selfTime.toFixed(3) + " ms, " +
+            "total: " + statistic.totalTime.toFixed(3) + " ms."
+          );
+        }
```

But I hadn't yet fully parameterized it. Would something like that address this use case well enough?


> It'd be really nice to save the profile in JSON and build a profile viewer tool that allows us to dynamically inspect the profile.

Yes, it would be great to have something similar to about:memory's import feature, i.e. a way to import a JSON profile into a page that produces the flame chart as well as a text representation of the profile.

marco-c:
Yes, I think CSV would be enough (the viewer would be very nice, but would require way more time).

marco-c:
Another thing that would really help understanding a profile, is to print what threads are blocked on when they're blocked (for example if thread A is waiting thread B).

mykmelez:
> Yes, I think CSV would be enough (the viewer would be very nice, but would require way more time).

Ok, I've finished this in #1574.

mykmelez:
> Another thing that would really help understanding a profile, is to print what threads are blocked on when they're blocked (for example if thread A is waiting thread B).

Indeed, that's a good idea. I'm unsure exactly how to do this, but we do pass a generic "reason" to *Runtime.pause* and *Runtime.yield*. Perhaps we can annotate those strings with additional info that the functions forward to *Context.pauseMethodtimeline*.

PR #1569, opened 2015-05-19
trace aggregate method timeline statistics

mykmelez:
This branch reports aggregate (in addition to per-thread) method timeline statistics when PROFILE=2|3, so you can see the overall cost of methods across all the threads in which they're invoked. For example, when profiling warm startup of a complex midlet, I see that `java/lang/Object.wait.(J)V` is called from a number of different threads:

```
java/lang/Object.wait.(J)V: count: 1, self: 1961.835 ms, total: 1961.835 ms.
java/lang/Object.wait.(J)V: count: 2, self: 1956.661 ms, total: 1956.661 ms.
java/lang/Object.wait.(J)V: count: 3, self: 1102.677 ms, total: 1102.677 ms.
java/lang/Object.wait.(J)V: count: 4, self: 1141.161 ms, total: 1141.161 ms.
java/lang/Object.wait.(J)V: count: 2, self: 1827.515 ms, total: 1827.515 ms.
java/lang/Object.wait.(J)V: count: 1, self: 1961.270 ms, total: 1961.270 ms.
java/lang/Object.wait.(J)V: count: 5, self: 1956.100 ms, total: 1956.100 ms.
java/lang/Object.wait.(J)V: count: 1, self: 1961.218 ms, total: 1961.218 ms.
java/lang/Object.wait.(J)V: count: 1, self: 1961.171 ms, total: 1961.171 ms.
java/lang/Object.wait.(J)V: count: 1, self: 1961.152 ms, total: 1961.152 ms.
java/lang/Object.wait.(J)V: count: 1, self: 1961.056 ms, total: 1961.056 ms.
java/lang/Object.wait.(J)V: count: 1, self: 836.142 ms, total: 836.142 ms.
java/lang/Object.wait.(J)V: count: 1, self: 1960.920 ms, total: 1960.920 ms.
java/lang/Object.wait.(J)V: count: 1, self: 225.098 ms, total: 225.098 ms.
java/lang/Object.wait.(J)V: count: 3, self: 498.602 ms, total: 498.602 ms.
java/lang/Object.wait.(J)V: count: 1, self: 1114.266 ms, total: 1114.266 ms.
java/lang/Object.wait.(J)V: count: 1, self: 1115.048 ms, total: 1115.048 ms.
java/lang/Object.wait.(J)V: count: 2, self: 863.410 ms, total: 863.410 ms.
java/lang/Object.wait.(J)V: count: 1, self: 135.771 ms, total: 135.771 ms.
[...]

Issue #1541, opened 2015-05-11
Thread.start0 looks expensive


marco-c:
*java/lang/Thread.start0.()V: count: 6, self: 1709.144 ms, total: 1709.144 ms.*

marco-c:
On desktop, *java/lang/Thread.start0.()V: count: 6, self: 1.317 ms, total: 1.317 ms.*

marco-c:
I manually added performance.now() calls in *start0* to verify this value, it's 107 ms.

marco-c:
In https://github.com/mozilla/j2me.js/pull/1572 there's an explanation of why in the Java profiler it looks extremely slow.
It's still somewhat slow with manual performance.now() calls, so I'll leave this open.

PR #1572, opened 2015-05-20
pre-allocate circular buffers to avoid expensive allocations during profiling

mykmelez:
*TimelineBuffer* initializes itself lazily, including creating two circular buffers (one wrapping an *Int32Array* and another wrapping a *Float64Array*). That can skew results when a method being measured by one timeline triggers the creation of another timeline, since occasionally the browser takes a while to create those typed arrays (particularly the *Float64Array* one).

In particular, when profiling a midlet that creates many threads on desktop, *Thread.start0* is one of the most expensive methods in the aggregate method statistics produced by #1569:

> java/lang/Thread.start0.()V: count: 11, self: 94.898 ms, total: 94.898 ms.

But most of the cost is because it creates a new *Context*, which creates a new *TimelineBuffer*, and then starts the context, which enters *org/mozilla/internal/Sys.runThread.(Ljava/lang/Thread;)V* in the new context's timeline, triggering initialization of the timeline (and potentially expensive allocations of the circular buffers). So the cost is actually excise, overhead from profiling, rather than an accurate reflection of the cost of the method.

This branch frontloads that cost during startup by pre-allocating a set of circular buffers for the timeline buffers to use, so it doesn't affect the cost of such methods. With this change, *Thread.start0* is cheap:

> java/lang/Thread.start0.()V: count: 11, self: 2.800 ms, total: 2.800 ms.

### DIFF
--- a/shumway/profiler/timelineBuffer.ts
+++ b/shumway/profiler/timelineBuffer.ts
@@ -76,8 +76,10 @@ module Shumway.Tools.Profiler {
       this._data = [];
       this._kinds = [];
       this._kindNameMap = Object.create(null);
-      this._marks = TimelineBuffer.preAllocInt32ArrayBuffers.pop() || new Shumway.CircularBuffer(Int32Array, BUFFER_SIZE);
-      this._times = TimelineBuffer.preAllocFloat64ArrayBuffers.pop() || new Shumway.CircularBuffer(Float64Array, BUFFER_SIZE);
+      this._marks = TimelineBuffer.preAllocInt32ArrayBuffers.pop() ||
+                    new Shumway.CircularBuffer(Int32Array, TimelineBuffer.BUFFER_SIZE);
+      this._times = TimelineBuffer.preAllocFloat64ArrayBuffers.pop() ||
+                    new Shumway.CircularBuffer(Float64Array, TimelineBuffer.BUFFER_SIZE);
     }
 
     private _getKindId(name: string):number {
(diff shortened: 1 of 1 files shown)


brendandahl:
BTW, if you've had certain timelines overflow, I recommend doing:

-      this._marks = new Shumway.CircularBuffer(Int32Array, 20);
-      this._times = new Shumway.CircularBuffer(Float64Array, 20);
+      this._marks = new Shumway.CircularBuffer(Int32Array, 23);
+      this._times = new Shumway.CircularBuffer(Float32Array, 23);

marco-c:
This explains https://github.com/mozilla/j2me.js/issues/1541

mykmelez:
> BTW, if you've had certain timelines overflow, I recommend doing:

Hmm, I'm unsure if I've seen that. It doesn't look like *CircularBuffer* warns in that case, nor does *TimelineBuffer* check *CircularBuffer.isFull* before calling *CircularBuffer.write*. I suppose it would show up in missing profile data.

In any case, that issue seems orthogonal to the changes here, which just alter the time at which the buffers are allocated. They still start (and stop) being used at the same time, so they should still overflow (or not) as they did before.

mykmelez:
> BTW, if you've had certain timelines overflow, I recommend doing:

Per our discussion in IRC, I parameterized the buffer size in 1d7905d and 8923e58 to make it easier to change it.

PR #1574, opened 2015-05-20
enable PROFILE_FORMAT=csv to output text profile as comma-separated values

mykmelez:


marco-c:
Nit: can you store the file with the ".csv" extension and create the Blob with "text/csv"?

mykmelez:
> Nit: can you store the file with the ".csv" extension and create the Blob with "text/csv"?

Ah, indeed, I've done that in 20db8ba. In the process, I also made the media type for the *PLAIN* format be `text/plain`.

PR #1575, opened 2015-05-21
Support destroyApp for MIDlets that use localmsg

marco-c:


mykmelez:
@TimAbraldes As you've been working in this area with @marco-c, can you take on review for this PR?

midp/localmsg.js
```diff
@@ -18,6 +18,15 @@ var LocalMsgConnection = function() {
     this.clientMessages = [];
 }
 
+LocalMsgConnection.prototype.reset = function() {
```
TimAbraldes: nit: Consider implementing `LocalMsgConnection` (the constructor) in terms of `LocalMsgConnection.prototype.reset`. That way when we modify this class we only need to worry about initialization/reset in one place. I think the body of `LocalMsgConnection` could just be `this.reset()` if the `reset` function is modified slightly.

midp/localmsg.js
```diff
@@ -18,6 +18,15 @@ var LocalMsgConnection = function() {
...
+    this.clientWaiting = [];
+    this.clientMessages = [];
+    while (this.serverWaiting.length > 0) {
+        this.serverWaiting.shift()(false);
+    }
+    this.serverMessages = [];
+}
```
TimAbraldes: This doesn't reset `clientConnected` or `waitingForConnection`. I'm not familiar with this class so maybe that's by design?

midp/localmsg.js
```diff
@@ -78,9 +87,16 @@ LocalMsgConnection.prototype.getServerMessage = function(data) {
...
+
     asyncImpl("I", new Promise((function(resolve, reject) {
-        this.serverWaiting.push(function() {
-            resolve(this.getServerMessage(data));
+        this.serverWaiting.push(function(successful) {
+            if (successful) {
+                resolve(this.getServerMessage(data));
+            } else {
+                ctx.setAsCurrentContext();
+                reject($.newIOException("Client disconnected"));
```
TimAbraldes: I like the new design of sending success/failure across the wire as the first part of the protocol :)

midp/localmsg.js
```diff
@@ -1041,6 +1057,11 @@ Native["org/mozilla/io/LocalMsgConnection.init.(Ljava/lang/String;)V"] = functio
 
     this.connection = typeof MIDP.LocalMsgConnections[this.protocolName] === 'function' ?
         new MIDP.LocalMsgConnections[this.protocolName]() : MIDP.LocalMsgConnections[this.protocolName];
+
+    var ctx = $.ctx;
+    this.connection.reset();
+    ctx.setAsCurrentContext();
```
TimAbraldes: Maybe the context switching should happen as part of the call to `reset` so that callers don't have to remember to do that?

midp/localmsg.js
marco-c: I haven't reset `waitingForConnection` because there could be someone waiting and in `LocalMsgConnection::init` we call `notifyConnection` after reset.
This is never happening at the moment, but I didn't want to have a possible race condition cause.

midp/localmsg.js
marco-c: Done in b38fe60.

TimAbraldes:
LGTM. Consider addressing nits but not necessary. Please rebase into a single patch and I'll merge.

midp/localmsg.js
marco-c: With 1d589fa, I've added resetting `clientConnected`.

PR #1576, opened 2015-05-21
Add MIDlet to test content handlers

marco-c:


marco-c:
We don't support content handlers yet, so this MIDlet doesn't work yet, but it will be useful to implement them.

PR #1578, opened 2015-05-22
Implement content handler natives

marco-c:
This PR is implementing the content handler natives and adding a few simple tests for them.
The only missing part is hooking them up to MozActivity, which I'll do in a followup.

PR #1591, opened 2015-05-25
Support sharing content on Firefox OS

marco-c:
Depends on #1578 and #1589.


marco-c:
Just need to test this again with the latest changes and I'll remove the WIP.

midp/content.js
```diff
@@ -4,10 +4,12 @@
 'use strict';
 
 var Content = (function() {
-  var chRegisteredClassName = null;
-  var chRegisteredStorageID = -1;
-  var chRegisteredID = null;
-  var chRegisteredRegistrationMethod = -1;
+  // Allow configuring these values for MIDlets that don't register
+  // the content handler in time.
```
TimAbraldes: I'm a little leery of implementing dynamic registration in terms of static configuration that will have to be set for each MIDlet suite that uses this mechanism. What happens if a new version of a MIDlet changes one of these values? Will its configuration need to be updated and a new FirefoxOS app version pushed? If so, won't that version not work with the older MIDlet?

It seems like what we need is for the FG MIDlet never to launch until the BG MIDlet has finished launching. Somewhere you mentioned that this caused a startup perf regression. Do you have info on the method you used and how much it hurt startup time?

marco-c: > What happens if a new version of a MIDlet changes one of these values? Will its configuration need to be updated and a new FirefoxOS app version pushed? If so, won't that version not work with the older MIDlet?

Yes, we'll need to update the configuration and the configuration won't work with the older MIDlet. But this is a problem that we already have with many of the configuration options.
Since in this case we have a way to detect the mismatch, in a followup we could ask the user to update the app when needed.

> It seems like what we need is for the FG MIDlet never to launch until the BG MIDlet has finished launching. 

Unfortunately, I think we don't really have a way to tell when the BG MIDlet has finished starting.

> Somewhere you mentioned that this caused a startup perf regression. Do you have info on the method you used and how much it hurt startup time?

We've landed a few improvements over the last months that made the FG MIDlet start earlier, I run the startup benchmark to measure them at the time.

PR #1593, opened 2015-05-26
Always launch FG MIDlet

TimAbraldes:
This is a least-effort method of always launching the FG MIDlet upon launching the BG MIDlet. Upcoming PRs may change the way we launch but this patch should accomplish the goal quickly and simply.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ The tooltip displays:
 
 If you build with `PROFILE=2` or `PROFILE=3`, then the timeline will be saved to a text file instead of being shown in the flame chart. On desktop, you will be prompted to save the file. On the phone, the file will automatically be saved to `/sdcard/downloads/profile.txt`, which you can later pull with `adb pull`. Note that no timeline events under 0.1 ms are written to the file output. You can change this in `main.js` if you'd like.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -7,13 +7,13 @@ The current goals of j2me.js are:
 1. Run MIDlets in a way that emulates the reference implementation of phone ME Feature MR4 (b01)
 1. Keep j2me.js simple and small: Leverage the phoneME JDK/infrastructure and existing Java code as much as we can, and implement as little as possible in JavaScript
 
-## Install, Cold, and Warm runs
+## Install vs Hot runs
 
-J2ME.js launches MIDlets under three different circumstances. We are working to make the most common scenario the fastest, sometimes at the expense of making the least frequent scenarios a little slower.
+J2ME.js launches MIDlets under two different circumstances.
```
marco-c: We'll need to add a description for the *sharing content* and *notification* cases, but we can do that in a followup (once we actually implement them :smile:).
We'll also need to build a new benchmarking mode (probably very similar to the old warm benchmark) and a profiling mode for these new kinds of startup (which is also useful to accurately separate the startup time between BG startup time and FG startup time).

PR #1520, opened 2015-05-08
ASM.jsify VM, WIP

mbebenita:
This is an attempt to lower the abstraction level on top of which the VM is built.

1. The first part of this PR changes the way we represent the call stack. Instead of representing the Java stack as array of frames (arrays) into which we push/pop values of different types, we now use one large typed array and do our own stack frame handling within it.
2. The VM aliases a typed array heap from a `asm.js` module and uses `malloc`/`free` to allocate stack space. This is convenient because it allows us to call into `C/C++` code whenever we need to do some heavy lifting.

PR #1598, opened 2015-05-27
asm.jsify VM

mykmelez:
@mbebenita, @brendandahl, @marco-c, @TimAbraldes, @wfwalker: This is an updated version of #1520, with a recent merge from master. It uses an *intex* branch in the central repository to make it easier for us to collaborate on the branch.

Branch rules: push at will, review not needed. But do ask for review anyway (i.e. request a pull into this branch) on changes you're uncertain about.


mykmelez:
@mbebenita, @brendandahl, @marco-c: What do we still need to do to merge this branch to master?

In my startup benchmark runs, this branch starts faster now that the compiler and compiled method cache have been enabled. It still uses a bit more memory on startup, but that's an apples-to-oranges comparison, given that we pre-allocate the asm.js heap.

Runtime perf seems equivalent or better, although I don't have a good runtime metric, just a subjective sense of the performance of RunWhatsApp.

Runtime memory usage may be a bit worse, however. I did some measurements at three different times (after starting RunWhatsApp to the conversation list, after tapping a conversation to view it, and after scrolling a long contact list), and this branch seems to be using more memory on my Aries, although that device has a user build, so it's providing limited stats (only VSIZE and RSS, which @EricRahm has said is inaccurate). I'll test again on a Flame with an eng build.

Is there any additional cleanup we need to do? Things we disabled that need to be enabled (and possibly fixed) again? XXX or REDUX comments to revisit? Other changes that should land before rather than after the merge?

My primary metric here is: runs RunWhatsApp with equivalent or better performance and memory usage. Once we're there, we should merge. But let me know if there's something else to unbreak as well.

marco-c:
I've gone through all the issues marked with the "intex" label (which were the issues we initially thought should block merging intex). I've closed some of them and moved others to "intex-followup", but there are still a few left in the "intex" bucket.

The only ones that concern me are the ones related to GC-JIT. Even if they're only easily reproducible if you force a collection for every allocation, it doesn't mean they couldn't hit at any time.

> Is there any additional cleanup we need to do? Things we disabled that need to be enabled (and possibly fixed) again?

I wouldn't block on the cleanups or things to re-enable (for example AOT compilation) as long as intex is faster than master, but I'd file an issue for anything that we need to do (otherwise we'll definitely forget something).

mykmelez:
@marco-c @mbebenita @brendandahl I wonder if we should merge this branch to master. If we restart working on the project at some point, or someone else picks it up, then the implementation on this branch will be the one to use, since it has seen the lion's share of recent development.

As I recall, the only remaining issue is a perf regression on startup. That mattered a lot when we were trying to ship software that had startup perf issues, but it doesn't matter as much now that we aren't trying to ship that software. It doesn't seem like it should hold back the branch at this point.

marco-c:
> As I recall, the only remaining issue is a perf regression on startup. That mattered a lot when we were trying to ship software that had startup perf issues, but it doesn't matter as much now that we aren't trying to ship that software. It doesn't seem like it should hold back the branch at this point.

I think there were still a few issues with the integration of the JIT with the GC (see the issues marked with the `intex` label).
They are tracked though, so I'd tag the current master and merge this.

brendandahl:
I agree with merging, would hate to have someone start on the old approach.

mykmelez:
Ok, thanks @marco-c and @brendandahl, I've tagged the current tip of master as before-intex and will merge this forthwith!

PR #1599, opened 2015-05-27
Benchmark FG MIDlet restart

marco-c:
This benchmark is also useful to separate the FG MIDlet startup time and the rest of the startup time (for example it could be useful when testing something that supposedly improves FG MIDlet startup time).

Needs #1591.


mykmelez:
I've reviewed this branch a couple times, before and after #1591, and it looks good. My only nit is that it isn't clear that "Benchmark warm startup" is the checkbox that enables the "fgStartupTime" measurement, since the two have such different labels. I would make these consistent. But we can do that in a followup.

PR #1601, opened 2015-05-27
WIP: install/configure emscripten to compile libs/native.js

mykmelez:
@marco-c This gets us closer to compiling libs/native.js on Travis, but my build still fails with:

> Running CMake: ['cmake', '-G', 'Unix Makefiles', '-DCMAKE_BUILD_TYPE=Release', '-DLLVM_TARGETS_TO_BUILD=X86;JSBackend', '-DLLVM_INCLUDE_EXAMPLES=OFF', '-DLLVM_INCLUDE_TESTS=OFF', '-DCLANG_INCLUDE_EXAMPLES=OFF', '-DCLANG_INCLUDE_TESTS=OFF', '/home/travis/build/mykmelez/j2me.js/build_tools/emsdk_portable/clang/fastcomp/src']
> Re-run cmake no build system arguments
> CMake Error at CMakeLists.txt:3 (cmake_minimum_required):
> CMake 2.8.8 or higher is required.  You are running version 2.8.7

- https://travis-ci.org/mykmelez/j2me.js/builds/64341927


marco-c:
This is a working configuration: https://github.com/urho3d/Urho3D/blob/3b099a149bd51efde13f1d1b38538dc11c745c7c/.travis.yml

Maybe one of the options they're using is necessary, for example it could be the `addons: {apt: {sources: [george-edison55-precise-backports, kubuntu-backports, ubuntu-toolchain-r-test], packages: [cmake, doxygen, graphviz, g++-4.9, libasound2-dev]}}` line.

Makefile
```diff
@@ -237,6 +237,13 @@ build_tools/closure.jar: build_tools/.closure_compiler_version
 	wget -P build_tools https://github.com/mykmelez/closure-compiler/releases/download/$(CLOSURE_COMPILER_VERSION)/closure.jar
 	touch build_tools/closure.jar
 
+build_tools/emsdk-portable.tar.gz:
+	wget -P build_tools https://s3.amazonaws.com/mozilla-games/emscripten/releases/emsdk-portable.tar.gz
```
marco-c: We should probably host this ourselves, so we can control the version.

marco-c:
They've also hosted this on https://github.com/urho3d/emscripten-sdk: "Installed Emscripten SDK for speedy cloning into Travis CI worker VM"

Makefile
mykmelez: > We should probably host this ourselves, so we can control the version.

The Portable SDK supports installing older versions of the SDK and toolchains, so we can control their version (although there may be something in the Portable SDK that could still break in an update).

marco-c: We'd also need a way to update the SDK when we want to use a different version, like we do with the other build tools.

marco-c:
> Maybe one of the options they're using is necessary, for example it could be the addons: {apt: {sources: [george-edison55-precise-backports, kubuntu-backports, ubuntu-toolchain-r-test], packages: [cmake, doxygen, graphviz, g++-4.9, libasound2-dev]}} line.

To fix the cmake problem what we need is probably:
`addons: { apt: { sources: [kubuntu-backports], packages: [cmake]}}`

https://launchpad.net/~kubuntu-ppa/+archive/ubuntu/backports includes cmake 2.8.12.

https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test contains some updates for gcc, binutils, etc., but I wouldn't add it unless necessary.

mykmelez:
> We'd also need a way to update the SDK when we want to use a different version, like we do with the other build tools.
> And probably also a way to update the SDK installer.

https://github.com/mykmelez/j2me.js/commit/b8b20effbdba9f8d0ec27c7ad054cc9d69837706 adds version tracking for both the SDK itself and its package (using a fake version for the package so we can force an update if we need/want to).

> To fix the cmake problem what we need is probably:
> addons: { apt: { sources: [kubuntu-backports], packages: [cmake] } }

That fixed that problem, but now I'm seeing:

> make: Circular build_tools/emsdk_portable/clang/fastcomp/src/tools/clang/test/ARCMT/with <- build_tools/emsdk_portable/clang/fastcomp/src/tools/clang/test/ARCMT/with dependency dropped.
> make: *** No rule to make target `space/test1.m.in', needed by `build_tools/emsdk_portable/clang/fastcomp/src/tools/clang/test/ARCMT/with'.  Stop.

marco-c:
Weird, the file should be `emsdk_portable/clang/fastcomp/src/tools/clang/test/ARCMT/with space/test1.m.in`. There's probably some error caused by the whitespace character.

marco-c:
> They've also hosted this on https://github.com/urho3d/emscripten-sdk: "Installed Emscripten SDK for speedy cloning into Travis CI worker VM"

Ah, I think they're installing the emscripten-sdk in a repository, so they don't have to do:
```
./emsdk update && ./emsdk install $(EMSDK_VERSION) && ./emsdk activate $(EMSDK_VERSION)
```
but only:
```
./emsdk activate $(EMSDK_VERSION)
```

marco-c:
I'm testing that in https://github.com/mozilla/j2me.js/pull/1606, now it's failing when sourcing emsdk_env.sh

marco-c:
In #1606, I've also tried to install it in the .travis.yml before_script, to avoid problems with sourcing in a makefile. It doesn't work anyway (emcc not found).

marco-c:
Looks like the PATH does include the correct directories: https://travis-ci.org/mozilla/j2me.js/builds/64496785

marco-c:
I've found the issues in my branch, this last build is able to compile with Emscripten and run tests: https://travis-ci.org/mozilla/j2me.js/builds/64603538 (tests are failing because of a timeout, which is expected)

marco-c:
So, my approach in #1606 would only work for Travis, so it would require people to install Emscripten manually (like we're currently doing with TypeScript and a few other dependencies).

Installing Emscripten automatically like this branch does would make it easier for developers, but it would also require more work to fix the building issues.

If we want to go with #1606, I can clean it up and submit a new PR.

PR #1603, opened 2015-05-28
Implement `AppProxy` natives

TimAbraldes:
Implement `AppProxy.midletIsAdded`, `AppProxy.midletIsRemoved`,
`AppProxy.isMidletRunning`, and `AppProxy.isSuiteRunning`. I implemented
these as part of the FG MIDlet work. They don't seem to be used that
much by popular MIDlets, but since I have the implementation sitting
around I figured I'd submit the PR.

### DIFF
--- a/midp/content.js
+++ b/midp/content.js
@@ -69,7 +69,6 @@ var Content = (function() {
     return 0;
   };
 
-  addUnimplementedNative("com/sun/j2me/content/AppProxy.midletIsRemoved.(ILjava/lang/String;)V");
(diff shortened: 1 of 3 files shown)


marco-c:
I'd implement new natives only when needed, these seem unneeded in our architecture (two of them are never called) and could probably be removed altogether.

TimAbraldes:
> I'd implement new natives only when needed, these seem unneeded in our architecture (two of them are never called) and could probably be removed altogether.

I figured that'd be the response. I just wanted to submit the PR so the code didn't get lost forever. If we need these in the future we have them available.

PR #1604, opened 2015-05-28
Improve NATIVE_MIDLET_* handling

TimAbraldes:
Make `MIDletProxyList` responsible for handling `NATIVE_MIDLET_*` events. Create a consumer and a listener class similar to the existing `MIDletControllerEventConsumer` and `MIDletControllerEventListener` but for native MIDlet controller events. Add arguments to the `NATIVE_MIDLET_EXECUTE_REQUEST` and `NATIVE_MIDLET_DESTROY_REQUEST` events. Make `MIDP.sendDestroyMIDletEvent` use `NATIVE_MIDLET_DESTROY_REQUEST` instead of `DESTROY_MIDLET_EVENT`.

These changes came out of the work I was doing with FG MIDlet launching.

This builds on #1600.

@marco-c This conflicts with some of the work you've done in #1591 since we both extended the "execute" and "destroy" handling to include more args. This change modifies the way we handle sending execute and destroy events. Perhaps you wouldn't mind integrating the work here into your work for #1591?

### DIFF
--- a/java/custom/com/nokia/mid/s40/bg/BGUtils.java
+++ b/java/custom/com/nokia/mid/s40/bg/BGUtils.java
@@ -1,55 +1,9 @@
 package com.nokia.mid.s40.bg;
 
-import com.sun.cldc.isolate.Isolate;
-import com.sun.midp.main.AmsUtil;
-import com.sun.midp.midletsuite.MIDletSuiteStorage;
-
-import com.sun.midp.events.Event;
-import com.sun.midp.events.NativeEvent;
-import com.sun.midp.events.EventQueue;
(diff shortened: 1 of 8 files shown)


marco-c:
> @marco-c This conflicts with some of the work you've done in #1591 since we both extended the "execute" and "destroy" handling to include more args. This change modifies the way we handle sending execute and destroy events. Perhaps you wouldn't mind integrating the work here into your work for #1591?

@TimAbraldes what's the purpose of these changes? If it's only a refactoring, then I think we should first land #1591 that is adding some new tests and then look into this.

mykmelez:

PR #1607, opened 2015-05-29
Remove use of long.js library.

### LINKED ISSUE #1581: Completely remove the long library.
The only use of it right now is to convert from doubles to longs:

```typescript
  export function numberToLong(v: number): number {
    // TODO Extract logic from Long so we don't allocate here.
    var long = Long.fromNumber(v);
    tempReturn0 = long.high_;
    return long.low_;
  }
```

brendandahl:
Closes #1581


vm/runtime.ts
```diff
@@ -1913,8 +1897,8 @@ module J2ME {
     }
   }
 
-  export function checkDivideByZeroLong(value: Long) {
-    if (value.isZero()) {
+  export function checkDivideByZeroLong(low: number, high: number) {
```
mykmelez: BaselineCompiler.emitDivideByZeroCheck will need updating to reflect this change. I guess it doesn't need to happen in conjunction with this change, since the compiler doesn't work at the moment. :-)

mykmelez:
Reverted due to a perf regression that causes build failures like this one on [https://travis-ci.org/mozilla/j2me.js/builds/65105990](build 65105990):

>1.2:1.5  | Running javax.microedition.lcdui.TestGraphicsClipping 
>FAIL unresponsive Script: http://localhost:8000/bld/j2me.js:5971

PR #1608, opened 2015-05-29
Install Emscripten on Travis

marco-c:


vm/native/Makefile
```diff
@@ -26,3 +25,6 @@ all: native.cpp
 	echo "  return Module;" >> native.js
 	echo "})(ASM);" >> native.js
 	mv native.js ../../libs/native.js
+
+Boehm.js/.libs/$(BOEHM_LIB):
+	cd Boehm.js && emconfigure ./configure --without-threads --disable-threads __EMSCRIPTEN__=1 && emmake make && cd ..
```
mykmelez: Nit: This doesn't need to `cd ..`, as Make resets the cwd after each line of the recipe (in this case, there also aren't any subsequent lines of the recipe).

marco-c:
The build still fails because we need a longer timeout, but compiling with Emscripten works.

PR #1610, opened 2015-06-01
WIP Implement an AMS MIDlet

TimAbraldes:
Up to now, we've been telling the `MIDletSuiteLoader` class to launch
whatever user `MIDlet` we're trying to launch. We've been targeting the
background MIDlet of each suite we're running and then expecting the BG
MIDlet to call `BGUtils.setBGMIDletResident`. At that point, we've been
launching our best guess at what the FG MIDlet is (we currently assume
that there are only ever a max of 2 MIDlets in a suite).

This has a few important implications.

Because we launch the FG MIDlet when the BG MIDlet calls
`BGUtils.setBGMIDletResident`, and because that is frequently during the
BG MIDlet's `startApp` or constructor, the BG MIDlet does not get a
chance to fully initialize before the FG MIDlet starts up and takes the
foreground. This means that certain registrations that should be present
in the system by the time the FG MIDlet launches will not be present if
they were supposed to have been made by the BG MIDlet. It also means
that we cannot pause the BG MIDlet because the system will believe that
it is in a pre-startup state and will terminate it instead.

The MIDlet launched by `MIDletSuiteLoader` is supposed to be a special
"AMS" MIDlet that is responsible for managing user MIDlets
(launching/pausing/destroying, etc). The Isolate of the MIDlet launched
by `MIDletSuiteLoader` gets special privileges and is treated
specially by the MIDP system.

This patch implements an AMS MIDlet to be loaded into the special AMS
isolate. As a result of this patch we no longer load user MIDlets into
the AMS isolate. The AMS MIDlet implemented by this patch launches the
BG MIDlet if there is one specified in the JAD we are loading, then it
launches the FG MIDlet specified in the j2me.js configuration, or it
guesses like we've currently been doing if the configuration targets the
BG MIDlet.

The AMS MIDlet is now the handler for `NATIVE_MIDLET_EXECUTE_REQUEST`
events which we send to request MIDlet launches, and for
`NATIVE_MIDLET_DESTROY_REQUEST` which we now send to request MIDlet
destruction. These are implemented in terms of the `AmsUtil` and
`MIDletProxyList` classes which make certain methods available
only to the AMS MIDlet.

I ran the startup benchmarks on device and found that this patch makes startup slower by an amount that is probably not acceptable. WIP while I see if I can bring the startup hit down. Also the "bgStartupTime" and "fgStartupTime" values don't work because the functions that they rely on have been eliminated. I'll fix that in another push to this branch.

|         Test         | Baseline Mean |    Mean    |    +/-    |    %    |     P    |     Min    |     Max     | 
|:---------------------|--------------:|-----------:|----------:|--------:|---------:|-----------:|------------:|
| startupTime          |      16,618ms |   19,494ms |   2,876ms |   17.31 |    WORSE |   17,221ms |    22,608ms | 
| vmStartupTime        |       2,662ms |    2,702ms |      39ms |    1.47 |    WORSE |    2,397ms |     2,987ms | 
| bgStartupTime        |         906ms |      NaNms |     NaNms |     NaN |     SAME | Infinityms | -Infinityms | 
| fgStartupTime        |      11,476ms |      NaNms |     NaNms |     NaN |     SAME |      NaNms |       NaNms | 
| fgRefreshStartupTime |       1,573ms |    1,126ms |    -447ms |  -28.43 |   BETTER |      782ms |     1,909ms | 
| *totalSize*          |    *26,506kb* | *26,731kb* |   *224kb* |  *0.85* |  *WORSE* | *26,017kb* |  *27,282kb* | 
| *domSize*            |        *74kb* |     *74kb* |     *0kb* | *-0.02* |   *SAME* |     *74kb* |      *74kb* | 
| *styleSize*          |       *311kb* |    *311kb* |     *0kb* |  *0.00* |   *SAME* |    *304kb* |     *316kb* | 
| *jsObjectsSize*      |    *18,122kb* | *18,412kb* |   *290kb* |  *1.60* |  *WORSE* | *18,389kb* |  *18,435kb* | 
| *jsStringsSize*      |     *1,016kb* |    *928kb* |   *-88kb* | *-8.69* | *BETTER* |    *927kb* |     *928kb* | 
| *jsOtherSize*        |     *6,719kb* |  *6,731kb* |    *12kb* |  *0.18* |   *SAME* |  *6,093kb* |   *7,209kb* | 
| *otherSize*          |       *264kb* |    *275kb* |    *10kb* |  *3.88* |   *SAME* |    *213kb* |     *343kb* | 
| *USS*                |    *58,046kb* | *59,174kb* | *1,128kb* |  *1.94* |  *WORSE* | *57,416kb* |  *60,848kb* | 
| *peakRSS*            |         *0kb* |    *NaNkb* |   *NaNkb* |   *NaN* |   *SAME* |    *NaNkb* |     *NaNkb* |

### DIFF
--- a/java/custom/com/nokia/mid/s40/bg/BGUtils.java
+++ b/java/custom/com/nokia/mid/s40/bg/BGUtils.java
@@ -1,55 +1,6 @@
 package com.nokia.mid.s40.bg;
 
-import com.sun.cldc.isolate.Isolate;
-import com.sun.midp.main.AmsUtil;
-import com.sun.midp.midletsuite.MIDletSuiteStorage;
-
-import com.sun.midp.events.Event;
-import com.sun.midp.events.NativeEvent;
-import com.sun.midp.events.EventQueue;
-import com.sun.midp.events.EventTypes;
-import com.sun.midp.events.EventListener;
-
-public class BGUtils implements EventListener {
-    private static native int getFGMIDletNumber();
-    private static native String getFGMIDletClass();
-    public static native void maybeWaitUserInteraction(String midletClassName);
-
-    public static void setBGMIDletResident(boolean param) {
-        AmsUtil.executeWithArgs(MIDletSuiteStorage.getMIDletSuiteStorage(), 0, BGUtils.getFGMIDletNumber(),
-                                BGUtils.getFGMIDletClass(), null, null, null, null, -1, -1, Isolate.MAX_PRIORITY,
-                                null, false);
-    }
-
+public class BGUtils {
+    public static native void setBGMIDletResident(boolean param);
     public static native boolean launchIEMIDlet(String midletSuiteVendor, String midletName, int midletNumber, String startupNoteText, String args);
-
-    static {
-        EventQueue.getEventQueue()
-                  .registerEventListener(EventTypes.NATIVE_MIDLET_EXECUTE_REQUEST,
-                                         BGUtils.getBGUtilsInstance());
-    }
-    private static BGUtils bgUtils = null;
-    private static BGUtils getBGUtilsInstance() {
-        if (bgUtils == null) {
-            bgUtils = new BGUtils();
-        }
-
-        return bgUtils;
-    }
-
-    public boolean preprocess(Event event, Event waitingEvent) {
-        return true;
-    }
-
-    public void process(Event event) {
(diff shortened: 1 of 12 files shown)


mykmelez:

PR #1622, opened 2015-06-05
Set pointers to object interiors as invalid

marco-c:
I think this should work in our case, but I want to test it more after we'll actually have enabled the GC.


mykmelez:
@marco-c Will this become relevant once you enable GC for some objects in #1731?

marco-c:
> @marco-c Will this become relevant once you enable GC for some objects in #1731?

Yes, I want to write a simple GC benchmark and test if this option improves collection time (I hope so).

marco-c:
This shouldn't actually affect performance, but should make less objects wrongly marked as still live.

PR #1630, opened 2015-06-05
Fix building on Windows

marco-c:
There are still a couple of problems that we need to tackle after these fixes:
 - The build appears to fail at `js bld/jsc.js -cp java/classes.jar -d -jf java/classes.jar -mff aot-methods.txt > bld/classes.jar.js`, even if the file is actually created. So you need to run *make* again to complete the build.
 - tests/runtests.py doesn't work correctly, all the servers fail to start (`[Errno 112] Address already in use`) and casperjs fails to start (`[Errno 2] No such file or directory; did you install slimerjs?`)


marco-c:
ecc22ee fixes the first problem.

mykmelez:
Thanks for removing one more recursive Makefile!

PR #1634, opened 2015-06-09
Add a native to force a garbage collection

marco-c:
This is useful to test the GC behavior both while integrating the GC and when we'll need to write tests for it (there are some tests on https://github.com/marco-c/Boehm.js that are running automatically, but they aren't specific to our VM).


mykmelez:
Looks good, needs conflict resolution.

marco-c:
Done in e4af25c.

PR #1636, opened 2015-06-09
Allow running a subset of the basic unit tests without launching a MIDlet

marco-c:
Sometimes it's useful to run tests without launching a MIDlet.
Launching a MIDlet is a quite complex process, so if you're debugging a low-level feature the chances that you can't even start the current test harness are high.

### DIFF
--- a/tests/Harness.java
+++ b/tests/Harness.java
@@ -1,0 +1,77 @@
+/* vim: set filetype=java shiftwidth=4 tabstop=4 autoindent cindent expandtab : */
+
+import gnu.testlet.*;
+
+import com.sun.cldchi.jvm.JVM;
+import javax.microedition.lcdui.Display;
+import javax.microedition.lcdui.Form;
+import javax.microedition.midlet.*;
+import java.lang.Exception;
+import java.util.Vector;
+
+public class Harness extends TestHarness {
+    private String testName;
+    private int testNumber = 0;
+    private String testNote = null;
+    private int pass = 0;
+    private int fail = 0;
+    private int knownFail = 0;
+    private int unknownPass = 0;
+
+    public Harness(String note, Display d) {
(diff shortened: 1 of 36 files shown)


tests/Harness.java
```diff
@@ -0,0 +1,77 @@
...
+import gnu.testlet.*;
+
+import com.sun.cldchi.jvm.JVM;
+import javax.microedition.lcdui.Display;
+import javax.microedition.lcdui.Form;
+import javax.microedition.midlet.*;
+import java.lang.Exception;
+import java.util.Vector;
+
+public class Harness extends TestHarness {
```
mykmelez: Nit: it's strange that the name of this class, *Harness*, is more generic than the name of the superclass it extends, *TestHarness*. I would give it a more specific name than that class, like *RunTestsHarness* or *j2mejsTestHarness*.

tests/gnu/testlet/MIDletTestlet.java
```diff
@@ -0,0 +1,4 @@
+package gnu.testlet;
+
+public interface MIDletTestlet extends Testlet {
```
mykmelez: Is there something you plan to put in *MIDletTestlet*? It isn't clear why this is necessary when it doesn't actually change or add to the *Testlet* class.

marco-c: I added this interface to be able to simply grep to find classes to add to the MIDletTestlets list.

mykmelez: Ok, that's fine. ;-)

tests/Harness.java
marco-c: Renamed Harness to j2mejsTestHarness in a884190.

PR #1638, opened 2015-06-10
Set DEBUG_PRINT_ALL_EXCEPTIONS to true in non-release builds

marco-c:


brendandahl:
This makes it hard to filter out what are really exceptions.

marco-c:
> This makes it hard to filter out what are really exceptions.

Yes, this is true, but it also gives more info when you're testing a MIDlet and hit a bug, which could be important in two cases:
 - A bug that you can't reproduce, so you can't simply run the MIDlet again modifying the property
 - A bug that a non-developer hits, who doesn't know how to build j2me.js

mykmelez:
Can we parameterize this with a URL parameter instead? I.e. something like *printCaughtExceptions*. Then we can enable it without modifying source and independently from *release*.

mykmelez:
I'm marking this request as incomplete and closing it, as the work here appears to have stalled. Please reopen the request if you resume work on it!

marco-c:
Good idea, I've decided not to do this because it isn't as useful as I initially thought (people are not going to see these messages anyway).

PR #1639, opened 2015-06-10
Benchmark with a I/O-heavy background MIDlet and a foreground MIDlet …

marco-c:
…that needs to be responsive

### DIFF
--- a/bench/IOStressBench.java
+++ b/bench/IOStressBench.java
@@ -1,0 +1,75 @@
+package benchmark;
+
+import javax.microedition.midlet.MIDlet;
+import javax.microedition.lcdui.Canvas;
+import javax.microedition.lcdui.Display;
+import javax.microedition.lcdui.Graphics;
+
+public class IOStressBench extends MIDlet {
+    static class BouncyColors extends Canvas {
+        int x = 0;
+        int y = 0;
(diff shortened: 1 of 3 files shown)


mykmelez:
Nit: can you rename *StressBench* to *SocketStressBench* to distinguish the two stress benchmarks more clearly? It might also be worth calling this something like *FileStressBench*, since both benchmarks are doing IO.

It'd also be interesting to measure IO throughput (for both benchmarks) during these benchmarks, to see how prioritizing the foreground thread affects it. But that's a followup.

bench/IOStressBench.java
```diff
@@ -0,0 +1,75 @@
...
+        public void run() {
+            while (true) {
+                FileConnectionBench bench = new FileConnectionBench();
+                bench.runBenchmark();
+            }
+        }
+    }
+
+    public static void main(String args[]) {
+      System.out.println("Run the StressBench benchmark as a midlet: midletClassName=benchmark.IOStressBench");
```
mykmelez: Nit: *StressBench* -> *IOStressBench* (or, if you rename it, that name)

marco-c:
Renamed the benchmarks in faabf85 and fd0a33e.

PR #1642, opened 2015-06-10
WIP: scale canvas by device pixel ratio

mykmelez:
This is based on http://www.html5rocks.com/en/tutorials/canvas/hidpi/ and https://github.com/luser/asteroids/blob/master/asteroids.js#L42 and seems to improve clarity, although it also makes text and graphics too small. Presumably that's because I commented out the context.scale call; but with it, elements disappear from the screen (presumably off the sides). I also couldn't get events to dispatch. So there's something else missing here.

cc: @TimAbraldes , @marco-c

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,8 +55,8 @@ JSR_179 ?= 1
 export JSR_179
 
 # Content handler support
-CONTENT_HANDLER_SUPPORT ?= 1
-export CONTENT_HANDLER_SUPPORT
+CONTENT_HANDLER_FILTER ?= '["image/*", "video/*", "audio/*"]'
+export CONTENT_HANDLER_FILTER
 
 ifeq ($(PROFILE),0)
   J2ME_JS_OPTIMIZATION_LEVEL = J2ME_OPTIMIZATIONS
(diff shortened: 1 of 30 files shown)


mykmelez:
Incidentally, see https://bugzilla.mozilla.org/show_bug.cgi?id=780362 for the current thinking on a HiDPI backing store for canvas. It sounds like that won't happen anytime soon, so we shouldn't expect to query for the backing store ratio.

mykmelez:
I'm afraid this is quite daunting. I can scale the canvas by the display pixel ratio, which renders text and images sharply but makes them twice as small.

I can then increase the font size to compensate for the small text, but some graphic elements that contain text, like buttons and list items, are sized without regard for the font size, so their text becomes too big for them.

And I don't know of a way to increase the image sizes. The standard advice, as reported in the aforementioned blogs, assumes you control your images, whereas in our case we rely on the midlet to provide them.

Plus, there are midlets that draw graphical elements at a fixed position/size regardless of the screen size. Our own TestCanvas does this, as does a popular midlet.

There may be something I'm missing, but at the moment I don't see how we can maintain the proportions of a midlet while eliminating graphical and textual fuzziness.

mykmelez:
I'm still working on this, but on a new branch. So I'm closing this request and will open a new one if I come up with a fix for the issue.

PR #1644, opened 2015-06-12
Remove J2ME security checks

marco-c:
We don't need them because we rely on the web security model.


mykmelez:
Nice. Does this show up as improvements in any benchmarks?

marco-c:
I haven't run any measurements, but I saw these functions called a lot of times in a Java profile (indeed some of them were in aot-methods.txt)

PR #1647, opened 2015-06-15
Inline *read* calls in DataInputStream and *write* calls in DataOutputStream

marco-c:
I've noticed *readInt* and *readShort* are called many times during startup of a MIDlet. This micro-optimization improves their perf by nearly 2x.


mykmelez:
It's strange that these simple passthrough methods are so expensive, especially since they are presumably compiled (some AOT, others JIT). But perhaps it's only because the underlying file write operation is so cheap? In any case, we should look at the interpreter's implementation of the INVOKE* bytecodes to see if there are perf improvements we can make.

(It might also be worth reviewing the list of these methods that are AOT-compiled, to see if any no longer need to be, or new ones should be added.)

PR #1653, opened 2015-06-15
Don't stop running tests when there's a null testlets entry

marco-c:


mykmelez:
Looks good, ready to merge once #1652 is merged (so this doesn't break the build).

PR #1654, opened 2015-06-16
Merge from master into intex

marco-c:


mykmelez:
@marco-c This has that issue I'm seeing locally as well: TestBouncyCastleSHA256 takes forever, and eventually the test runner times out. Perhaps we can disable it on the intex branch while we wait for the compiler to land?

PR #1655, opened 2015-06-17
update README for new name of project

mykmelez:
This updates the README to reflect the new name of the project (and adds a reference to the trademark for JavaScript).


marco-c:
We should update this image as well:
https://cloud.githubusercontent.com/assets/812428/7572315/0a3a174e-f7d2-11e4-811d-5e1a38caa439.png

mykmelez:
> We should update this image as well:
> https://cloud.githubusercontent.com/assets/812428/7572315/0a3a174e-f7d2-11e4-811d-5e1a38caa439.png

Yes. I think @wfwalker created that image. @wfwalker: do you have the source for it? Otherwise, I may be able to simply trim off the part that references the old name.

mykmelez:
> Yes. I think @wfwalker created that image. @wfwalker: do you have the source for it? Otherwise, I may be able to simply trim off the part that references the old name.

Here's a version of the image with the name trimmed off. I'll update the reference in the README to point to this image.

install-cold-warm

PR #1659, opened 2015-06-18
add header to object layout and reference objects by address

mykmelez:
@marco-c This is still rough but more-or-less ready for review. There's still lots to do, but I think it can be done in followups.

This is basically an intermediate state between "every Java object has a corresponding JS object, and references are those JS objects" and "every Java object is allocated only in the ASM heap, and references are the heap addresses of those Java objects."

The branch doesn't create a persistent JS object for each Java object (except for Class objects), but it does frequently create JS handles, including a handle for the *this* object when it calls into a native and a handle for each object argument to the native.  It also automatically dereferences object return values from natives.

The goal was to avoid touching natives more than absolutely necessary.  The branch still has to touch natives that set properties on the JS representations of Java objects and expect them to persist across native method invocations, since the handles aren't persistent.  It does that by creating a map from Java object addresses to JS objects (which we'll need to figure out how to collect when we collect the Java objects).

Also, a handle property that references another object will contain its address rather than a handle to it.  So natives have to get those handles themselves. But other than that, they can mostly behave as before. Which minimizes the scope of this branch significantly.

Note that the handles are still created by the same Klass constructors that the current implementation on master (and the intex branch) uses to construct the JS objects we currently pass as references. And those are fairly heavyweight constructors (with prototype chains that mirror the Java class hierarchy).  We should do something more lightweight.

There's also a lot of code to deal with references that may or may not be addresses, since primitive and object arrays are still represented by JS objects, as are null values (which are represented by the JS null object).  We should be able to get rid of that code once we stop passing around non-addresses.

And there's a separate monitor map that stores monitor objects for Java objects with locks. We may be able to move lock state into the object headers. That'd need further research (looking at how other VMs do it).

Here's a (probably incomplete) list of followups:

* Stop wrapping the *this* object in a handle when calling natives.  This'll require us to update all natives that access the *this* object to wrap it themselves.
* Stop unwrapping the return value after calling natives.  This'll require us to update all natives that return a value to return an unwrapped value themselves.
* Remove the Thread.pid binding, which appears to not be used anymore.
* Replace heavyweight Klass constructor handles with a lightweight approach.
* Consider aligning memory to certain byte boundaries.
* Replace getNative(getHandle()) calls with direct calls to NativeMap.get().
* Replace J2ME.fromJavaString(getHandle(this.nativePath)) with handle-less conversion of strings.
* Consolidate the objects we store in nativeConnectionMap with those in NativeMap.
* Figure out if we can store lock state in the object header.


mykmelez:
(Also note that perf is slower on this branch, perhaps because of the extra code to handle non-address references. We should see what it's like once we start storing arrays in the ASM heap.)

int.ts
```diff
@@ -1704,7 +1782,16 @@ module J2ME {
                       args.unshift(i32[--sp]);
                       break;
                     case Kind.Reference:
-                      args.unshift(ref[--sp]);
+                      // XXX Update natives to expect addresses and stop passing
+                      // handles.
+                      address = ref[--sp];
+                      if (typeof address === "number") {
+                        args.unshift(getHandle(address));
+                      } else if (typeof address === "object") {
```
marco-c: Nit: this branch could be removed in favor of the next more general one.

midp/localmsg.js
```diff
@@ -1034,16 +1034,26 @@ MIDP.LocalMsgConnections["nokia.product-info"] = NokiaProductInfoLocalMsgConnect
 
 var localmsgServerWait = null;
 
+// XXX Consolidate the objects we store in this map with those in NativeMap.
+nativeConnectionMap = Object.create(null);
```
marco-c: Is there a reason why you needed a different map or is it just a temporary thing?

marco-c:
After applying this patch, I can't run WhatsApp. I see two instances of the following message:
```
1.2:34.5 | java.lang.ArrayIndexOutOfBoundsException 
1.2:34.5 |    (stack trace incomplete) 
```

midp/localmsg.js
mykmelez: > Is there a reason why you needed a different map or is it just a temporary thing?

It's just a temporary thing. I happened to create *nativeConnectionMap* for *connection* before I started to use *NativeMap* for the other two properties, which I bundled together into an *info* object. And it wasn't immediately clear how to integrate *connection* into the *info* object. So I left it alone.

Ideally, *server* and *protocolName* would both be stored in Java fields, since they're primitive values. So the only native will be *connection*. But I left that for a followup.

mykmelez:
> After applying this patch, I can't run WhatsApp.

5e2143e fixes the bug.

vm/runtime.ts
```diff
@@ -2097,8 +2224,21 @@ var AOTMD = J2ME.aotMetaData;
  * Are we currently unwinding the stack because of a Yield? This technically
  * belonges to a context but we store it in the global object because it is
  * read very often.
+ *
+ * We define it as a getter/setter to enable breaking when it is set in Chrome:
+ *   http://stackoverflow.com/questions/11618278/how-to-break-on-property-change-in-chrome
  */
-var U: J2ME.VMState = J2ME.VMState.Running;
+var _U: J2ME.VMState = J2ME.VMState.Running;
```
brendandahl: This checked extremely often, can we remove this or enable it only for debugging?

vm/runtime.ts
```diff
@@ -2146,3 +2290,8 @@ var TN = J2ME.throwNegativeArraySizeException;
 
 var PE = J2ME.preempt;
 var PS = 0; // Preemption samples.
+
+var OBJ_HDR_SIZE = J2ME.Constants.OBJ_HDR_SIZE;
```
brendandahl: These should be marked as constants for closure to inline.  For all TS code we should should just use the full name Constant.XXX so it gets inlined by TS (and includes a nice debug comment about what the constant is).

jit/compiler.ts
```diff
@@ -64,8 +79,17 @@ module J2ME {
...
     // Emit class initializer.
-    writer.enter("function " + mangledClassName + "() {");
-    writer.writeLn("this._address = ASM._gcMalloc(" + classInfo.sizeOfFields + ");");
+    var klassId = generateKlassId();
+    writer.enter("function " + mangledClassName + "(address) {");
+    writer.writeLn("if (address) {");
+    writer.writeLn("  this._address = address;");
+    // writer.writeLn("  this._hashCode = i32[this._address + HASH_CODE_OFFSET >> 2];");
+    writer.writeLn("} else {");
+    writer.writeLn("  this._address = ASM._gcMalloc(OBJ_HDR_SIZE + " + classInfo.sizeOfFields + ");");
```
brendandahl: Use the actual OBJ_HDR_SIZE value here so there's no property lookup.

vm/runtime.ts
brendandahl: Now that I look around I don't see these used in JS, so let's just remove them from the globals.

vm/runtime.ts
mykmelez: > This checked extremely often, can we remove this or enable it only for debugging?

Indeed, and it isn't likely to be used very often (I only did so once, when I initially added it to troubleshoot a bug), so I've reverted it in ba95470 (leaving this version in a comment in case someone needs it in the future).

vm/runtime.ts
mykmelez: > Now that I look around I don't see these used in JS, so let's just remove them from the globals.

That's true for *HASH_CODE_OFFSET* too, so I've removed both globals (and made *emitKlass* use the actual *OBJ_HDR_SIZE* value) in a63369d.

mykmelez:
There was an emoji issue with one of the tests (TextEditorGfxTest), which didn't show up on Travis but did show up on my machine, where the number of differing pixels was a bit too high (because the three emoji in the text didn't appear). I fixed that issue in cb1d3c5.

PR #1673, opened 2015-06-24
remove unused confirm dialog pseudo element

mykmelez:
This :before pseudo-element isn't used, and Chrome computes its height differently than Firefox, which causes the SMS dialog to appear in the wrong place, so this branch removes it.


mykmelez:
Although the pseudo-element doesn't appear to be used, it does affect the appearance of alert dialogs, since it takes up vertical space, pushing down the dialog's *section*. We could do the same by adding a top margin to the section, but it isn't clear that we need that space, so instead of adding it back in, I've updated the gfx Alert tests in 2eb499f so they don't expect it.

PR #1676, opened 2015-06-26
pass addr to native methods

mykmelez:
This passes the address of the *this* object as the first argument to every native function call. It doesn't yet make the interpreter stop binding natives to handles, however. Nor does it make the natives use the *addr* value to get handles (except in a few test cases in nat.ts).

This is based on top of intex-object-layout (#1659), but I can't request a pull into that branch within the https://github.com/mozilla/pluotsorbet repository, since that branch is in my personal repository, so the diff view doesn't here show an accurate set of changes.

### DIFF
--- a/int.ts
+++ b/int.ts
@@ -1768,7 +1768,7 @@ module J2ME {
 
                 thread.set(fp, sp, opPC);
 
-                returnValue = callee.call(object, object ? object._address : 0);
+                returnValue = callee.call(object, object ? object._address : Constants.NULL);
               } else {
                 args.length = 0;
 
@@ -1822,7 +1822,7 @@ module J2ME {
                   // assert(callee.length === args.length, "Function " + callee + " (" + calleeTargetMethodInfo.implKey + "), should have " + args.length + " arguments.");
                 }
 
-                args.unshift(object ? object._address : 0);
+                args.unshift(object ? object._address : Constants.NULL);
                 returnValue = callee.apply(object, args);
               }
 
--- a/int.ts
+++ b/int.ts
@@ -1756,7 +1756,7 @@ module J2ME {
 
                 thread.set(fp, sp, opPC);
 
-                returnValue = callee.call(object);
+                returnValue = callee.call(object, object ? object._address : 0);
               } else {
                 args.length = 0;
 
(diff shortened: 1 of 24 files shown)


mykmelez:
The actual changes on this branch are currently contained within the single commit https://github.com/mozilla/pluotsorbet/commit/aa53559d5453193cea37f33771edf34c447f1e4e.

mykmelez:
This is a WIP because it depends on changes that @marco-c has on a branch for which he'll submit a pull request shortly.

mykmelez:
Now that intex-object-layout has been merged into intex, this branch contains just its own changes.

marco-c:
I believe this is ready now, can you merge the latest changes from intex?

mykmelez:
> I believe this is ready now, can you merge the latest changes from intex?

I merged the latest changes from the intex branch. I also merged #1696 so this branch would pass tests. This is "ready to merge", but it's also just the first step, and I'm wondering which of these additional steps it would make sense to do here vs. in followups:

* replace references to the *this* object inside natives with calls to *getHandle(addr)*;
* replace the *callee.call(…)* call with a direct call to *callee(addr)* in the part of the INVOKE* switch case that optimizes the call for a native with no arguments;
* figure out if it's possible to optimize the *callee.apply(object, args)* call in some way (at the very least, we should be able to stop binding it to the *this* object);
* stop passing an *addr* argument to static natives (per @brendandahl), which would require us to identify them first, probably via a script (or painstakingly by hand), since they aren't currently distinguished from other natives in any way.

brendandahl:
And
- Stop passing handles as arguments to natives.

marco-c:
There's certainly a lot of stuff to clean up (as in the rest of the intex branch), but merging this first could unlock other things (for example https://github.com/mozilla/pluotsorbet/pull/1690).
So, it depends on how much time you think the follow-ups are going to take.

mykmelez:
@marco-c Ok, if it unblocks other work, then let's merge it now, and I'll continue working on it in followups.

int.ts
```diff
@@ -1768,7 +1768,7 @@ module J2ME {
 
                 thread.set(fp, sp, opPC);
 
-                returnValue = callee.call(object);
+                returnValue = callee.call(object, object ? object._address : 0);
```
marco-c: Here and in the next change we should use Constants.NULL instead of 0.

mykmelez: > Here and in the next change we should use Constants.NULL instead of 0.

Yup, I did that in a2f756d.

mykmelez:
> @marco-c Ok, if it unblocks other work, then let's merge it now, and I'll continue working on it in followups.

Erm, in which case it's no longer WIP, so I've removed that designation from the summary.

PR #1677, opened 2015-06-26
Store arrays as addresses on the ref array

marco-c:


int.ts
```diff
@@ -714,59 +703,66 @@ module J2ME {
...
           case Bytecodes.DALOAD:
             index = i32[--sp];
-            array = ref[--sp];
-            if ((index >>> 0) >= (array.length >>> 0)) {
+            arrayAddr = ref[--sp];
+            if ((index >>> 0) >= (i32[arrayAddr >> 2] >>> 0)) {
               thread.throwException(fp, sp, opPC, ExceptionType.ArrayIndexOutOfBoundsException, index);
             }
-            aliasedF64[0] = array[index];
+            aliasedF64[0] = getArrayFromAddr(arrayAddr)[index];
```
mykmelez: Can we get the value directly from memory instead of indirecting through *getArrayFromAddr*? In this case we know (I think) that *arrayAddr* is an address.

int.ts
```diff
@@ -820,93 +816,88 @@ module J2ME {
...
             value = aliasedF64[0];
             index = i32[--sp];
-            array = ref[--sp];
-            if ((index >>> 0) >= (array.length >>> 0)) {
+            arrayAddr = ref[--sp];
+            if ((index >>> 0) >= (i32[arrayAddr >> 2] >>> 0)) {
               thread.throwException(fp, sp, opPC, ExceptionType.ArrayIndexOutOfBoundsException, index);
             }
-            array[index] = value;
+            getArrayFromAddr(arrayAddr)[index] = value;
```
mykmelez: Same question here, for which you probably have the same answer. :-)

int.ts
```diff
@@ -1786,7 +1808,16 @@ module J2ME {
...
-                        args.unshift(getHandle(address));
+                        if (address === 0) {
+                          args.unshift(null);
+                        } else {
+                          var tmp = getArrayFromAddr(address);
+                          if (tmp) {
+                            args.unshift(tmp);
+                          } else {
+                            args.unshift(getHandle(address));
+                          }
```
mykmelez: Nit: you could avoid the *tmp* with `args.unshift(getArrayFromAddr(address) || getHandle(address))`.

midp/gfx.js
```diff
@@ -23,9 +23,10 @@ var currentlyFocusedTextEditor;
     tempContext.canvas.height = 0;
 
     Native["com/sun/midp/lcdui/DisplayDeviceContainer.getDisplayDevicesIds0.()[I"] = function() {
-        var ids = J2ME.newIntArray( 1);
+        var idsAddr = J2ME.newIntArray( 1);
```
mykmelez: Nit: any reason to retain the spare space?

native.js
```diff
@@ -431,26 +442,32 @@ Native["java/lang/Throwable.fillInStackTrace.()V"] = function() {
 };
 
 Native["java/lang/Throwable.obtainBackTrace.()Ljava/lang/Object;"] = function() {
-    var result = null;
+    var resultAddr = 0;
+    // XXX: Untested.
```
mykmelez: Heh, *fillInStackTrace* is all REDUX anyway!

mykmelez: Incidentally, there's also now `$.ctx.nativeThread.frame.traceStack(new J2ME.IndentingWriter());`

int.ts
marco-c: Done in https://github.com/marco-c/pluotsorbet/commit/31af65155f995cc3b044e7caf0ea39991351bd2a.

midp/gfx.js
marco-c: Removed in 7218297.

int.ts
marco-c: Definitely! There wasn't actually a reason to use getArrayFromAddr, I probably used it while debugging something.
Fixed in 595647a.

mykmelez:
I'm seeing an unresponsive script error in *newArray* while running unit tests in Nightly on Mac. It happens during execution of the *com.sun.midp.crypto.TestDES* test, upon memory allocation (```var addr = ASM._gcMallocAtomic(offset + size * constructor.prototype.BYTES_PER_ELEMENT);```). (Tests pass when I run them in XULRunner via the automation script, however.)

mykmelez:
(I see a similar problem running WhatsApp, although it appears to work in Chrome.)

marco-c:
Unfortunately I can't reproduce your issue. I'll try reproducing it on Mac when I get home.

marco-c:
> Unfortunately I can't reproduce your issue. I'll try reproducing it on Mac when I get home.

I tried on my Mac, but I can't reproduce the issue (both tests and WhatsApp work fine).

mykmelez:
The test failure is an occasional intermittent failure, so I've restarted the build.

PR #1680, opened 2015-06-27
Make newString return addresses

marco-c:
Depends on https://github.com/mozilla/pluotsorbet/pull/1677.

### DIFF
--- a/midp/content.js
+++ b/midp/content.js
@@ -141,7 +141,7 @@ var Content = (function() {
       return -1;
     }
 
-    invocArguments[0] = J2ME.newString(invocation.argument)._address;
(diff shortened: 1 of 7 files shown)


int.ts
```diff
@@ -84,20 +84,10 @@ module J2ME {
       return reference;
     }
 
-    if (reference === null) {
-      return reference;
+    if (reference === null || reference === 0) {
+      return 0;
```
brendandahl: Can we make null a constant? It's going to be much harder to find these changes after this pull request.

mykmelez: > Can we make null a constant? It's going to be much harder to find these changes after this pull request.

@brendandahl Does the branch look good now with the merge from array_addresses and 7c0dab3?

PR #1684, opened 2015-06-28
Add some more checks to the TestIsolate testlet

marco-c:


mykmelez:
@marco-c I reverted this because it caused #1698. Can you figure out the test failure and then resubmit it?

PR #1685, opened 2015-06-28
Some cleanup after #1677

marco-c:


marco-c:
@mykmelez can you test this on your machine? It cleans #1677 up a bit, maybe this doesn't hang on your machine (although I wouldn't expect the result to be different than #1677).

mykmelez:
Aside: I wonder if it would improve readability to convert some of these integer literals into constants.

mykmelez:
(Not specific to this changeset, but I also wonder if *NULL* is really the best name for the constant, especially if we still use the JavaScript *null* value to represent the Java *null* value in some places.)

PR #1686, opened 2015-06-29
Update default app's name and origin to PluotSorbet

marco-c:


Makefile
```diff
@@ -33,10 +33,10 @@ ifeq ($(PACKAGE_TESTS),1)
   TESTS_JAR = tests/tests.jar
 endif
 
-NAME ?= j2me.js
+NAME ?= PluotSorbet
 MIDLET_NAME ?= midlet
 DESCRIPTION ?= j2me interpreter for firefox os
```
mykmelez: Aside: we should update this description too, making it "a J2ME-compatible virtual machine written in JavaScript".

mykmelez: > Aside: we should update this description too, making it "a J2ME-compatible virtual machine written in JavaScript".

Done in #1695.

PR #1690, opened 2015-06-30
Use addresses in Object::wait and Object::notify and cleanup invoke*

marco-c:
This depends on #1685 and #1676.

Passing addresses to Object::wait and Object::nofify allows us to remove the "_address" property from arrays.

### DIFF
--- a/int.ts
+++ b/int.ts
@@ -1709,7 +1709,7 @@ module J2ME {
               object = null;
             } else {
               address = ref[sp - calleeMethodInfo.argumentSlots];
-              if (address === null || address === Constants.NULL) {
+              if (address === Constants.NULL) {
                 object = null;
                 klass = null;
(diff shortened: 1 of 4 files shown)


marco-c:
#1685 and #1676 have been merged, so this is now ready.

int.ts
```diff
@@ -1706,19 +1706,10 @@ module J2ME {
             var callee = null;
 
             if (isStatic) {
-              object = null;
+              address = Constants.NULL;
```
mykmelez: After these changes, the only use of the *object* local in the *interpret()* function is in the *Bytecodes.SWAP* case, which could use the *address* local instead, so we should remove *object* completely.

marco-c: Right, done in 88773e4.

PR #1696, opened 2015-07-01
update SpiderMonkey version for jsshell

mykmelez:
The version of SpiderMonkey we were using, 37.0b7, seems to have disappeared from ftp.mozilla.org. The latest release from the 37 branch is 37.0.2, so let's use that instead.


marco-c:
What do you think about switching to a long term release so we don't have to do this too often?

mykmelez:
> What do you think about switching to a long term release so we don't have to do this too often?

Seems fine to me.

marco-c:
See https://github.com/mozilla/pluotsorbet/issues/1698#issuecomment-117861122 for the failure.

mykmelez:
7a20255 uses 38.1.0esr, the most recent long term release. But it fails locally on the same test.

PR #1701, opened 2015-07-02
Get rid of canonicalizeRef (almost)

marco-c:
There are still three places where it's used:
- *fastInterpreterFrameAdapter*, used by natives that call getLinkedMethod;
- getfield or getstatic, because some natives set JS objects as Java properties. Removing this instance of canonicalizeRef makes WeChat fail (but not tests or WhatsApp), so there are probably just a few natives that are doing that.
- invoke*, for natives that return objects instead of addresses.

I'm submitting this PR anyways because it removes all instances of Java objects and arrays from the *ref* array (there were still a few added by getfield or getstatic).
So, after this PR, on the ref array Java objects and arrays are always represented as addresses.

### DIFF
--- a/int.ts
+++ b/int.ts
@@ -1852,7 +1852,7 @@ module J2ME {
                   i32[sp++] = returnValue;
                   continue;
                 case Kind.Reference:
-                  release || assert(returnValue === "number", "native return value is a number");
+                  release || assert(returnValue !== "number", "native return value is a number");
                   ref[sp++] = returnValue;
                   continue;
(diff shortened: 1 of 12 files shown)


marco-c:
I've removed canonicalizeRef from *getfield* and *getstatic* (and, to avoid adding it to the field setters, I've merged #1680 into this).

marco-c:
dceb4b3 removes canonicalizeRef for natives' return values. Only *fastInterpreterFrameAdapter* missing now.

PR #1702, opened 2015-07-02
Revert "Add some more checks to the TestIsolate testlet"

mykmelez:
Reverts mozilla/pluotsorbet#1684


mykmelez:
(The test here will still fail because #1696 isn't landed.)

PR #1704, opened 2015-07-02
Use constants for accessing array values.

brendandahl:


vm/runtime.ts
```diff
@@ -2141,6 +2141,12 @@ module J2ME {
     // to the location of the hash code.
     HASH_CODE_OFFSET = 4,
 
+    ARRAY_HDR_SIZE = 8,
+
+    ARRAY_KLASS_ID_OFFSET = 0,
```
marco-c: This should be the same for arrays and objects.

PR #1705, opened 2015-07-02
remove references to *this* object from natives

mykmelez:
This branch removes references to the *this* object from natives. It's a followup to (and depends on) #1676.

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,9 @@ We use `Native` object in JS to handle creation and registration of `native` fun
 
 e.g.:
 
-    Native["java/lang/System.arraycopy.(Ljava/lang/Object;ILjava/lang/Object;II)V" = function(src, srcOffset, dst, dstOffset, length) {...};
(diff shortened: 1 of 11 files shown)


mykmelez:
And once there are no references to the *this* object, we can stop binding it to natives when we call them, which 72a03e8 does.

mykmelez:
Besides the two aforementioned changes, this branch also does:

* Update the README to document the *addr* parameter and stop telling readers about the *this* object in natives.
* In places where we were doing `getNative(this)`, instead of converting that to `var self = getHandle(addr) … getNative(self)`, convert it to get the native object directly via `NativeMap.get(addr)`.
* Refactor *calcStringWidth* in gfx.js to avoid having to get a handle to pass to it when the only thing it does with the handle is get its native.
* Update the theoretical implementations of *Throwable.fillinStackTrace* and *Throwable.obtainBackTrace* based on current native implementation requirements. (These implementations probably still don't work in practice.)
* Add the *addr* parameter to some (commented-out) string natives that I missed in #1676 (plus a comment that the string natives need to be updated to work if we ever decide to use them again).

PR #1706, opened 2015-07-02
Make null pointer checks in the interpreter explicit

marco-c:


marco-c:
This doesn't really require a review.

PR #1708, opened 2015-07-04
Use i32 instead of ref

marco-c:
There are still two test failures (tests/gnu/testlet/vm/FieldNotFoundException.java and tests/gnu/testlet/vm/MethodNotFoundException.java).

After this, we should be able to enable garbage collection.

(Depends on #1690)

### DIFF
--- a/int.ts
+++ b/int.ts
@@ -184,7 +184,7 @@ module J2ME {
     setParameter(kind: Kind, i: number, v: any) {
       switch (kind) {
         case Kind.Reference:
-          ref[this.fp + this.parameterOffset + i] = v;
+          i32[this.fp + this.parameterOffset + i] = v;
           break;
         case Kind.Int:
           i32[this.fp + this.parameterOffset + i] = v;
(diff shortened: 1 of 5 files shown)


int.ts
```diff
@@ -911,48 +911,48 @@ module J2ME {
             sp -= 2;
             continue;
           case Bytecodes.DUP: // ... a -> ... a, a
-            i32[sp] = i32[sp - 1];          ref[sp] = ref[sp - 1];
+            i32[sp] = i32[sp - 1];          i32[sp] = i32[sp - 1];
```
brendandahl: These need to be removed, not renamed.

marco-c: I admit I've simply blindly replaced "ref" with "i32", without actually looking at the code. I'll clean the patch up when I fix tests.

vm/parser.ts
```diff
@@ -636,7 +636,7 @@ module J2ME {
               r = classInfo.getMethodByName(name, type);
             }
             if (!r) {
-              throw $.newRuntimeException(classInfo.getClassNameSlow() + "." + fromUTF8(name) + "." + fromUTF8(type) + " not found");
+              throw classInfo.getClassNameSlow() + "." + fromUTF8(name) + "." + fromUTF8(type) + " not found";
```
brendandahl: How about we throw a JS object (maybe JavaRuntimeException???) and then in translate exception we'll convert that into a proper java exception. This way we don't have to introduce the try/catch's you added in int.ts.

marco-c: Done!

PR #1709, opened 2015-07-04
Use a node.js https server

marco-c:
The Python HTTPS server doesn't work on Fedora 22, so I've rewritten it in JS.

If it works on Travis and Mac, I'll rewrite the other SSL server as well.

### DIFF
--- a/tests/httpsServer.js
+++ b/tests/httpsServer.js
@@ -1,0 +1,21 @@
+const fs    = require("fs");
+const https = require("https");
+
+var pem = fs.readFileSync('cert.pem');
+
+var options = {
+  key: pem,
+  cert: pem,
+  secureProtocol: 'SSLv3_method',
+};
+
+https.createServer(options, function(req, res) {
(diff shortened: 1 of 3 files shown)


marco-c:
On Mac, SSLv3 is disabled (despite what the documentation says: https://nodejs.org/api/tls.html#tls_protocol_support).

mykmelez:
Presumably this is because OpenSSL on Mac has been compiled with those protocols disabled, so Node can't use them even if you configure it to do so. And the Homebrew configuration file for OpenSSL says that [it's keg-only](https://github.com/chdiza/homebrew/blob/master/Library/Formula/openssl.rb#L19-L20) because "Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries".

It should be possible to force Node to build with the version of OpenSSL provided by Homebrew, although I don't think we want to make PluotSorbet developers jump through this hoop. So I'd rather we keep the Python server working on Mac and use the Node server just on Linux.

PR #1712, opened 2015-07-06
pass objects/arrays to natives as addresses

mykmelez:
This changeset makes the interpreter pass objects/arrays to natives as addresses rather than handles.

To distinguish parameters that take addresses (and reduce the scope of changes), it appends "Addr" to the names of all object/array parameters and calls *getHandle* within natives to get handles as needed, frequently assigning them to a local variable with the same name as the parameter, minus the "Addr" suffix.

This changeset also replaces *fromJavaString* with *fromStringAddr* for conversion from a Java string (address) to a JS string.

In a few cases, it renames parameters to their names in the corresponding Java native method declarations. In a few cases, it inserts line breaks to limit line widths to <= 120 characters (since the "Addr" suffixes add length to already-long native method declarations). It also removes a java.lang.String.viewString function that doesn't seem to be used anywhere.

### DIFF
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -587,10 +587,12 @@ var currentlyFocusedTextEditor;
         return self.creator;
     };
 
-    Native["javax/microedition/lcdui/Graphics.setCreator.(Ljava/lang/Object;)V"] = function(addr, creator) {
+    Native["javax/microedition/lcdui/Graphics.setCreator.(Ljava/lang/Object;)V"] = function(addr, creatorAddr) {
         var self = getHandle(addr);
-        if (!self.creator) {
-            self.creator = creator;
+        // Per the original, non-native implementation of this method,
+        // ignore repeated attempts to set creator.
+        if (self.creator === J2ME.Constants.NULL) {
+            self.creator = creatorAddr;
         }
     };
 
@@ -957,7 +959,7 @@ var currentlyFocusedTextEditor;
         var self = getHandle(addr);
         self.displayId = displayId;
         NativeMap.set(addr, new GraphicsInfo(screenContextInfo));
-        self.creator = null;
+        self.creator = J2ME.Constants.NULL;
     };
 
     Native["javax/microedition/lcdui/Graphics.initImage0.(Ljavax/microedition/lcdui/Image;)V"] =
@@ -966,7 +968,7 @@ var currentlyFocusedTextEditor;
         var img = getHandle(imgAddr);
         self.displayId = -1;
         NativeMap.set(addr, new GraphicsInfo(getNative(getHandle(img.imageData))));
-        self.creator = null;
+        self.creator = J2ME.Constants.NULL;
     };
 
     function isScreenGraphics(g) {
--- a/midp/gfx.js
+++ b/midp/gfx.js
@@ -245,7 +245,9 @@ var currentlyFocusedTextEditor;
     }
 
     Native["javax/microedition/lcdui/ImageDataFactory.createImmutableImageDecodeImage.(Ljavax/microedition/lcdui/ImageData;[BII)V"] =
-    function(addr, imageData, bytes, offset, length) {
+    function(addr, imageDataAddr, bytesAddr, offset, length) {
+        var imageData = getHandle(imageDataAddr);
+        var bytes = J2ME.getArrayFromAddr(bytesAddr);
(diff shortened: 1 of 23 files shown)


midp/gfx.js
```diff
@@ -1619,8 +1648,8 @@ var currentlyFocusedTextEditor;
     var CONTINUOUS_RUNNING = 2;
 
     Native["javax/microedition/lcdui/GaugeLFImpl.createNativeResource0.(ILjava/lang/String;IZII)I"] =
-    function(addr, ownerId, label, layout, interactive, maxValue, initialValue) {
-        if (label !== null) {
+    function(addr, ownerId, labelAddr, layout, interactive, maxValue, initialValue) {
+        if (labelAddr !== null) {
```
brendandahl: This now needs to be compared against the null constant.

midp/media.js
```diff
@@ -1067,15 +1068,16 @@ Native["com/sun/mmedia/MediaDownload.nGetFirstPacketSize.(I)I"] = function(addr,
...
 };
 
-Native["com/sun/mmedia/MediaDownload.nBuffering.(I[BII)I"] = function(addr, handle, buffer, offset, size) {
+Native["com/sun/mmedia/MediaDownload.nBuffering.(I[BII)I"] = function(addr, handle, bufferAddr, offset, size) {
     var player = Media.PlayerCache[handle];
     var bufferSize = player.getBufferSize();
 
     // Check the parameters.
-    if (buffer === null || size === 0) {
+    if (bufferAddr === null || size === 0) {
```
brendandahl: Null constant.

midp/gfx.js
```diff
@@ -775,26 +789,32 @@ var currentlyFocusedTextEditor;
...
         renderRegion(NativeMap.get(addr).getGraphicsContext(), getNative(getHandle(image.imageData)).context.canvas,
                      0, 0, image.width, image.height, TRANS_NONE, x, y, anchor);
         return 1;
     };
 
-    Native["javax/microedition/lcdui/Graphics.drawRegion.(Ljavax/microedition/lcdui/Image;IIIIIIII)V"] = function(addr, src, x_src, y_src, width, height, transform, x_dest, y_dest, anchor) {
-        if (null === src) {
+    Native["javax/microedition/lcdui/Graphics.drawRegion.(Ljavax/microedition/lcdui/Image;IIIIIIII)V"] =
+    function(addr, srcAddr, x_src, y_src, width, height, transform, x_dest, y_dest, anchor) {
+        if (srcAddr === null) {
```
brendandahl: Null constant.

midp/gfx.js
```diff
@@ -775,26 +789,32 @@ var currentlyFocusedTextEditor;
...
+        var src = getHandle(srcAddr);
         renderRegion(NativeMap.get(addr).getGraphicsContext(), getNative(getHandle(src.imageData)).context.canvas,
                      x_src, y_src, width, height, transform, x_dest, y_dest, anchor);
     };
 
-    Native["javax/microedition/lcdui/Graphics.drawImage.(Ljavax/microedition/lcdui/Image;III)V"] = function(addr, image, x, y, anchor) {
-        if (image === null) {
+    Native["javax/microedition/lcdui/Graphics.drawImage.(Ljavax/microedition/lcdui/Image;III)V"] =
+    function(addr, imageAddr, x, y, anchor) {
+        if (imageAddr === null) {
```
brendandahl: Null constant.

midp/gfx.js
```diff
@@ -940,8 +960,10 @@ var currentlyFocusedTextEditor;
         self.creator = null;
```
brendandahl: Don't really need to address in this PR, but should this be set to the null constant?

mykmelez:
Note to self: there are some other null checks that currently check for truthiness, and I should also convert those to compare their addresses to *Constants.NULL*.

midp/gfx.js
mykmelez: > This now needs to be compared against the null constant.

adcf2ec updates all these conditionals to compare addresses with the null constant.

midp/gfx.js
mykmelez: > Don't really need to address in this PR, but should this be set to the null constant?

Nice catch, yes, it should, per the implementations of Graphics.getCreator and Graphics.setCreator. 8c73b14 makes the change.

PR #1719, opened 2015-07-07
Remove canonicalizeRef.

brendandahl:


native.js
```diff
@@ -405,7 +405,7 @@ Native["java/lang/Class.newInstance1.(Ljava/lang/Object;)V"] = function(addr, oA
   if (!methodInfo) {
     throw $.newInstantiationException("Can't instantiate classes without a nullary constructor");
   }
-  J2ME.getLinkedMethod(methodInfo).call(o);
+  J2ME.getLinkedMethod(methodInfo)(oAddr);
```
marco-c: We should file a follow to clean up the natives that use getHandle only to get the klass property of the object (and use klassIdMap instead).

vm/context.ts
```diff
@@ -497,7 +497,7 @@ module J2ME {
       var exception = new classInfo.klass();
       var methodInfo = classInfo.getMethodByNameString("<init>", "(Ljava/lang/String;)V");
       preemptionLockLevel++;
-      getLinkedMethod(methodInfo).call(exception, message ? newString(message) : null);
+      getLinkedMethod(methodInfo)(exception._address, message ? newString(message) : null);
```
marco-c: Here the NULL constant instead of null.

PR #1720, opened 2015-07-08
[WhatsApp] Start the background alarm only after the verification process

marco-c:
I'm using a simple hack: creating a file (unaccessible to MIDlets) called "smsReg".

### DIFF
--- a/midp/background.js
+++ b/midp/background.js
@@ -54,7 +54,18 @@ function backgroundCheck() {
   fgMidletNumber = (bgServer == 2) ? 1 : 2;
   fgMidletClass = MIDP.manifest["MIDlet-" + fgMidletNumber].split(",")[2];
 
-  DumbPipe.close(DumbPipe.open("backgroundCheck", {}));
+  if (fs.exists("/smsReg")) {
+    startBackgroundAlarm();
+  }
+}
+
+var backgroundAlarmStarted = false;
+
+function startBackgroundAlarm() {
+  if (!backgroundAlarmStarted) {
+    backgroundAlarmStarted = true;
+    DumbPipe.close(DumbPipe.open("backgroundCheck", {}));
+  }
 }
 
 Native["com/nokia/mid/s40/bg/BGUtils.getFGMIDletClass.()Ljava/lang/String;"] = function() {
--- a/midp/background.js
+++ b/midp/background.js
@@ -54,8 +54,7 @@ function backgroundCheck() {
   fgMidletNumber = (bgServer == 2) ? 1 : 2;
   fgMidletClass = MIDP.manifest["MIDlet-" + fgMidletNumber].split(",")[2];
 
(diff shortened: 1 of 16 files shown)


mykmelez:
Hmm, this breaks existing installations that have already gone through the registration process, which now never start in the background.

marco-c:
> Hmm, this breaks existing installations that have already gone through the registration process, which now never start in the background.

Right, I hadn't thought about that.

mykmelez:
Is there an API that WhatsApp only uses after registration is complete, but which it does use on every startup afterwards? Perhaps some contacts API?

marco-c:
> Is there an API that WhatsApp only uses after registration is complete, but which it does use on every startup afterwards? Perhaps some contacts API?

Unfortunately the contacts API is used also on first run (before registration). There could be some API that is used only after registration, but I can't think of anything at the moment.
One thing we could do, very hacky, is checking if one of the WA files that are created after registration exists.

mykmelez:
> One thing we could do, very hacky, is checking if one of the WA files that are created after registration exists.

That doesn't seem much hackier than checking if our own file exists.

marco-c:
I'll update this PR (since I can reuse the test code)

midp/background.js
```diff
@@ -54,7 +54,19 @@ function backgroundCheck() {
   fgMidletNumber = (bgServer == 2) ? 1 : 2;
   fgMidletClass = MIDP.manifest["MIDlet-" + fgMidletNumber].split(",")[2];
 
-  DumbPipe.close(DumbPipe.open("backgroundCheck", {}));
+  if (!MIDlet.shouldStartBackgroundService ||
```
mykmelez: Instead of checking if the property is truthy, I would add a default function (to config/default.js) that always returns false.

marco-c: Done in 5849ea0 (with a default of *true*)

mykmelez: Why `true` by default? Messaging midlets like WhatsApp seem like the exception rather than the rule, so my inclination would be to set it to `false` by default, so other midlets don't start in the background unless we explicitly enable them to do so.

marco-c: With it set to *true*, we'd only start in the background MIDlets that have the *Nokia-MIDlet-bg-server* property in their manifest (honoring the property seemed the best default option to me, while the heuristic is more of a hack to avoid starting MIDlets in the background when it isn't actually needed).

mykmelez: > With it set to true, we'd only start in the background MIDlets that have the Nokia-MIDlet-bg-server property in their manifest (honoring the property seemed the best default option to me, while the heuristic is more of a hack to avoid starting MIDlets in the background when it isn't actually needed).

Ah, that makes sense, thanks for the explanation! Can you add this to the comment on the default version of the function? That way the next hacker who sees it will understand why we set the default this way.

marco-c: Sure! Done in 593c6e4.

PR #1722, opened 2015-07-08
pass objects/arrays to natives as addresses

mykmelez:
Re-request of intex-pass-addresses branch from #1712 to address issue that caused it to be reverted in #1721. WIP until issue has been addressed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ We use `Native` object in JS to handle creation and registration of `native` fun
 
 e.g.:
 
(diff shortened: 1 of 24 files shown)


mykmelez:
Regression fixed by 3e4ace1. This is ready for unreversion.

mykmelez:
Erm, except I see another regression: when pressing the "back" button to quit WeChat on its initial screen, the midlet doesn't exit.

mykmelez:
> Erm, except I see another regression: when pressing the "back" button to quit WeChat on its initial screen, the midlet doesn't exit.

This one is fixed by e8c15b5.

PR #1725, opened 2015-07-08
Always pass addresses to getMonitor

marco-c:


int.ts
```diff
@@ -443,9 +443,9 @@ module J2ME {
         frame.setParameter(kinds[i], index++, arguments[i]);
       }
       if (methodInfo.isSynchronized) {
-        var monitor = methodInfo.isStatic ? methodInfo.classInfo.getClassObject() : getMonitor(arguments[0]);
+        var monitor = methodInfo.isStatic ? methodInfo.classInfo.getClassObject()._address : arguments[0];
```
mykmelez: Nit: I would rename this (and the other *monitor* variable below) to *monitorAddr*, since otherwise the `getMonitor(monitor)` calls are a bit confusing.

vm/runtime.ts
```diff
@@ -750,33 +750,10 @@ module J2ME {
...
-      return monitorMap[ref._address] || (monitorMap[ref._address] = Object.create(null));
-    }
-
-    // An object with a _lock property is itself a monitor.
-    // XXX Seems dicey, warn or assert?
-    if ("_lock" in ref) {
-      return ref;
-    }
+  export function getMonitor(ref: number): any {
+    release || assert(typeof ref === "number", "monitor reference is not a number");
```
mykmelez: Nit: make this message describe what the assertion expects (i.e. what it asserts): "monitor reference is a number".

PR #1730, opened 2015-07-11
Use requestAnimationFrame in DisplayDevice.refresh

TimAbraldes:
This way we only paint when it actually makes sense (i.e. we use
requestAnimationFrame how it was intended).

This patch removes the waitForAnimationFrame function but still accomplishes the throttling that that function intended to provide. This patch also ensures that we paint at the appropriate time. Previously we were queuing a thread for execution during the requestAnimationFrame callback, but then we weren't actually painting, so we were probably painting at inopportune times.

On my Macbook Pro, this drew smoother animations for the Asteroids MIDlet.

This is the result of the startup bench on a Flame device. 100 rounds, 25000ms delay, GC enabled:

|         Test         | Baseline Mean |    Mean    |    +/-    |    %    |     P    |     Min    |     Max     | 
|:---------------------|--------------:|-----------:|----------:|--------:|---------:|-----------:|------------:|
| startupTime          |      16,598ms |   16,874ms |     276ms |    1.66 |     SAME |   14,529ms |    19,236ms | 
| vmStartupTime        |       2,783ms |    2,543ms |    -240ms |   -8.62 |   BETTER |    2,049ms |     2,895ms | 
| bgStartupTime        |         897ms |      874ms |     -23ms |   -2.52 |     SAME |      630ms |     1,086ms | 
| fgStartupTime        |      11,483ms |   12,016ms |     532ms |    4.64 |    WORSE |   10,040ms |    14,775ms | 
| fgRefreshStartupTime |       1,435ms |    1,441ms |       6ms |    0.42 |     SAME |      648ms |     2,706ms |

### DIFF
--- a/java/custom/com/sun/midp/lcdui/RepaintEventProducer.java
+++ b/java/custom/com/sun/midp/lcdui/RepaintEventProducer.java
@@ -174,10 +174,6 @@ public class RepaintEventProducer implements EventListener {
         return true;
     }
 
-    // Called from `process` to ensure that we don't try to repaint
-    // more than once per animation frame
-    private native void waitForAnimationFrame();
-
     /**
      * Process an event.
      *
@@ -185,8 +181,6 @@ public class RepaintEventProducer implements EventListener {
      */
     public void process(Event genericEvent) {
         RepaintEvent event = (RepaintEvent)genericEvent;
-
-        waitForAnimationFrame();
 
         synchronized (this) {
             queuedEvent = null;
--- a/jit/analyze.ts
+++ b/jit/analyze.ts
@@ -25,7 +25,7 @@ module J2ME {
    * at the right spots.
    */
   export var yieldMap = {
-    "com/sun/midp/lcdui/RepaintEventProducer.waitForAnimationFrame.()V": YieldReason.Root,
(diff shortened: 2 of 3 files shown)


brendandahl:
How's scrolling with this on the flame? I remember we tried putting the request animation frame here before, but then the repaint events don't get merged and we execute the repaint code more frequently than we can repaint.  Though I also remember with the new scheduling and preemption I was able to remove the request animation frames stuff all together and things were working fine.

TimAbraldes:
> How's scrolling with this on the flame?

I compared master to a build of this branch. Scrolling perf might have been slightly worse with this branch only. However, with this branch and #1735 applied, scroll perf seemed to be as good as master. This is all just my manual, visual testing so YMMV

mykmelez:
Scrolling perf with this branch seems a bit better for me on an Aries, although it's within the margin of error for a subjective assessment. It seems about the same if I also merge #1735.

marco-c:
Have you checked if the repaint events get merged after this change?

TimAbraldes:
> Have you checked if the repaint events get merged after this change?

I put a println here - https://github.com/mozilla/pluotsorbet/blob/master/java/custom/com/sun/midp/lcdui/RepaintEventProducer.java#L133 - and then ran WhatsApp and scrolled the country list; I saw the message print multiple times.

marco-c:
Great!

PR #1731, opened 2015-07-13
Enable GC for some objects

marco-c:
This PR enables garbage collection for objects created via *allocObject*.

### DIFF
--- a/tests/gctests.js
+++ b/tests/gctests.js
@@ -118,6 +118,22 @@ tests.push(function() {
   next();
 });
 
+tests.push(function() {
+  var addr = ASM._gcMallocUncollectable(4);
+  var objAddr = J2ME.allocObject(CLASSES.java_lang_Object.klass);
+  i32[addr >> 2] = objAddr;
+  ASM._forceCollection();
+  isNot(freedAddr, addr, "Object allocated with GC_MALLOC_UNCOLLECTABLE isn't collected");
+  isNot(freedAddr, objAddr, "Object referenced by someone isn't freed");
+
+  i32[addr >> 2] = 0;
+
+  ASM._forceCollection();
+  is(freedAddr, objAddr, "Object that isn't referenced by anyone anymore is freed");
+
+  next();
+});
+
 // XXX: Replace ASM._gcMalloc with J2ME.newArray when J2ME.newArray will
 //      use _gcMalloc and not _gcMallocUncollectable.
 tests.push(function() {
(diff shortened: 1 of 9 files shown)


tests/gctests.js
```diff
@@ -0,0 +1,162 @@
...
+  } else {
+    ++failed;
+    print("fail " + msg);
+  }
+}
+
+var tests = [];
+
+function next() {
+  for (var i = 0; i < 100; i++) {
```
mykmelez: Is there any particular relevance to this loop count?

tests/gctests.js
```diff
@@ -0,0 +1,162 @@
...
+  var objAddr = J2ME.allocObject(CLASSES.java_lang_Object.klass);
+  i32[addr >> 2] = objAddr;
+  ASM._forceCollection();
+  isNot(freedAddr, addr, "Object allocated with GC_MALLOC_UNCOLLECTABLE isn't collected");
+  isNot(freedAddr, objAddr, "Object referenced by someone isn't freed");
+
+  ASM._gcFree(addr);
+
+  ASM._forceCollection();
+  is(freedAddr, objAddr, "Object that isn't referenced by anyone anymore is freed");
```
mykmelez: It'd also be useful to check that objAddr is freed if *addr* isn't freed, but *i32[addr >> 2]* is reset such that it no longer references *objAddr*.

tests/runtests.py
```diff
@@ -133,7 +133,20 @@ def run_test(script_path):
...
-if exit_code == 0 and not shell_success:
+if not shell_success:
+    sys.stdout.write("FAIL - Basic shell test failed\n")
+    exit_code = 1
+
+p = subprocess.Popen(['js', os.path.join('tests','gctests.js')], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+gc_tests_success = True
+for line in iter(p.stdout.readline, b''):
+    if "fail" in line:
+        sys.stdout.write("PORCO DIO")
```
mykmelez: Erm… I understand the sentiment, but let's still omit it. ;-)

marco-c: Oops, I'll remove it immediately :smile:

tests/gctests.js
marco-c: I added this loop to make the tests as independent as possible of each other, but I'll remove it now because I think it's not enough.

marco-c: Done in 2adca73f82286adaf79899c5e92e42c02f152ca0.

tests/gctests.js
marco-c: Done in ccab8d871fdb81dc0e337ad4b23abc34e0e2d898.

PR #1735, opened 2015-07-15
Coalesce pen drag events

TimAbraldes:
@brendandahl made a comment in another PR that made me think this might be a good idea. WIP because there's probably a better implementation (coalescing drag events when putting them into the queue rather than when removing them). This is just investigative work so far


TimAbraldes:
Removing the WIP tag. There is a tradeoff here that's worth mentioning: With this patch applied, we avoid doing a bunch of work to process stale pen drag events, which means that things like scroll/pan/zoom should be less taxing on the CPU. The downside is that, if the VM isn't able to keep up with the number of mousemove events we're receiving, we won't get smooth motion paths anymore. This could affect paint-like apps.

midp/midp.js
```diff
@@ -927,7 +927,17 @@ var MIDP = (function() {
       var nativeEventQueue = nativeEventQueues[isolateId];
 
       if (nativeEventQueue.length !== 0) {
-        copyEvent(nativeEventQueue.shift(), nativeEvent);
+        var ev = nativeEventQueue.shift();
+        if (ev.type === PEN_EVENT
+         && ev.intParam1 === DRAGGED) {
```
marco-c: Nit: the indentation here and in the *while* is off, we usually do:
```
if (aFirstCondition &&
    aSecondCondition) {
```

In this particular case, the two conditions could fit in one line.

midp/midp.js
```diff
@@ -927,7 +927,17 @@ var MIDP = (function() {
...
+          while (nativeEventQueue.length !== 0
+              && nativeEventQueue[0].type === PEN_EVENT
+              && nativeEventQueue[0].intParam1 === DRAGGED) {
+            ev = ev2;
```
marco-c: There's probably something missing here, *ev2* isn't defined.

marco-c:
It'd be great if you could also write a couple of tests (e.g. you could create a fake native that sends a few PEN_EVENT events and check that you only receive one event).

TimAbraldes:
> It'd be great if you could also write a couple of tests (e.g. you could create a fake native that sends a few PEN_EVENT events and check that you only receive one event).

Good idea! I'll do this shortly.

midp/midp.js
TimAbraldes: > There's probably something missing here, ev2 isn't defined.

Not sure what happened here. I updated the patch with a new approach so let's discuss there.

midp/midp.js
```diff
@@ -759,28 +759,107 @@ var MIDP = (function() {
...
+    // function otherwise
+    function wait(id, resolve, e) {
+      waiting[id] = { resolve: resolve, e: e };
+    }
+
+    function put(e, id) {
+      var waiter = waiting[id];
+      if (waiter) {
+        copyEvent(e, waiter.e);
+        waiter.resolve(0);
```
marco-c: This should be resolved with the number of events left in the queue.

midp/midp.js
```diff
@@ -759,28 +759,107 @@ var MIDP = (function() {
...
+      } else {
+        delete gestureDrag[id];
+        delete penDrag[id];
+      }
+
+      q.push(e);
+    }
+
+    return {
+      send: put,
```
marco-c: Nit: can we use the same name here?

tests/midlets/drag/coalesce.jad
```diff
@@ -0,0 +1,10 @@
+Manifest-Version: 1.0
```
marco-c: Nit: the JAD manifest should be unneeded for this test

tests/automation.js
```diff
@@ -295,6 +295,16 @@ casper.test.begin("unit tests", 39 + gfxTests.length, function(test) {
     });
 
     casper
+    .thenOpen("http://localhost:8000/index.html?midletClassName=tests/drag/Coalesce&jad=tests/midlets/drag/coalesce.jad&jars=tests/tests.jar&logConsole=web,page&logLevel=log")
```
marco-c: Nit: this only tests pen events, we could test gesture events as well (they should work in the same way, so it's less important)

tests/native.js
```diff
@@ -290,3 +290,27 @@ Native["tests/midlets/background/ForegroundEnableBackgroundServiceMIDlet.started
...
+  setTimeout(function() {
+    MIDP.sendPenEvent({x: 0, y: 0}, PRESSED);
+    MIDP.sendPenEvent({x: 1, y: 1}, DRAGGED);
+    MIDP.sendPenEvent({x: 3, y: 5}, DRAGGED);
+    MIDP.sendPenEvent({x: 7, y: 9}, DRAGGED);
+    MIDP.sendPenEvent({x: 30, y: 30}, DRAGGED);
+    MIDP.sendPenEvent({x: 13, y: 27}, DRAGGED);
+    MIDP.sendPenEvent({x: 13, y: 27}, RELEASED);
+  }, 100);
+}
```
marco-c: Could you add a test where there are multiple events and pen drag events aren't coalesced?
For example, two PRESSED-DRAGGED-RELEASED sequences.

tests/automation.js
TimAbraldes: > Nit: this only tests pen events, we could test gesture events as well (they should work in the same way, so it's less important)

Yeah. I wasn't sure how to make the MIDlet receive/process gesture events, but we should definitely test them. I'll look into adding that.

tests/native.js
TimAbraldes: > Could you add a test where there are multiple events and pen drag events aren't coalesced?
For example, two PRESSED-DRAGGED-RELEASED sequences.

Sure! Good idea!

tests/midlets/drag/coalesce.jad
TimAbraldes: > Nit: the JAD manifest should be unneeded for this test

I'll remove it

midp/midp.js
TimAbraldes: > This should be resolved with the number of events left in the queue.

Shouldn't that number always be 0 in this case?

midp/midp.js
TimAbraldes: > Nit: can we use the same name here?

Yeah, I'll rename `put` to `send`

midp/midp.js
marco-c: The old implementation did: `.resolve(nativeEventQueues[isolateId].length);`, but now that I think more about it you're probably right.
This will be called as soon as an event is sent to an empty queue.

tests/automation.js
marco-c: There should be an example MIDlet somewhere.

TimAbraldes:
Updated; I think all review comments are now addressed

mykmelez:

PR #1739, opened 2015-07-17
Enable GC for arrays

marco-c:


midp/sensor.js
```diff
@@ -249,9 +250,10 @@ function(addr, sensorsNumber, number, modelAddr) {
     var n = c.mrangeArray.length;
     var arrayAddr = J2ME.newArray(J2ME.PrimitiveClassInfo.J.klass, n);
     var array = J2ME.getArrayFromAddr(arrayAddr);
+    var i32array = new Int32Array(array.buffer, array.byteOffset, array.length * 2);
```
marco-c: This is unrelated to the PR, but I think the current code on intex had a bug so I fixed it while I was here.

marco-c:
There are two problems related to GC and natives.

1) Natives that allocate more than one object need to somehow disable GC for them. For example:
```JavaScript
var arr = J2ME.newIntArray(4);
var obj = J2ME.allocObject(CLASSES.java_lang_Object.klass);
```
Any allocation could trigger a collection, so if you don't protect *arr* here, it could be released when you call *allocObject*.
The natives I've touched only do stuff like:
```JavaScript
var arrayAddr = J2ME.newStringArray(4);
var array = J2ME.getArrayFromAddr(arrayAddr);
for (var i = 0; i < 4; i++) {
  array[i] = J2ME.newString("ciao");
}
```
In this case, the objects created by newString are automatically protected because they're stored in the asm.js heap through `array[i]`.
Since I only needed to protect one object at a time in the natives that I touched, I've added the functions `setUncollectable` and `unsetUncollectable` that allow to protect only one object at a time.

2) Natives that allocate an object and return it via *asyncImpl*. We need to protect these as well because, while the thread is inactive, other threads could run and allocate other objects that could potentially trigger a collection.
I've found only one native that needed protecting. To protect the object, I allocated an uncollectable area and referenced the object from there. I then freed the uncollectable area in a newly introduced *cleanup* function called by asyncImpl.


(By "object" I mean "object or array")

marco-c:
For (1) instead of `setUncollectable` and `unsetUncollectable` we could disable/reenable the GC, I'm not sure what's the best solution.

nat.ts
```diff
@@ -28,8 +28,8 @@ module J2ME {
   import Bytecodes = Bytecode.Bytecodes;
   import toHEX = IntegerUtilities.toHEX;
 
-  export function asyncImplOld(returnKind: string, promise: Promise<any>) {
-    return asyncImpl(kindCharacterToKind(returnKind), promise);
+  export function asyncImplOld(returnKind: string, promise: Promise<any>, cleanup?: Function) {
```
mykmelez: Hmm, the caller could handle its own cleanup by chaining another function to the promise via *then*. But then it'd have to make sure to return the resolved value. This *cleanup* parameter is more explicit, so I like it better. (But I wish DOM Promise had *finally*, which would be even better, since then the caller could chain it without having to worry about it returning the resolved value.)

tests/gctests.js
```diff
@@ -148,7 +209,80 @@ tests.push(function() {
...
+    }
+
+    for (var j = 0; j < 8; j += 4) {
+      i32[addr + j >> 2] = 0xDEADBEEF;
+    }
+
+    ASM._forceCollection();
+  }
+
+  ok(!zeroedOut, "gcMallocAtomic doesn't zero-out allocated memory");
```
mykmelez: If I read this test correctly, it only checks that gcMallocAtomic sometimes doesn't zero-out allocated memory. So the test will pass even if gcMallocAtomic sometimes does zero-out such memory, as long as it sometimes doesn't (within up to 1000 allocations). Is there any way to make the test more deterministic?

nat.ts
marco-c: Another reason why I added the cleanup function is that it needs to be called after the reference is in the asm.js heap (so after the line *i32[sp++] = l;*).
In practice, we need to avoid running any Java code between the call to _gcFree and the line I mentioned above (otherwise a GC could happen in the meantime freeing the array).

tests/gctests.js
marco-c: Since I couldn't control where the memory was allocated, I tried allocating it multiple times hoping that within 1000 iterations the areas allocated by the gcMallocAtomic calls would collide (in my testing this happens within 2 iterations, so I used 1000 to increase the possibility of success).

One way to make the test more robust would be to set the entire asm.js heap to a value != 0 and check that the memory allocated via gcMallocAtomic isn't initialized to 0 (this would be more robust, but I think still as deterministic as the other).

mykmelez: Hmm, I like that alternative, although perhaps it's slow? If so, then maybe we can run the tests with a smaller heap.

marco-c: Now that I think more about it, it isn't really feasible because by clearing the heap we mess with data allocated by the garbage collector for its own use.

mykmelez: Ok, let's stick with the current test, but add a comment explaining why we're doing it this way.

marco-c: Done in 44b52faf7d313d592116bf49ed4f65eaa02a1e67.

PR #1742, opened 2015-07-18
Intex remove klass squash

brendandahl:
Depends on #1739
Still having some issues with GC enabled. If I use a regular malloc for primitive arrays it seems to work.


marco-c:
I'll pick up the work where it's left and open another PR.

PR #1744, opened 2015-07-19
Allocate collectable thread objects and enable collection of thread stacks

marco-c:
This also contains code to support collecting thread stacks, but it isn't enabled yet because there's still one bug left to fix.

Depends on #1739 and #1741.

### DIFF
--- a/int.ts
+++ b/int.ts
@@ -190,7 +190,7 @@ module J2ME {
       var fp = this.fp;
       var sp = this.sp;
       var pc = this.pc;
-      while (this.fp > this.thread.tp) {
+      while (this.fp > (this.thread.tp >> 2)) {
         writer.writeLn((this.methodInfo ? this.methodInfo.implKey : "null") + ", FP: " + this.fp + ", SP: " + this.sp + ", PC: " + this.pc);
         this.set(this.thread, i32[this.fp + FrameLayout.CallerFPOffset],
                  this.fp + this.parameterOffset,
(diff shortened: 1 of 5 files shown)


marco-c:
In detail:
- 61078ee2a2491adec10dec1ae4e016315875960d allocates thread objects with allocObject instead of allocObjectUncollectable and removes allocObjectUncollectable altogether.
- ce326ffcd474836aad693a223c63890fe078ef0e stores the thread stack base pointer in the *threadData* memory area, which is freed when a context is deleted. The thread stack is still allocated with GC_MALLOC_UNCOLLECTABLE.
- d76db7c0f99297aa61daa528c6863ebd72702b68 fixes a bug that would be present if we used GC_MALLOC to allocate the thread stack.

There's still one bug left to actually use GC_MALLOC to allocate the thread stack. It's one JCK test failure that is pretty difficult to debug (and doesn't show up if you use GC_MALLOC instead of GC_MALLOC_ATOMIC to allocate primitive arrays).
In particular, I've noticed two objects are allocated consecutively by allocObject with the same address, even if the first object isn't finalized before the second is allocated (so the second object overwrites the memory of the first object).
In TestCase.java:
```Java
	public final boolean isTestValid(String testName) {
		StringBuffer fullName = new StringBuffer();
		fullName.append(this.getClass().getName());
		fullName.append("#");
		fullName.append(testName);
		currentTestName = fullName.toString();
		testLog = new StringBuffer();
		boolean result = !runner.isExcluded(currentTestName) && 
						  runner.isFiltered(currentTestName);
		startTime = System.currentTimeMillis();
		return result;
	}
```

The two objects are *fullName* in the above function and a String object in *StringBuffer.toString()*.

Anyway, I think we can start by landing this PR and work on enabling GC_MALLOC for thread stacks in a follow-up.

marco-c:
Fixed the last bug with 19ef982 and enabled collection of thread stacks (which indirectly makes a lot more objects collectable) in f402f18.

mykmelez:
Merged with trailing semicolon added.

PR #1747, opened 2015-07-20
Register finalizers only when needed

marco-c:
Depends on #1744.


mykmelez:
At some point we'll also want to implement support for *finalize* methods. But we can put that off until after we merge intex back to master, since we don't support those today, so they'll be a new feature, not a regression.

PR #1751, opened 2015-07-21
Require a message in assert

TimAbraldes:
Because "assertion failed" is almost as disheartening and unhelpful
as "Segmentation fault"


marco-c:
Looks good to me, but it's likely going to conflict when we'll merge master into intex. Could you also open a PR to do that?

PR #1752, opened 2015-07-21
Remove Klass

marco-c:
This is now passing tests locally with GC enabled.

### DIFF
--- a/int.ts
+++ b/int.ts
@@ -18,7 +18,7 @@ module J2ME {
         x.push(String.fromCharCode(o[i]));
       }
       var suffix = (o.length > 8 ? "..." : "");
-      return fromUTF8(o.klass.classInfo.utf8Name) +
+      return fromUTF8(o.classInfo.utf8Name) +
         ", length: " + o.length +
         ", values: [" + s.join(", ") + suffix + "]" +
         ", chars: \"" + x.join("") + suffix + "\"";
@@ -27,15 +27,15 @@ module J2ME {
       if (o.length !== undefined) {
         return getArrayInfo(o);
       }
-      return fromUTF8(o.klass.classInfo.utf8Name) + (o._address ? " " + toHEX(o._address) : "");
+      return fromUTF8(o.classInfo.utf8Name) + (o._address ? " " + toHEX(o._address) : "");
     }
-    if (o && !o.klass) {
+    if (o && !o.classInfo) {
       return o;
     }
-    if (o && o.klass === Klasses.java.lang.Class) {
-      return "[" + getObjectInfo(o) + "] " + o.runtimeKlass.templateKlass.classInfo.getClassNameSlow();
+    if (o && o.classInfo === CLASSES.java_lang_Class) {
+      return "[" + getObjectInfo(o) + "] " + classIdToClassInfoMap[o.vmClass].getClassNameSlow();
     }
-    if (o && o.klass === Klasses.java.lang.String) {
+    if (o && o.classInfo === CLASSES.java_lang_String) {
       return "[" + getObjectInfo(o) + "] \"" + fromStringAddr(o._address) + "\"";
     }
     return o ? ("[" + getObjectInfo(o) + "]") : "null";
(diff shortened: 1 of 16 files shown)


marco-c:
Most of the code here was written by @brendandahl in https://github.com/mozilla/pluotsorbet/pull/1742, I've only fixed a couple of bugs with garbage collection and cleaned up the patch.

int.ts
```diff
@@ -522,12 +520,13 @@ module J2ME {
...
     var tag: TAGS;
     var type, size;
-    var value, index, arrayAddr: number, klass, offset, buffer, tag: TAGS, targetPC;
+    var value, index, arrayAddr: number, offset, buffer, tag: TAGS, targetPC;
     var address = 0, isStatic = false;
     var ia = 0, ib = 0; // Integer Operands
     var ll = 0, lh = 0; // Long Low / High
 
     var classInfo: ClassInfo;
+    var otherCassInfo: ClassInfo;
```
mykmelez: Nit: otherCassInfo -> otherClassInfo

tests/org/mozilla/test/ClassCast.java
```diff
@@ -0,0 +1,27 @@
...
+    class ClassC extends ClassB {}
+
+    static TestHarness th;
+
+    public void test(TestHarness th) {
+        // This test illuminates an issue where ClassB's interfaces are setup
+        // in the runtime before the super class has been loaded.
+        Object o = new ClassC();
+        Interface iface = (Interface)o;
+        th.check(true);
```
mykmelez: It would be useful to note what happens here when the test fails (presumably an exception).

vm/parser.ts
```diff
@@ -1070,6 +1070,19 @@ module J2ME {
...
+   * ClassInfo objects, storing the ID for the Class in the first four bytes
+   * of the memory allocated for the Java object in the ASM heap.
+   *
+   */
+  export var generateClassId = (function() {
+    var nextId = 1;
+    return function() {
+      return nextId++;
+    }
+  })();
```
mykmelez: This doesn't appear to be used outside ClassInfo, so it'd be better as a private static property of that class.

marco-c: Right, this used to be used both in jit/compiler.ts and vm/runtime.ts (for objects and arrays), but that's no longer the case.

marco-c: Done in b284680.

int.ts
marco-c: Done in  9fa95b6.

tests/org/mozilla/test/ClassCast.java
marco-c: This is a test for a regression that Brendan saw while developing the patch, so I'm not totally sure what caused it.
I suppose the failure was a ClassCastException (which is the exception that would be thrown under the conditions outlined in the comment), but I can't tell for sure.

I will add a check for ClassCastException anyway.

marco-c: Done in 89e38d3.

vm/parser.ts
mykmelez: Nit: I would actually move this into the *ClassInfo* class, making it a `private static` property of that class:

```
    /**
     * Generate an ID for a ClassInfo.  We use this to map Java objects to their
     * ClassInfo objects, storing the ID for the Class in the first four bytes
     * of the memory allocated for the Java object in the ASM heap.
     *
     */
    private static nextId: number = 1;
    private static generateClassId(): number {
      return ClassInfo.nextId++;
    }
```

Then, in the *ClassInfo* constructor:

```
      this.id = ClassInfo.generateClassId();
```

That way it's only exposed to the one class that calls it. And we use the encapsulation built into TypeScript to hide *nextId* rather than bespoke JavaScript encapsulation.

(You could even get rid of *generateClassId* and simply set *this.id* to `ClassInfo.nextId++`.)

marco-c: Done in e68d089.

PR #1756, opened 2015-07-22
Remove util::stringToCharArray

marco-c:
Depends on #1752.


mykmelez:
Looks good, just needs #1752 to land.

PR #1761, opened 2015-07-24
Add a couple of GC benchmarks

marco-c:


mykmelez:
What's the license on GCBench.java? And where does BasicGCBench.java come from? I would have thought that it's original code, since it has no license/attribution header, except that it has a bunch of tab characters in it.

marco-c:
> What's the license on GCBench.java?

It isn't specified anywhere, this is the page where there's a link to it: http://www.hboehm.info/gc/gc_bench.html

> And where does BasicGCBench.java come from? I would have thought that it's original code, since it has no license/attribution header, except that it has a bunch of tab characters in it.

It is original code, I don't know what happened with the indentation :smile:

mykmelez:
@gerv The file GCBench.java in this pull request comes from http://www.hboehm.info/gc/gc_bench/, which is linked from http://www.hboehm.info/gc/gc_bench.html, which is linked from http://www.hboehm.info/gc/, which also links to http://www.hboehm.info/gc/license.txt.

Since http://www.hboehm.info/gc/gc_bench/ is a subdirectory of http://www.hboehm.info/gc/, is it reasonable to assume that the license file in that parent directory also applies to the files in the child subdirectory (and thus to GCBench.java)?

gerv:
myk: I'd say so, yes. If you check it in somewhere, accompany it with a copy of that license.txt. If you have modified it, add a single line at the top of license.txt which says:

"This code has been modified. The original file(s) can be found here: <URL>".

That's in keeping with the request in the license text itself.

Gerv

mykmelez:
Better to create a bench/GCBench/ subdirectory, I think, and add the file with the license file there.

marco-c:
Done in 7dd8113.

PR #1762, opened 2015-07-24
Fix frame adapter monitors

marco-c:


mykmelez:
Can we test for this, so we know if it breaks again?

marco-c:
This is only visible when compilation is enabled, so it should be covered by the basic unit tests with compilation enabled for all methods.

mykmelez:
Ok, fair enough.

PR #1763, opened 2015-07-25
Implement Runtime::gc, Runtime::totalMemory and Runtime::freeMemory

marco-c:


tests/org/mozilla/gc/TestRuntime.java
```diff
@@ -0,0 +1,37 @@
...
+        th.check(freeMemory > 0, "Free memory is strictly greater than 0");
+
+        for (int i = 0; i < 1000; i++) {
+            long[] array = new long[100000];
+        }
+
+        System.out.println("freeMemory2: " + Runtime.getRuntime().freeMemory());
+
+        th.check(totalMemory, Runtime.getRuntime().totalMemory(), "Total memory doesn't change");
+        th.check(freeMemory > Runtime.getRuntime().freeMemory(), "Free memory decreases after allocating some objects");
```
mykmelez: Since *array* is out of scope by this assertion, couldn't its memory have been collected already? If so, then this doesn't necessarily check that *array* allocation decreases free memory. So I would put the assertion in a block with the allocation:

```Java
{
  long[] array = new long[100000];
  th.check(freeMemory > Runtime.getRuntime().freeMemory(), "Free memory decreases after allocating some objects"); 
}
```

(Unsure if you need to allocate repeatedly, but if so, then that block could be the existing *for* loop, in which case the assertion would run for every iteration of the loop.)

marco-c: The problem with one small allocation is that you could end up freeing more memory than you've allocated (if a collection happens). A single big allocation should have the same effect of multiple small allocations, so I've done e976ccf.

mykmelez:
The current build is hung, but there are also two test failures in the basic unit tests:

>1.2:1.5  | Running gnu.testlet.vm.RuntimeTest 
>Native java/lang/Runtime.freeMemory.()J throws: TypeError: ASM._getUsedHeapSize is not a function
>1.2:1.5  | gnu.testlet.vm.RuntimeTest-0: fail [Test threw an unexpected exception: java.lang.NullPointerException: ASM._getUsedHeapSize is not a function] 
>1.2:1.5  | gnu.testlet.vm.RuntimeTest: 0 pass, 1 fail, 0 known fail, 0 unknown pass 
>1.2:1.5  | gnu.testlet.vm.RuntimeTest: test expected 1 passes, got 0 
>1.2:1.5  | gnu.testlet.vm.RuntimeTest: test expected 0 failures, got 1 
>1.2:1.5  | gnu.testlet.vm.RuntimeTest: class fail 
>…
>1.2:1.5  | Running org.mozilla.gc.TestRuntime 
>Native java/lang/Runtime.freeMemory.()J throws: TypeError: ASM._getUsedHeapSize is not a function
>1.2:1.5  | org.mozilla.gc.TestRuntime-0: fail [Test threw an unexpected exception: java.lang.NullPointerException: ASM._getUsedHeapSize is not a function] 
>1.2:1.5  | org.mozilla.gc.TestRuntime: 0 pass, 1 fail, 0 known fail, 0 unknown pass 
>1.2:1.5  | org.mozilla.gc.TestRuntime: test expected 5 passes, got 0 
>1.2:1.5  | org.mozilla.gc.TestRuntime: test expected 0 failures, got 1 
>1.2:1.5  | org.mozilla.gc.TestRuntime: class fail

marco-c:
There was en error in the latest merge from intex.

PR #1765, opened 2015-07-26
Some GC improvements

marco-c:
- Don't set the max heap size, this could be useful for debugging so I've left it commented out;
- Reduce the native.js size by not including two Emscripten library functions that aren't used;
- Ignore GC warnings, so we can build with NO_FILESYSTEM=1;
- Use GC_REGISTER_FINALIZER_NO_ORDER to fix the GC warnings about finalization cycles (see http://www.hpl.hp.com/hosted/linux/mail-archives/gc/2009-June/002786.html and http://www.hpl.hp.com/hosted/linux/mail-archives/gc/2009-June/002817.html).


vm/native/native.cpp
```diff
@@ -1,4 +1,6 @@
 //#define GC_DEBUG
+//#define GC_MAXIMUM_HEAP_SIZE (128 * 1024 * 1024)
```
mykmelez: After #1763, there'll be a Make variable, *ASMJS_TOTAL_MEMORY*, that defines this constant. I'm unsure if it's worth preprocessing this file just for that variable, although something to consider.

vm/native/native.cpp
```diff
@@ -81,5 +83,4 @@ extern "C" {
 
 int main() {
   GC_INIT();
-  GC_set_max_heap_size(128 * 1024 * 1024);
 }
```
mykmelez: Nit: while you're here, can you add a newline to the last line?

vm/native/native.cpp
marco-c: I think while debugging one wants to play with this value, so I'd just leave it as it is now.

PR #1768, opened 2015-07-27
WIP: support HiDPI displays

mykmelez:
@marco-c & @TimAbraldes: This is a WIP to support HiDPI displays by scaling the display canvas, and the gfx calls that draw to it, relative to the device pixel ratio. So a device pixel ratio of 2, as on Retina displays and Aries devices, results in a display canvas that is twice as large, and text, lines, and images are all scaled to be twice their specified size. The WIP then sets the CSS size of the canvas to its original size, so it fits into the device's screen.

Most tests work, but I haven't yet tackled the *DirectGraphics* gfx tests, and there's one class failure in the basic unit tests: *TestGraphicsClipping*. I haven't yet dug into these failures very much, but I guess it'll be hard to get these tests working, since they manipulate pixels directly. Do you have any suggestions for how to handle these cases?


mykmelez:
This now passes tests locally and on Travis, but WhatsApp still doesn't render correctly, and Asteroids renders too quickly, presumably because I haven't yet re-enabled *DisplayDevice.refresh0*.

mykmelez:
Closing in favor of non-WIP squashed branch in #1778.

PR #1774, opened 2015-07-28
Support multiple weak references to the same object

marco-c:
This fixes a bug with WhatsApp.
An "easy" way to reproduce it is to open the "Verify your phone number" screen and then open/close the country list several times.


marco-c:
https://github.com/mozilla/pluotsorbet/pull/1775 applies on top of this and contains a much better implementation.
There's no point in reviewing this by itself, so I'm closing it.

PR #1776, opened 2015-07-30
Fix TestRuntime

marco-c:
I think the problem was that we currently don't set pointers to object interiors as invalid.
Indeed, I was able to reproduce the bug on my machine if I set the size of the array to 10485760, but the bug wasn't reproducible if I merged https://github.com/mozilla/pluotsorbet/pull/1622 into my branch.

I've modified the test to make it reliable also for small allocations (so that https://github.com/mozilla/pluotsorbet/pull/1622 isn't required to make this test reliable) by forcing collections until every dead object is freed before running the test.


marco-c:
I've retriggered the build 5 times and the test is always passing.

PR #1778, opened 2015-08-04
scale display canvas by device pixel ratio to support HiDPI displays

mykmelez:
@TimAbraldes, @marco-c: This is the non-WIP squashed version of #1768. It scales the display by the device pixel ratio to support HiDPI displays.

It still isn't perfect, as it can't magically make images from image files be higher resolution than they actually are. It can also be defeated by midlets like WhatsApp that call ImageData.getRGB to retrieve pixel data from an image, manipulate it, and then call Graphics.drawRGB to write the manipulated pixel data back to the image.

But it sharpens other graphics on HiDPI displays, including generated graphics and text. In a separate pull request, I'll work on a hack that sharpens text in midlets like WhatsApp.


midp/gfx.js
```diff
@@ -98,10 +169,20 @@ var currentlyFocusedTextEditor;
...
+            return;
+        }
+
+        x1 = Math.max(0, x1) * DPR;
+        y1 = Math.max(0, y1) * DPR;
+        x2 = Math.max(0, x2) * DPR;
+        y2 = Math.max(0, y2) * DPR;
+
+        // XXX Isn't the offscreen and device canvas width and height always
+        // the same, so it isn't necessary to compute which one is more minimal?
```
TimAbraldes: Should be. If there are edge cases that cause us to need to do this min check (e.g. maybe when we rotate the screen during an animation that has repaint events pending) we should probably document those explicitly here.

midp/gfx.js
```diff
@@ -98,10 +169,20 @@ var currentlyFocusedTextEditor;
...
     Native["com/sun/midp/lcdui/DisplayDevice.refresh0.(IIIIII)V"] = function(hardwareId, displayId, x1, y1, x2, y2) {
-        x1 = Math.max(0, x1);
-        y1 = Math.max(0, y1);
-        x2 = Math.max(0, x2);
-        y2 = Math.max(0, y2);
+        if (!config.useOffscreenCanvas) {
+            // If the offscreen canvas is disabled, then we draw directly
+            // to the device canvas, so it's always up-to-date and never needs
+            // to be refreshed.
+            return;
```
TimAbraldes: Are we intending to disable the offscreen canvas for real MIDlets?

In the `DisplayDevice.refresh0` function we use `requestAnimationFrame` to ensure that we draw to the device canvas at an appropriate time. When the offscreen canvas is disabled, we draw to the device canvas directly and we don't take advantage of `requestAnimationFrame` at all.

If we're intending to disable the offscreen canvas for some MIDlets, we should file a follow-up issue to ensure that we only paint to the screen in a callback to `requestAnimationFrame`. We'll maybe have to create some way of batching paint requests.

midp/gfx.js
```diff
@@ -304,6 +424,8 @@ var currentlyFocusedTextEditor;
         // sync yet.
         var ctx = imageData.contextInfo.context;
 
+        // XXX Shouldn't we be able to simply fillRect here?
+
```
TimAbraldes: Possibly! But maybe the canvas context could have some kind of transformation or clipping rect currently applied?

midp/gfx.js
```diff
@@ -357,20 +479,19 @@ var currentlyFocusedTextEditor;
...
-    function calcStringWidth(font, str) {
+    function calcStringWidth(font, str, scaleByDPR) {
         var emojiLen = 0;
+        var fontSize = font.size * (scaleByDPR ? DPR : 1);
 
-        var len = font.context.measureText(str.replace(emoji.regEx, function() {
-            emojiLen += font.size;
+        tempContext.font = font.style + " " + fontSize + "px " + font.face;
+        var len = tempContext.measureText(str.replace(emoji.regEx, function() {
+            emojiLen += fontSize;
```
TimAbraldes: I think `calcStringWidth` is called a lot; creating the font every time hasn't created a performance penalty?

tests/gfx.js
```diff
@@ -0,0 +1,277 @@
+/* -*- Mode: Java; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- */
```
TimAbraldes: zomg awesome! This is so much better as its own file!

tests/gfx/DrawStringViaImageTest.java
```diff
@@ -0,0 +1,49 @@
+package gfx;
+
+import javax.microedition.lcdui.Canvas;
+import javax.microedition.lcdui.Display;
+import javax.microedition.lcdui.Graphics;
+import javax.microedition.lcdui.Image;
+import javax.microedition.midlet.MIDlet;
+import com.nokia.mid.ui.*;
+
+public class DrawStringViaImageTest extends MIDlet {
```
TimAbraldes: Sweet, this is a very necessary test given the way that WhatsApp in particular draws text

tests/gfx/DrawStringViaImageTest.java
```diff
@@ -0,0 +1,49 @@
...
+            Image image = Image.createImage(150, 150);
+
+            Graphics g = image.getGraphics();
+            g.setColor(0x00FFFFFF);
+            g.fillRect(0, 0, 150, 150);
+            g.setColor(0);
+            g.setStrokeStyle(Graphics.SOLID);
+            g.drawString("Roxors", 50, 50, Graphics.TOP | Graphics.LEFT);
+
+            graphics.drawImage(image, 50, 50, Graphics.TOP | Graphics.LEFT);
```
TimAbraldes: Maybe we should also do a getRGB/setRGB for completeness?

tests/javax/microedition/lcdui/TestGraphicsClipping.java
```diff
@@ -105,6 +106,21 @@ public void test(TestHarness th) {
...
                         th.check(verifyClip(unclippedData, clippedData, cx, cy, cw, ch));
+
+                        // Ensure that the various getClip* calls return
+                        // the correct values.
+                        th.check(g.getClipX(), cx);
+                        th.check(g.getClipY(), cy);
+                        th.check(g.getClipWidth(), cw);
+                        th.check(g.getClipHeight(), ch);
+
+                        // getClip returns a pair of coordinates rather than
```
TimAbraldes: nit: "a pair of" is confusing here since it returns 4 values

midp/gfx.js
mykmelez: Erm, my comment is a bit misleading. I meant to say that the offscreen and device canvas widths should always be the same, and their heights should similarly be the same. So these calls should be unnecessary:

```JavaScript
        var maxX = Math.min(offscreenCanvas.width, MIDP.deviceContext.canvas.width);
        …
        var maxY = Math.min(offscreenCanvas.height, MIDP.deviceContext.canvas.height);
```

I don't think there's ever a situation where they can be different, based on what I know of the code. As soon as the device canvas size changes, a canvasresize event is dispatched, and the canvasresize handler resizes the offscreen canvas to match the device canvas's size.

But perhaps I'm missing something. And changing it would be orthogonal to this patch. So I've left the code as-is but clarified the comment to read:

```JavaScript
        // XXX Aren't the offscreen and device canvas always the same size,
        // so it isn't necessary to compute their minimum width and height?
```

midp/gfx.js
mykmelez: > Are we intending to disable the offscreen canvas for real MIDlets?

I'm not planning to do this, I only added *config.useOffscreenCanvas* to help debug this branch. Perhaps I should simply remove it?

midp/gfx.js
mykmelez: > Possibly! But maybe the canvas context could have some kind of transformation or clipping rect currently applied?

Hmm, perhaps. I've added this question to the commentary.

midp/gfx.js
mykmelez: Clarification in 4dd0395.

midp/gfx.js
mykmelez: Question added in ce41bc1.

midp/gfx.js
mykmelez: > I think calcStringWidth is called a lot; creating the font every time hasn't created a performance penalty?

I didn't notice one, but I just ran GraphicsBench, and it reports that "Graphics.drawString baseline centered" and "Offscreen Graphics.drawString baseline centered" are roughly half as fast on this branch, regardless of DPR. So this branch does indeed regress string drawing perf. I'll look into it.

tests/gfx/DrawStringViaImageTest.java
mykmelez: > Maybe we should also do a getRGB/setRGB for completeness?

Good idea! Done in d0cd73a.

tests/javax/microedition/lcdui/TestGraphicsClipping.java
mykmelez: > nit: "a pair of" is confusing here since it returns 4 values

Indeed. It's technically correct, since a coordinate is a pair of values, but the "pair of pairs" implication is confusing. I've replaced it in ef8f471 with: "getClip returns the coordinates of the top-left and bottom-right corners rather than the top-left coordinate and a width/height."

midp/gfx.js
mykmelez: > I didn't notice one, but I just ran GraphicsBench, and it reports that "Graphics.drawString baseline centered" and "Offscreen Graphics.drawString baseline centered" are roughly half as fast on this branch, regardless of DPR. So this branch does indeed regress string drawing perf. I'll look into it.

291516a claws back most of the perf hit. It creates two contexts for measuring string widths, one for measuring unscaled widths and the other for measuring scaled widths (with an optimization to create only a single context if DPR == 1).

On my machine, the Graphics.drawString benchmarks perform 42-43K ops per second on master. On this branch, they now hit 34-35K. I think the rest is the necessary cost of drawing larger strings to a larger canvas when DPR == 2 (as it is on my machine).

Note that the other graphics benchmark, Graphics.drawRegion, also performs worse on this branch. That's expected since scaling the canvas means drawing more pixels, up to four times more when DPR == 2.

Regarding the use of two contexts to measure the scaled vs. unscaled widths of strings, I did some experimentation to confirm my earlier findings that merely multiplying an unscaled width by the DPR didn't give me an accurate scaled width.

Here are some measurements for "The quick brown fox jumps over the lazy dog." in two font styles (regular and "bold") at both 9px and 18px (per a theoretical DPR of 2). I calculate the result of simply multiplying the width at 9px by the theoretical DPR, and I also calculate what the return value of calcStringWidth would be using the current integer conversion mechanism (bitwise or) and a prospective alternative (Math.round).

9px sans-serif:
* 180.23333740234375
* 180.23333740234375 * 2 = 360.4666748046875
* 180.23333740234375 * 2 | 0 = 360
* Math.round(180.23333740234375) = 360

18px sans-serif:
* 360.73333740234375
* 360.73333740234375 | 0 = 360
* Math.round(360.73333740234375) = 361

bold 9px sans-serif:
* 194.43333435058594
* 194.43333435058594 * 2 = 388.8666687011719
* 194.43333435058594 * 2 | 0 = 388
* Math.round(194.43333435058594 * 2) = 389

bold 18px sans-serif:
* 389.4333190917969
* 389.4333190917969 | 0 = 389
* Math.round(389.4333190917969 | 0) = 389

Multiplying the unscaled widths by 2 gets close, but it can be a pixel off, regardless of the integer conversion. Perhaps that's ok, and we should simplify the scaled measurement. Unsure.

mykmelez:
I see some artifacts in the WhatsApp conversation view on this branch (as well as #1779, but it isn't specific to that branch), thin lines along parts of the borders of messages:



It isn't clear exactly why, though. Perhaps an off-by-one error somewhere?

mykmelez:
fb302a4 returns to the previous heuristic for calculating the default font size, as the new heuristic causes a few pixels of two text strings ("Update Now" and the version number "2.12.96" on a bar for updating the app) to get cut off along their right-hand edges in WA on my Aries.

Here's a comparison of a screen of the app in master (left), on this branch with the new heuristic (center), and on this branch with the old heuristic (right):

master fuzzy-wuzzy new heuristic fuzzy-wuzzy old heuristic

I'm not sure why this is happening, given that string widths should be calculated correctly regardless of whether the default font size on that device is 19 (with the old heuristic) or 18 (with the new one). But the heuristic change is actually unrelated to the changes in this branch, so I reverted them.

TimAbraldes:
@mykmelez I'm happy to land this. Is there any remaining work or known issues that you'd like to wait for?

mykmelez:
> @mykmelez I'm happy to land this. Is there any remaining work or known issues that you'd like to wait for?

The only remaining issue that I know about are the artifacts that appear in the screenshot in https://github.com/mozilla/pluotsorbet/pull/1778#issuecomment-131998271. I haven't had any luck figuring out why they occur. Do you have any insight?

mykmelez:

PR #1779, opened 2015-08-05
ignore pixel manipulations by special midlets

mykmelez:
@TimAbraldes, @marco-c: This branch applies on top of #1778 (see just the hack itself in commit 845d646) and hacks Graphics.drawRGB to ignore the modified data and instead draw the original image. When used with WhatsApp, there are artifacts in the parts of images that are being manipulated (f.e. text that is "faded out" on the right-hand side, as shown in the screenshot below), but otherwise the hack makes text sharp throughout the midlet.

And the only text that appears to be getting manipulated in this way is in the registration flow (f.e. the Terms screen and the Verify Your Phone Number screen). I don't see it in the Conversation List and Conversation View screens. So it might be possible to do something like this hack (but perhaps scrutinizing the pixel data to see if it's actually been manipulated, drawing the original image if it hasn't, and drawing the manipulated data if it has).

verify-your-phone-number


mykmelez:
The latest version of this hack fixes the issue with artifacts in the parts of the images being manipulated. It also limits the hack to whole-image changes, i.e. it only ignores the changes that a midlet is making via getRGB/drawRGB when the midlet is making them to an entire image. If the midlet calls getRGB for a subset of an image, that function downscales the image as before, and drawRGB then upscales it when drawing the modified pixels.

That means those subsets will still be blurry, but WhatsApp appears to do this only when fading the ends of lines of text in order to identify them as having been cropped, similar to the way ellipses are used. And the blurriness from the upscaling is not as noticeable there. Plus it's a small subset of text lines, which is still much better than all such lines being blurry.

Here's what the Verify Your Phone Number screen looks like with these changes:

verify-your-phone-number

And here's the Choose a Country list with a couple of items whose labels were long enough to have their ends faded:

choose-a-country

This may be the best we can do. I'm going to test with additional screens, and on device, to make sure it works correctly there and doesn't break anything.

TimAbraldes:
That looks AWESOME! Great work @mykmelez !!!

mykmelez:
One issue with ignoring changes to entire images is that WA uses that to change the hue of some text and make other text lighter. For example, in the screenshots below, it changes the hue of "Update Now" and makes "Search", "Hello, Myk!", "Thu", "2.12.96", "Options", and "Hide" lighter. Original on the left, this hack on the right:

TimAbraldes:
That doesn't seem like the worst thing in the world to me... but maybe it could cause usability issues with other text somewhere? I'm sort of conflicted since the image on the right really looks better to me, but since it's a correctness regression we might see something majorly busted down the road

mykmelez:
> That doesn't seem like the worst thing in the world to me... but maybe it could cause usability issues with other text somewhere?

In theory, it could, since a midlet can make arbitrary changes to the pixels. Of course this branch adds a configuration option, so we can selectively enable it for only those midlets that benefit from it without breaking because of it. But we still can't guarantee that it won't break an updated version of the midlet.

> I'm sort of conflicted since the image on the right really looks better to me, but since it's a correctness regression we might see something majorly busted down the road

I'm similarly conflicted! But I also haven't been able to think of a better way to support HiDPI displays.

mykmelez:
I've removed the WIP tag from the title of this pull request, as this is no longer a WIP, since I've verified that it does what I expect (modulo side-effects). Whether or not we take it is still a separate question.

mykmelez:

PR #1783, opened 2015-08-07
Fix race conditions in TestLocalMsgProtocol.

### LINKED ISSUE #1782: Fix race conditions in TestLocalMsgProtocol
TestLocalMsgProtocol has some logic that looks like this:

    synchronized (someLockObject) {
        someFlag = true;
        someLockObject.notifyAll();
    }

    someBlockingFunctionThatMakesFlagActuallyTrue();

That's clearly a race condition, as the condition flag is set before the condition is actually true. If we're relying on timing here, we're giving ourselves a false sense of security.

Additionally, TestLocalMsgProtocol uses wait-loops to wait for connections to be opened. However, it also opens those connections on separate threads, so the wait-loops are unnecessary because we can just `join` on the thread that creates the connection.

TimAbraldes:
TestLocalMsgProtocol has some logic that looks like this:

    synchronized (someLockObject) {
        someFlag = true;
        someLockObject.notifyAll();
    }

    someBlockingFunctionThatMakesFlagActuallyTrue();

That's clearly a race condition, as the condition flag is set before the
condition is actually true. If we're relying on timing here, we're
giving ourselves a false sense of security.

Additionally, TestLocalMsgProtocol uses wait-loops to wait for
connections to be opened. However, it also opens those connections on
separate threads, so the wait-loops are unnecessary because we can just
`join` on the thread that creates the connection.

Fixes #1782

### DIFF
--- a/tests/org/mozilla/io/TestLocalMsgProtocol.java
+++ b/tests/org/mozilla/io/TestLocalMsgProtocol.java
@@ -187,19 +187,9 @@ public class TestLocalMsgProtocol implements Testlet {
         }
     }
 
-    Object openLock = new Object();
-    boolean clientCreated = false;
-    boolean serverCreated = false;
-    boolean serverAcceptAndOpenCalled = false;
-
-    class ThreadClient extends Thread {
+    class ThreadClientCreate extends Thread {
         public void run() {
             try {
-                synchronized (openLock) {
-                    clientCreated = true;
-                    openLock.notifyAll();
-                }
-
                 client = (LocalMessageProtocolConnection)Connector.open("localmsg://" + PROTO_NAME + serverNum);
             } catch (Exception e) {
                 th.fail("Unexpected exception: " + e);
@@ -211,11 +201,6 @@ public class TestLocalMsgProtocol implements Testlet {
         public void run() {
             try {
                 server = (LocalMessageProtocolServerConnection)Connector.open("localmsg://:" + PROTO_NAME + serverNum);
-
-                synchronized (openLock) {
-                    serverCreated = true;
-                    openLock.notifyAll();
-                }
             } catch (Exception e) {
                     th.fail("Unexpected exception: " + e);
             }
@@ -225,11 +210,6 @@ public class TestLocalMsgProtocol implements Testlet {
     class ThreadServerAcceptAndOpen extends Thread {
         public void run() {
             try {
-                synchronized (openLock) {
-                    serverAcceptAndOpenCalled = true;
-                    openLock.notifyAll();
-                }
-
                 server.acceptAndOpen();
             } catch (Exception e) {
                 th.fail("Unexpected exception: " + e);
@@ -237,16 +217,9 @@ public class TestLocalMsgProtocol implements Testlet {
         }
     }
 
-    Object clientWaiting = new Object();
-    boolean clientIsWaiting = false;
-
     class ThreadClientWaitMessage extends Thread {
         public void run() {
             try {
-                synchronized (clientWaiting) {
-                    clientIsWaiting = true;
-                    clientWaiting.notifyAll();
-                }
                 clientReceiveData();
             } catch (IOException e) {
                 th.fail("Unexpected exception: " + e);
@@ -257,11 +230,6 @@ public class TestLocalMsgProtocol implements Testlet {
     class ThreadServerSendMessage extends Thread {
         public void run() {
             try {
-                synchronized (clientWaiting) {
-                    while (!clientIsWaiting) {
-                        clientWaiting.wait();
-                    }
-                }
                 serverSendData();
             } catch (Exception e) {
                 th.fail("Unexpected exception: " + e);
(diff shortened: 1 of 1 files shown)


tests/org/mozilla/io/TestLocalMsgProtocol.java
```diff
@@ -353,26 +309,13 @@ public void test(TestHarness th) {
...
-            synchronized (openLock) {
-                while (!serverAcceptAndOpenCalled) {
-                    openLock.wait();
-                }
-            }
-            Thread clientThread = new ThreadClient();
+            Thread clientThread = new ThreadClientCreate();
             clientThread.start();
-            serverCreateThread.join();
             serverAcceptAndOpenThread.join();
```
marco-c: Scenario 1 was:
1) Server creates a localmsg server
2) Server calls acceptAndOpen
3) Client opens a localmsg server

With this change, 3 could happen before 2.

tests/org/mozilla/io/TestLocalMsgProtocol.java
```diff
@@ -380,24 +323,11 @@ public void test(TestHarness th) {
...
-                    openLock.wait();
-                }
-            }
+            serverCreateThread.join();
             clientThread.start();
-            synchronized (openLock) {
-                while (!clientCreated) {
-                    openLock.wait();
-                }
-            }
```
marco-c: Scenario 2 was:
1) Server creates a localmsg server
2) Client opens a localmsg server
3) Server calls acceptAndOpen

With this change, 3 could happen before 2.

tests/org/mozilla/io/TestLocalMsgProtocol.java
```diff
@@ -405,24 +335,11 @@ public void test(TestHarness th) {
...
-            synchronized (openLock) {
-                while (!serverCreated) {
-                    openLock.wait();
-                }
-            }
-            serverAcceptAndOpenThread.start();
             serverCreateThread.join();
+            serverAcceptAndOpenThread.start();
             serverAcceptAndOpenThread.join();
             clientThread.join();
```
marco-c: Scenario 3 was:
1) Client opens a localmsg server
2) Server creates a localmsg server
3) Server calls acceptAndOpen

With this change, 2 could happen before 1 and 3 could happen before 1.

marco-c:
> That's clearly a race condition, as the condition flag is set before the
condition is actually true. If we're relying on timing here, we're
giving ourselves a false sense of security.

> Additionally, TestLocalMsgProtocol uses wait-loops to wait for
connections to be opened. However, it also opens those connections on
separate threads, so the wait-loops are unnecessary because we can just
join on the thread that creates the connection.

The test isn't very well documented, but there was a reason why I wrote it that way.
The simplest and probably best way to fix the race conditions is to use custom natives
for synchronization.

tests/org/mozilla/io/TestLocalMsgProtocol.java
TimAbraldes: > Scenario 1 was:
> 1) Server creates a localmsg server
> 2) Server calls acceptAndOpen
> 3) Client opens a localmsg server
>
> With this change, 3 could happen before 2.

Even without this change, 3 could happen before 2.

tests/org/mozilla/io/TestLocalMsgProtocol.java
TimAbraldes: > Scenario 2 was:
> 1) Server creates a localmsg server
> 2) Client opens a localmsg server
> 3) Server calls acceptAndOpen
>
> With this change, 3 could happen before 2.

Even without this change, 3 could happen before 2.

tests/org/mozilla/io/TestLocalMsgProtocol.java
TimAbraldes: > Scenario 3 was:
> 1) Client opens a localmsg server
> 2) Server creates a localmsg server
> 3) Server calls acceptAndOpen
>
> With this change, 2 could happen before 1 and 3 could happen before 1.

Even without this change, those outcomes are possible.

TimAbraldes:
I'll update the patch now that I have a better understanding of what the test is intending to accomplish:
  1. I'll add documentation for each case based on the "scenario" comments you wrote
  2. I'll add `Thread.sleep` calls where appropriate to hopefully mostly ensure that events happen in the order we want them to

marco-c:
> I'll add Thread.sleep calls where appropriate to hopefully mostly ensure that events happen in the order we want them to

Using custom natives for synchronization would actually fix the race conditions, but I'm OK with this workaround.

marco-c:
> I'll add documentation for each case based on the "scenario" comments you wrote

The scarce documentation was in the *th.check* calls, for example: `th.check(true, "Server create, server accept and open, client open");`.

tests/org/mozilla/io/TestLocalMsgProtocol.java
marco-c: Yes, in general it could, but it doesn't with the way the scheduler works now. It's definitely a weak assumption, so it's great you're fixing it.

TimAbraldes:
Updated: I think all review comments are now addressed

PR #1787, opened 2015-08-10
Compile native.cpp and Relooper.cpp together

marco-c:


mykmelez:
Looks good, but needs conflict resolution.

PR #1790, opened 2015-08-11
Fix NullPointerException caused by latest merge from master

marco-c:


mykmelez:
Aha, thanks, I missed that. I guess we don't have a test that would detect it.

PR #1794, opened 2015-08-11
Optimize newMultiArray

marco-c:
These changes speed up the attached microbenchmark (from ~5500 ms to ~4500 ms).


mykmelez:
Hmm, I wonder what we could do to speed up *newArray*.

brendandahl:
> Hmm, I wonder what we could do to speed up newArray.

A lot, it's currently slow.

mykmelez:
> A lot, it's currently slow.

Indeed, it seems like the main cause of slowness in *newMultiArray*.

PR #1800, opened 2015-08-13
use TypedArray.fill to fill new primitive arrays

mykmelez:
This helps *newArray* perf, albeit only on Gecko 37+. *MultiArrayBench* goes from ~3300 to ~2500 on my machine with the latest Firefox nightly build.


marco-c:
The loop version could be faster for small arrays. Could you add a benchmark that allocates arrays of different size?

marco-c:
(Another thing that we could try is zeroing-out the array directly in gcMallocAtomic in native.cpp)

mykmelez:
> The loop version could be faster for small arrays. Could you add a benchmark that allocates arrays of different size?

Yes, but the only change here is to the JS code that zeroes out the subset of the heap that represents the new array, so here's a JS benchmark that compares *TypedArray.fill* to a simple loop. It measures the time it takes to zero out various subsets of an Int8Array:

```JavaScript
var len, start, end = Math.pow(2, 16), i, j, k, ary, then, fillTime, loopTime,
    iterations = 30;
console.log("length: fillTime loopTime");
for (i = 0; i < 16; i++) {
  len = Math.pow(2, i);
  start = end - len;

  fillTime = 0;
  ary = new Int8Array(end);
  for (j = 0; j < iterations; j++) {
    ary.fill(1);
    then = performance.now();
    ary.fill(0, start, end);
    fillTime += performance.now() - then;
  }

  loopTime = 0;
  ary = new Int8Array(end);
  for (j = 0; j < iterations; j++) {
    ary.fill(1);
    then = performance.now();
    for (k = start; k < end; k++) {
      ary[k] = 0;
    }
    loopTime += performance.now() - then;
  }

  console.log(len + ": " + fillTime + " " + loopTime);
}
```

And here are some sample results when running this in the Web Console on Firefox Nightly:

> length: fillTime loopTime
> 1: 0.03539799968712032 0.05010700202547014
> 2: 0.030341999838128686 0.06441699992865324
> 4: 0.027742999838665128 0.09105099970474839
> 8: 0.030489999800920486 0.1501499987207353
> 16: 0.025906999362632632 0.2744600009173155
> 32: 0.026185001246631145 0.5251640006899834
> 64: 0.02671200013719499 1.039616000605747
> 128: 0.027396999998018146 2.0280289999209344
> 256: 0.028408999787643552 4.071157999103889
> 512: 0.031046000076457858 8.202111000427976
> 1024: 0.03623100044205785 16.777266999240965
> 2048: 0.04547200119122863 32.69952599960379
> 4096: 0.06874600006267428 66.11491499957629
> 8192: 0.10457400023005903 141.39908799901605
> 16384: 0.190492999041453 261.82732699927874
> 32768: 0.34059099946171045 524.4329090001993

Occasionally, the loop method will be faster when zeroing a single element. Reducing the number of iterations seems to make that more likely. But it also makes the results less significant. And it doesn't happen consistently even with a single iteration.

So it seems like *TypedArray.fill* is the winner here at all lengths.

marco-c:
> And here are some sample results when running this in the Web Console on Firefox Nightly:

IIRC the code executed in the Web Console isn't compiled, so the results might be different in a real page, for example in [1] looks like loop is faster than fill for an array of 4096 elements (unless I've done something wrong).

I suggested adding a Java benchmark to have something as close as possible to the real conditions (and that we can re-use for testing other improvements).

[1] http://jsperf.com/loop-vs-fill

mykmelez:
> IIRC the code executed in the Web Console isn't compiled, so the results might be different in a real page, for example in [1] looks like loop is faster than fill for an array of 4096 elements (unless I've done something wrong).

Putting that JS benchmark into a web page gives me similar results:

> length: fillTime loopTime
> 1: 0.27500000000000924 0.5550000000000068
> 2: 0.1850000000000307 0.24500000000003297
> 4: 0.03999999999996362 0.045000000000001705
> 8: 0.04499999999995907 0.03000000000002956
> 16: 0.044999999999987494 0.06000000000008754
> 32: 0.08999999999994657 0.14500000000003865
> 64: 0.08499999999992269 0.1449999999999818
> 128: 0.07999999999998408 0.20499999999992724
> 256: 0.16499999999990678 0.5200000000000955
> 512: 0.16000000000008185 1.0200000000001523
> 1024: 0.31499999999988404 2.029999999999859
> 2048: 0.6299999999998818 3.8749999999998863
> 4096: 1.1299999999998818 7.8900000000006685
> 8192: 2.165000000000191 16.16999999999996
> 16384: 4.454999999999586 31.89500000000021
> 32768: 9.425000000000182 62.11000000000058

The numbers bounce around a bit with small array lengths, and they're inconsistent across runs. But the gap widens as length increases, and it's quite significant at 4096 elements. I'm unsure why you see something different in your jsperf benchmark.

> I suggested adding a Java benchmark to have something as close as possible to the real conditions (and that we can re-use for testing other improvements).

Ok, I've added ArrayBench in 589bae2 to benchmark the allocation of arrays of various lengths. The numbers are imprecise for small array lengths, but they show equivalent performance at small lengths and TypedArray.fill becoming faster as length increases. Calling *memset* in *gcMallocAtomic* is even faster, however.

length | loop | fill | memset
------ | ---- | ---- | ------
1 | 38 | 36 | 34
2 | 2 | 2 | 2
4 | 2 | 2 | 2
8 | 2 | 2 | 1
16 | 2 | 1 | 2
32 | 1 | 2 | 1
64 | 2 | 2 | 2
128 | 2 | 2 | 2
256 | 3 | 4 | 3
512 | 6 | 5 | 6
1024 | 8 | 7 | 6
2048 | 34 | 32 | 31
4096 | 43 | 40 | 38
8192 | 37 | 32 | 25
16384 | 73 | 54 |  40
32768 | 102 | 85 | 53
65536 | 175 | 113 | 50
131072 | 293 | 221 | 112
262144 | 589 | 382 | 175
524288 | 1132 | 759 | 320

So a4839a0 replaces *TypedArray.fill* with *memset*, which brings *MultiArrayBench* down to ~1750 on Nightly, with similar relative perf gains on Firefox 36.0.4 (another benefit of *memset*, since *fill* would only help on 37+).

marco-c:
> The numbers bounce around a bit with small array lengths, and they're inconsistent across runs. But the gap widens as length increases, and it's quite significant at 4096 elements. I'm unsure why you see something different in your jsperf benchmark.

I wonder why, I simply copied your benchmark :astonished: 
I can reproduce your results in a page, perhaps the loop isn't compiled there anyway for some other reason?

Anyway since the C++ implementation is faster, we don't really need to answer the question :smile: 

We could probably write an optimized version of memset (since we only need to store 0 we can use HEAP64), but it's likely going to give a less interesting speed-up.

bench/ArrayBench.java
```diff
@@ -0,0 +1,30 @@
...
+        int len = 1;
+        int iterations = 1000;
+
+        System.out.println("Time to allocate int arrays of various lengths " + iterations + " times, by length, in ms:");
+
+        for (int i = 0; i < 20; i++) {
+
+            start = JVM.monotonicTimeMillis();
+            for (int j = 0; j < iterations; j++) {
+                int[] array = new int[len];
```
marco-c: Nit: It would be good if we could format the output of the benchmark in a way that is simple to parse, for example:
```
ArrayBench-SIZE: xxx
```

vm/native/native.cpp
```diff
@@ -69,7 +70,9 @@ extern "C" {
   }
 
   uintptr_t gcMallocAtomic(int32_t size) {
-    return (uintptr_t)GC_MALLOC_ATOMIC(size);
+    uintptr_t ptr = (uintptr_t)GC_MALLOC_ATOMIC(size);
+    memset((void*)ptr, 0, size);
```
marco-c: Nit: we could subtract the size of the header here because we don't need to zero it out, but I guess it doesn't make a big difference.

bench/ArrayBench.java
```diff
@@ -0,0 +1,30 @@
...
+        long start, time;
+        int len = 1;
+        int iterations = 1000;
+
+        System.out.println("Time to allocate int arrays of various lengths " + iterations + " times, by length, in ms:");
+
+        for (int i = 0; i < 20; i++) {
+
+            start = JVM.monotonicTimeMillis();
+            for (int j = 0; j < iterations; j++) {
```
marco-c: To see differences when the length of the array is small, we could use a larger number of iterations.

mykmelez:
> I can reproduce your results in a page, perhaps the loop isn't compiled there anyway for some other reason?

I don't think that's it, since loop numbers are much better in a page than in the console, suggesting it was compiled. Fill numbers are worse, though, which is strange.

mykmelez:
> We could probably write an optimized version of memset (since we only need to store 0 we can use HEAP64), but it's likely going to give a less interesting speed-up.

[emscripten's implementation of memset](https://github.com/kripken/emscripten/blob/3b99bb7bef220ad98760b117d14854098f86db9e/src/library.js#L3455-L3482) includes an optimization for large values of *num* that uses HEAP32, if I'm reading it correctly. So we might be able to optimize large sizes further.

marco-c:
> emscripten's implementation of memset includes an optimization for large values of num that uses HEAP32, if I'm reading it correctly. So we might be able to optimize large sizes further.

Yes, I think they can't use HEAP64 because they need to implement a generic memset, instead we only need to set the memory to 0 (and so we don't have problems with the conversion to double).

It'd also be interesting to test modifying Boehm.js to automatically clear the memory for GC_MALLOC_ATOMIC too. It could be done simply by modifying mark.c from:
```
/* PTRFREE */ { &GC_aobjfreelist[0], 0 /* filled in dynamically */,
                0 | GC_DS_LENGTH, FALSE, FALSE },
```
to:
```
/* PTRFREE */ { &GC_aobjfreelist[0], 0 /* filled in dynamically */,
                0 | GC_DS_LENGTH, FALSE, TRUE },
```

We can test these things in a follow-up though!

bench/ArrayBench.java
mykmelez: > Nit: It would be good if we could format the output of the benchmark in a way that is simple to parse, for example:

Ok, I made it output this way in a5800de.

vm/native/native.cpp
mykmelez: > Nit: we could subtract the size of the header here because we don't need to zero it out, but I guess it doesn't make a big difference.

Yeah, I suspect it doesn't make much difference, and it would complicate the implementation, since then gcMallocAtomic would need to know the header size. Just for kicks, I tried it over in this other branch, and it didn't seem to make a difference:

https://github.com/mykmelez/pluotsorbet/compare/intex-fill-new-primitive-array...mykmelez:intex-fill-new-primitive-array-offset

bench/ArrayBench.java
mykmelez: > To see differences when the length of the array is small, we could use a larger number of iterations.

Both kinds of benchmarks would be useful, so I've added a variable iteration benchmark in c42eeea. Results look like this:

> Time to allocate int arrays of variable lengths a constant 1000 times, by length, in ms: 
> ArrayBench-1: 6 
> ArrayBench-2: 2 
> ArrayBench-4: 2 
> ArrayBench-8: 2 
> ArrayBench-16: 1 
> ArrayBench-32: 2 
> ArrayBench-64: 2 
> ArrayBench-128: 2 
> ArrayBench-256: 3 
> ArrayBench-512: 7 
> ArrayBench-1024: 7 
> ArrayBench-2048: 31 
> ArrayBench-4096: 37 
> ArrayBench-8192: 25 
> ArrayBench-16384: 39 
> ArrayBench-32768: 34 
> ArrayBench-65536: 56 
> ArrayBench-131072: 92 
> ArrayBench-262144: 163 
> ArrayBench-524288: 327 
> Time to allocate int arrays of variable lengths a variable number of times, by length * iterations, in ms: 
> ArrayBench-1*524288: 225 
> ArrayBench-2*262144: 74 
> ArrayBench-4*131072: 37 
> ArrayBench-8*65536: 21 
> ArrayBench-16*32768: 11 
> ArrayBench-32*16384: 5 
> ArrayBench-64*8192: 3 
> ArrayBench-128*4096: 2 
> ArrayBench-256*2048: 1 
> ArrayBench-512*1024: 4 
> ArrayBench-1024*512: 2 
> ArrayBench-2048*256: 1 
> ArrayBench-4096*128: 1 
> ArrayBench-8192*64: 1 
> ArrayBench-16384*32: 1 
> ArrayBench-32768*16: 1 
> ArrayBench-65536*8: 1 
> ArrayBench-131072*4: 1 
> ArrayBench-262144*2: 0 
> ArrayBench-524288*1: 0

mykmelez:
(1d37750 was accidentally introduced when I merged intex-fill-new-primitive-array-offset into this branch to pick up c42eeea, which I'd accidentally committed on that branch instead of this one. I've reverted it in df19f34.)

mykmelez:
The comment about the gcMallocAtomic test in gctests.js actually applies to all the tests that check to see if allocating memory zeroes it out. So I've moved the comment out from inside the gcMallocAtomic test function and modified it to reference all those tests in 4371388.

mykmelez:
> It'd also be interesting to test modifying Boehm.js to automatically clear the memory for GC_MALLOC_ATOMIC too. It could be done simply by modifying mark.c from:

I tried this, and it might generate slightly better results for larger array lengths, although the numbers bounced around a bit across successive benchmark runs, and it was hard to discern an obvious win. It might be worth looking into more, but I don't think it's significant enough to wait for.

PR #1808, opened 2015-08-14
Set GC_INITIAL_HEAP_SIZE for WA and reduce ASMJS_TOTAL_MEMORY

marco-c:
This should improve startup time.

In my testing for WA, the heap size right after startup is 2285568 bytes (often it is 1658880 bytes, but I wouldn't expect the difference to matter much). It never grows to more than 3 MB, so I've set ASMJS_TOTAL_MEMORY to 4 MB for WA (the error that occurs when the memory isn't enough is pretty clear, so we can increase it easily if needed).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,10 @@ export CONFIG
 ASMJS_TOTAL_MEMORY ?= 16*1024*1024
 export ASMJS_TOTAL_MEMORY
 
+# Initial size of the GC heap
+GC_INITIAL_HEAP_SIZE ?= 0
+export GC_INITIAL_HEAP_SIZE
+
 # Whether or not to package test files like tests.jar and support scripts.
 # Set this to 1 if running tests on a device or building an app for a midlet
 # in the tests/ subdirectory, like Asteroids.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -315,7 +319,7 @@ endif
 bld/native.js: Makefile vm/native/native.cpp vm/native/Boehm.js/.libs/$(BOEHM_LIB) jit/relooper/Relooper.cpp jit/relooper/Relooper.h
 	mkdir -p bld
 	rm -f bld/native.js
-	emcc -DNDEBUG -Ivm/native/Boehm.js/include/ vm/native/Boehm.js/.libs/$(BOEHM_LIB) -Oz -O3 vm/native/native.cpp jit/relooper/Relooper.cpp -o native.raw.js --memory-init-file 0 -s TOTAL_STACK=16*1024 -s TOTAL_MEMORY=$(ASMJS_TOTAL_MEMORY) \
+	emcc -DNDEBUG -Ivm/native/Boehm.js/include/ vm/native/Boehm.js/.libs/$(BOEHM_LIB) -Oz -O3 vm/native/native.cpp jit/relooper/Relooper.cpp -o native.raw.js --memory-init-file 0 -s TOTAL_STACK=16*1024 -s TOTAL_MEMORY=$(ASMJS_TOTAL_MEMORY) -DGC_INITIAL_HEAP_SIZE=$(GC_INITIAL_HEAP_SIZE) \
```
mykmelez: Nit: could you wrap this onto a new line? These long lines are hard to read.

marco-c: Done in d454314.

PR #1811, opened 2015-08-14
Intex jit

brendandahl:


marco-c:
There are a couple of errors if you force a collection for every GC_MALLOC* call:
```
1.2:1.5  | org.mozilla.io.TestLocalMsgProtocolClose-4: fail [Test threw an unexpected exception: java.lang.NullPointerException: from is undefined] 
1.2:1.5  | org.mozilla.io.TestLocalMsgProtocolClose: 4 pass, 1 fail, 0 known fail, 0 unknown pass 
1.2:1.5  | org.mozilla.io.TestLocalMsgProtocolClose: test expected 16 passes, got 4 
1.2:1.5  | org.mozilla.io.TestLocalMsgProtocolClose: test expected 0 failures, got 1 
1.2:1.5  | org.mozilla.io.TestLocalMsgProtocolClose: class fail 
1.2:1.5  | Running org.mozilla.io.TestNokiaContactsServer 
1.2:63.5 | org.mozilla.io.TestLocalMsgProtocolClose-5: fail [Unexpected exception: java.lang.NullPointerException: classInfo is undefined] 
```

marco-c:
There's an exception in WA if you force a collection for every GC_MALLOC_* call:
```
2.3:18.5 | java.io.EOFException 
2.3:18.5 |    (stack trace incomplete) 
```

marco-c:
And if I start WA to the contact list, I see https://github.com/mozilla/pluotsorbet/issues/1813 if I disable runtime compilation or if I enable it but force a collection for every GC_MALLOC_* call.

marco-c:
(Notice that the stop function is still respected, so we don't collect if nativeFrameCount > 0)

marco-c:
An issue that could be related is:
```
0.2:0.5  | segmentation fault loading 1 bytes from address 0 
```

It's printed if you enable the SAFE_HEAP emcc option.

It could be related because some of the NullPointerException that I'm seeing are due to a invoke with address = 0.

marco-c:
One last issue is that WeChat doesn't work (also without forcing collections).
There's an exception during startup: `Exception from is undefined`.

marco-c:
21ff7ce fixed WeChat and appears to have fixed the Travis failure.
#1816 should fix the NullPointerException I was seeing here: https://github.com/mozilla/pluotsorbet/pull/1811#issuecomment-131356773

marco-c:
In the default configuration, I'm still seeing a lot of errors with SAFE_HEAP=1.

If I force collections for every GC_MALLOC* call, TestLocalMsgProtocolClose fails.

With compilation disabled and forced collections, there's an *EOFException* exception during startup  (but WA works anyway).

I'll file issues for them.

PR #1828, opened 2015-08-19
Force collection at every unwind.

brendandahl:
To allow this I had to:
-Fix local message stuff to hold on to objects it uses
-Use object hashes as the monitor id's, since objects may re-use the same address.
 
TODO monitors can leak memory in the new and old system, we should properly replace them.


marco-c:
>  -Fix local message stuff to hold on to objects it uses

There could be other natives that need to be fixed.
Before the JIT, everything that was passed to a native was always on the stack until the native returned, so we didn't have to worry about objects passed to async natives being collected (we only had to worry about objects created by async natives being collected, because those objects weren't on the stack).

brendandahl:
> There could be other natives that need to be fixed.

Yes, though I was unable to get any other failures enabling force collections on all allocations.

vm/runtime.ts
```diff
@@ -767,7 +767,12 @@ module J2ME {
   export function getMonitor(ref: number): any {
     release || assert(typeof ref === "number", "monitor reference is a number");
 
-    return monitorMap[ref] || (monitorMap[ref] = Object.create(null));
+    var hash = i32[ref + Constants.HASH_CODE_OFFSET >> 2];
+    if (hash === Constants.NULL) {
+      hash = i32[ref + Constants.HASH_CODE_OFFSET >> 2] = $.nextHashCode()
```
mykmelez: Nit: there's a semicolon missing here.

mykmelez:
This seems to resolve the memory issues in WhatsApp. I used to run out of memory almost as soon as I started scrolling a long contact list, and I could sometimes run out while navigating around other parts of the app. But so far I haven't run out of memory once despite scrolling through the list and back a few times and navigating around.

brendandahl:
Forcing only when we need to speeds up the yieldbench considerably. Times were around 3500ms, whereas forcing always I got inpatient after 20s and killed it.

vm/runtime.ts
```diff
@@ -666,6 +666,9 @@ module J2ME {
     TWO_PWR_32_DBL = 4294967296,
     TWO_PWR_63_DBL = 9223372036854776000,
 
+    // The target amount of free memory before garbage collection is forced on unwind.
+    FREE_MEMORY_TARGET = 4086,
```
mykmelez: We set *ASMJS_TOTAL_MEMORY* to 32MiB in the default Makefile, and we change its value to 8MiB in runwhatsapppreview.mk, so we should set *FREE_MEMORY_TARGET* in the Make files too, so we can set it relative to *ASMJS_TOTAL_MEMORY*.

vm/runtime.ts
```diff
@@ -1271,6 +1279,10 @@ module J2ME {
     return address;
   }
 
+  export function freeMemory(): number {
```
mykmelez: Nit: *free* is both an adjective and a verb, and function names often start with verbs, although *free* is actually being used as an adjective here, which makes this name confusing; so I would call it *getFreeMemory*.

vm/runtime.ts
mykmelez: > We set ASMJS_TOTAL_MEMORY to 32MiB in the default Makefile, and we change its value to 8MiB in runwhatsapppreview.mk, so we should set FREE_MEMORY_TARGET in the Make files too, so we can set it relative to ASMJS_TOTAL_MEMORY.

Erm, this was based on my faulty understanding that FREE_MEMORY_TARGET is being set to 4086K, i.e. about 4MB, whereas @brendandahl corrected me on IRC that it's actually being set to 4086 *bytes*, to minimize the frequency with which we force collection, while also leaving some space for the collector to allocate memory if needed (although we aren't sure if it's actually needed.)

In that case, we don't need to adjust *FREE_MEMORY_TARGET* relative to *ASMJS_TOTAL_MEMORY*. We might want to adjust this value in the future, depending on how effectively it balances the various concerns. But we can do that just easily here as in the Make files. So there's no good reason to move it.

marco-c: We should probably set it to a lower value, otherwise we'll run collections too rarely.

midp/localmsg.js
```diff
@@ -1101,33 +1102,40 @@ Native["org/mozilla/io/LocalMsgConnection.waitConnection.()V"] = function(addr)
...
         }
 
-        connection.sendMessageToServer(data, offset, length);
+        connection.sendMessageToServer(dataClone, 0, dataClone.length);
     }
 };
 
 Native["org/mozilla/io/LocalMsgConnection.receiveData.([B)I"] = function(addr, dataAddr) {
-    var data = J2ME.getArrayFromAddr(dataAddr);
+    J2ME.setUncollectable(dataAddr);
```
marco-c: Nit: we should set it uncollectable only in the wait* cases

PR #1833, opened 2015-08-20
enable compiler on intex branch

mykmelez:
With #1828, the compiler should be ready to enable on the intex branch. This branch does so.

It also increases *FREE_MEMORY_TARGET* to 4MiB as I noticed that WhatsApp was OOMing when it was set to 2MiB or less. We could use more insight into which code is allocating memory, and in what chunks. But this works for now.

### DIFF
--- a/tests/automation.js
+++ b/tests/automation.js
@@ -180,7 +180,7 @@ function syncFS() {
     });
 }
 
-casper.test.begin("unit tests", 37 + gfxTests.length, function(test) {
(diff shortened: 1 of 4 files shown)


marco-c:
I requested a PR on your branch to re-enable the tests that required the JIT to run (and the test that forced compilation for all methods): https://github.com/mykmelez/pluotsorbet/pull/5

PR #1838, opened 2015-08-21
Use Boehm's GC heuristics instead of our own

marco-c:
I still need to test this thoroughly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ else
 endif
 
 ifeq ($(RELEASE),1)
-	CLOSURE_FLAGS = 
+	CLOSURE_FLAGS =
 else
 	CLOSURE_FLAGS = --formatting PRETTY_PRINT
 endif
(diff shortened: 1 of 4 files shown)


mykmelez:
@marco-c I tested with RunWhatsApp on an Aries, and it doesn't get oomed, nor does perf seem to suffer. If anything, scrolling through a long contact list seems faster and smoother, perhaps because Boehm's heuristic is more aggressive than our current one, so it collects more frequently, regularly, and quickly. That certainly seems to be the case when I sample free memory using https://github.com/mykmelez/pluotsorbet/tree/intex-sample-free-memory.

I'll await your further testing before merging this, however.

marco-c:
> I'll await your further testing before merging this, however.

Done, I just wanted to measure memory usage.
After using WA for a while (opening/closing conversations, viewing images, scrolling conversation history, etc.) the heap size is 2813952 bytes (used memory 2347008 bytes), which seems pretty good to me!

PR #1850, opened 2015-08-25
Use a more predicable isolate test with assertions.

brendandahl:
Started this awhile ago and needed it again in another PR.
It should test nearly all the same things as the old one except it only verifies that the isolate id's increase, not that they are specific value.

### DIFF
--- a/tests/isolate/TestIsolate.java
+++ b/tests/isolate/TestIsolate.java
@@ -2,88 +2,65 @@ package tests.isolate;
 
 import com.sun.cldc.isolate.*;
 import tests.isolate.IsolatedClass;
+import gnu.testlet.DumpTestHarness;
 
 import java.lang.String;
 
 public class TestIsolate {
-    static int dumpNumber = 0;
-    public static void dump(String s) {
-        System.out.println((dumpNumber++) + ": " + s);
-    }
-    public static void dump(int s) {
-        dump(s + "");
-    }
-
     public static void main(String args[]) {
-        dump(IsolatedClass.val);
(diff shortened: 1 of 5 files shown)


tests/isolate/TestIsolate.java
```diff
@@ -2,88 +2,65 @@
...
-            Isolate iso1 = new Isolate("tests.isolate.IsolatedClass", new String[] { "1" });
-            Isolate iso2 = new Isolate("tests.isolate.IsolatedClass", new String[] { "2" });
+            Isolate iso1 = new Isolate("tests.isolate.IsolatedClass", new String[] { "1", "new isolate" });
+            Isolate iso2 = new Isolate("tests.isolate.IsolatedClass", new String[] { "2", "new isolate" });
 
-            dump(iso1.id());
-            dump(iso2.id());
+            int iso1Id = iso1.id();
+            th.check(iso1Id > myIsoId, "First isolate id is larger than main isolate id.");
+            th.check(iso2.id() > iso1Id, "Second isolate is larger than first isolate id.");
```
mykmelez: Nit: s/Second isolate/Second isolate id/.

tests/isolate/TestIsolate.java
```diff
@@ -2,88 +2,65 @@
...
 
-            if (Isolate.getIsolates().length == 1) {
-                dump("1 isolate");
-            }
+            th.check(Isolate.getIsolates().length, 1, "1 isolate created.");
```
mykmelez: Nit: s/created/started/.

PR #1852, opened 2015-08-26
Make it possible to run the shell in Node.

mbebenita:
```
node nodeshell.js Hello
```


mbebenita:
Node is considerably slower on benchmarks:

```
node nodeshell.js benchmark.BubbleSort
Total Time: 6326.0000 ms

js jsshell.js benchmark.BubbleSort
Total Time: 229.9600 ms
```

PR #1860, opened 2015-08-27
Use throw instead function calls for exceptions.

brendandahl:
Mbx and I found this when running the bubble sort benchmark. Seems to do nothing for startup time, but does improve benchmarks from 8600ms to 7900ms on my machine. Also improves bubble sort bench quite a bit if more rounds are run.

### DIFF
--- a/jit/baseline.ts
+++ b/jit/baseline.ts
@@ -960,7 +960,7 @@ module J2ME {
       if (this.isPrivileged) {
         return;
       }
-      this.blockEmitter.writeLn(length + "<0&&TS();");
+      this.blockEmitter.writeLn("if(" + length + "<0)throw " + ExceptionType.NegativeArraySizeException + ";");
     }
 
     emitBoundsCheck(array: string, index: string) {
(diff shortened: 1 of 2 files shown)


mbebenita:
I think what is happening is that function calls in loops (even though they are never called) prevent some optimizations from happening.

This Java code compiles to:

```java
for (int i = 0; i < 10000; i++) {
  a[i] = i;
}
```

```
for (int i = 0; i < 10000; i++) {
  !a&&TN();
  if((i>>>0)>=(i32[a>>2]>>>0))TI(i);
  a[i] = i;
}
```

I think the presence of TN and TI prevent some optimizations, otherwise it doesn't really explain why the new patch is faster. This is now:

```
for (int i = 0; i < 10000; i++) {
  if(!a)throw 0;
  if((i>>>0)>=(i32[a>>2]>>>0))throw 1;
  a[i] = i;
}
```

Where `0` means a null pointer exception, and `1` is an array index out of bounds exception.

brendandahl:
> What do we do about the index? Also, what do we do about negative array indices? Is that handled correctly down the line?

There's a constructor for the exception without the index. Negative array indices are array bounds exceptions, creation with a negative array length is the NegativeArraySizeException.

mykmelez:
@mbebenita Are you satisfied with this change now?

mykmelez:

PR #1865, opened 2015-08-28
Add a partially java compatible shell.

brendandahl:
Makes it much easier to compare output to hotspot from the command line and makes it possible to easily switch between node and spidermonkey.

Changes from the current shell:
- tests.jar and benchmark.jar are no longer automatically included you'll need to add it to the class path ```js shell/pluot.js -cp bench/benchmark.jar benchmark.InvokeVirtual```
- options are now more like java where a ':' separator is used e.g. flag is now ```-w:jc```
- ```-jar``` option was added so it is now possible to run a jar that has the main class in the manifest file
- Node uses the same file so just exchange the above ```js``` for ```node```

Maybe future work:
- The classpath support still needs some work to support directories and not just jar files to match java.
- Remove the default of printing stats
- Remove message from emscripten about "pre-main prep time: 0 ms"


mbebenita:
:+1:

PR #1867, opened 2015-08-29
remove obsolete methodInfo property

mykmelez:
I don't think this is being used anymore, and tests pass without it, but requesting a pull so someone else can confirm. @brendandahl or @mbebenita?


brendandahl:
Yeah, I think that's a relic of the old system. We don't do it for linkMethodFunction.

PR #1874, opened 2015-12-07
Issues 1873 - Set typescript version for tsc compile error.

Seinlin:


mykmelez:
Thanks @Seinlin!

There's still a test failure, but it's unrelated to this change, which just uncovers it. I filed that one as #1875.

PR #1877, opened 2017-01-05
Fix j2me for type script v1.8.10 and TextEncoder changes.

brendandahl:


brendandahl:
Failure is expired cert.

